Issue #86, opened 2014-06-04
support for HTTPS on express 


dutu:
please include HTTPS support on express, preferably this should be the default options, perhaps with a check on available certificates

PR #140, opened 2014-08-16
adding https option for running with TLS/SSL

jloveland:
This addresses issue #86 by adding the option to configure express using https instead of http. The 'all' and 'production' environment configuration includes a property named 'secure' to enable https.

### DIFF
--- a/config/express.js
+++ b/config/express.js
@@ -5,10 +5,6 @@
  */
 var fs = require('fs'),
 	http = require('http'),
-	https = require('https'),
-	privateKey  = fs.readFileSync('./config/sslcert/key.pem', 'utf8'),
-	certificate = fs.readFileSync('./config/sslcert/cert.pem', 'utf8'),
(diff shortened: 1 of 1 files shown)


brokyo:
Apologies for bumping a year old thread but did anything ever come of this? Looking to run an app through SSL and hoping I'm just overlooking a config variable somewhere

codydaig:
@brokyo The latest version has https support. There's instructions in the README, and there is a config option to set the `'secure' : true`

jloveland:
@brokyo this was merged and you can run Mean.js with TLS/SSL. You need to do the following:
1. generate your ssl certificates `./scripts/generate-ssl-certs.sh`
2. run the app in production mode `grunt prod` or `gulp prod`
3. navigate to `https://localhost:8443/`

brokyo:
God that's beyond embarrassing. It's right there in the docs! Total user error :(

Thanks so much to both of you

PR #164, opened 2014-08-30
Adding Service Layer Implementation

podoler:
Separate controller logic into two separated layers:
1. Controller layer - to handle http transport layer logic
2. Service layer - to handle business logic

### DIFF
--- a/app/controllers/articles.server.controller.js
+++ b/app/controllers/articles.server.controller.js
@@ -6,23 +6,21 @@
 var mongoose = require('mongoose'),
 	errorHandler = require('./errors'),
 	Article = mongoose.model('Article'),
-	_ = require('lodash');
+	_ = require('lodash'),
+    articleService = require('../services/articles.server.service');
 
 /**
  * Create a article
  */
 exports.create = function(req, res) {
-	var article = new Article(req.body);
-	article.user = req.user;
+    articleService.create(req.body, req.user, function(err, article){
+        if (err) {
+            return res.status(400).send({
+                message: errorHandler.getErrorMessage(err)
+            });
+        }
 
-	article.save(function(err) {
-		if (err) {
-			return res.status(400).send({
-				message: errorHandler.getErrorMessage(err)
-			});
-		} else {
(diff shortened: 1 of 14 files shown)


ilanbiala:
@lirantal @amoshaviv @roieki @NeverOddOrEven this really just moves the code to another file, which I don't see as necessary. Do you guys have any strong preference for this change?

podoler:
When the project is small, this might seem as a redundant overhead,
but when things start to grow and get more volume, this abstraction will be necessary for code reuse, better testing, easy to read, decoupled code etc.

ilanbiala:
@podoler right, but we shouldn't force this on users, especially with such a small codebase to start. Also, the service layer isn't really decoupling or making this easier to test, it just splits it into 2 functions basically...

podoler:
what @NeverOddOrEven suggested sounds right, if you do not want to force this upon users.
As for the decoupling and testing, from my experience, it really helps... We could reuse the service layer more easily and testing became straight forward...

ilanbiala:
This seems like too much maintenance even for a sub-generator, I would just add a page to the Wiki saying here's a suggestion for larger projects.

lirantal:
honestly speaking this is quite a large PR and I'm unable to review it properly in terms of schedule. I would anyway suggest to do this PR with the 0.4 branch, if at all.

ilanbiala:
@lirantal the main idea is whether to move the actual logic into a separate method, and then call that function then do `res.json()`. The issue I have is that it doesn't suit most users' needs and it just moves code to a different folder, which I feel is unnecessary.

lirantal:
@podoler looks like a considerable PR but we can't merge it into master sadly.
Is this still relevant for 0.4.0?

codydaig:
Since there is no active discussion, and not many people are onboard with this idea, I'd suggest we close. If someone wants to bring this back up for discussion, it would be done against the latest version.

lirantal:
@podoler this an old PR that we can't merge but if you'd like to keep up with PR contributions to the project we'd love to hear more from you.

PR #171, opened 2014-09-05
Developer Support - Warn when libs are not specified in config

NeverOddOrEven:
There are numerous forum topics where developers are forgetting to update all.js. This is a diagnostic that runs when not in production, to assist developers.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -9,7 +9,7 @@ var _ = require('lodash'),
 /**
  * Load app configurations
  */
-module.exports = _.extend(
+module.exports =  _.extend(
 	require('./env/all'),
 	require('./env/' + process.env.NODE_ENV) || {}
 );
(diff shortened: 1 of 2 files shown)


config/config.js
```diff
@@ -61,16 +61,53 @@ module.exports.getJavaScriptAssets = function(includeTests) {
...
+  if (!process.env.DISABLE_PUBLIC_LIB_CHECK && process.env.NODE_ENV !== 'production') {
+    var configJsAssets = this.getJavaScriptAssets(),
+        configCssAssets = this.getCSSAssets(),
+        configTestsAssets = this.getTestAssets('public/');
+    var allConfigAssets = _.union(configJsAssets, configCssAssets, configTestsAssets);
+    
+    // Get all the assets that have been loaded from NPM
+    var assetPathPresentInProject = this.getGlobbedFiles('./public/lib/*', './public/');
+    
+    console.info('\x1b[30m\x1b[47m');
```
ilanbiala: Use chalk here and later to make the code cleaner.

lirantal:
@NeverOddOrEven I'm not sure I understand what are the configuration checks for exactly. What libs are supposed to be specified in all.js?

NeverOddOrEven:
This may not necessarily be relevant anymore. The support forum was seeing a lot of issues where people would add a dependency to their project, but not add it to all.js. I haven't seen that issue as much lately.

This pull request would parse the all.js external files against those in the lib folder. Any top level library not in the all.js would be flagged and displayed with a warning.

ilanbiala:
@NeverOddOrEven I think people are understanding the build and deployment process now, and I think it is less common. We can always revisit this later, but I think to keep the code simple and applicable to most people, we should leave it out unless the issues come up again.

PR #195, opened 2014-10-04
html5Mode support.

rhutchison:
Open for discussion.  This pull request enables use of html5Mode.  This should also still work with html5Mode(false).


woniesong92:
Is it merged?

simison:
@xxorax @woniesong92 @ebraminio perhaps here? https://github.com/meanjs/mean/tree/0.4.0

karolkochan:
I have found some issue. Direct access to the routes that are 'first level' works just fine, but when I try direct access on 'second level' routes, angular (I assume it is angular from error in chrome inspector) gets into infinite loop. Example with sample app on localhost:

Type in browser: 
`http://localhost:3000/signin` = works fine

Now change url route in `/public/modules/users/config/users.client.routes.js` from `/signin` to `/signin/in` and try to type in a browser: 
`http://localhost:3000/signin/in` = gets into infinite loop 

However, if I crawl to this route through homepage (not direct access) it works ok. I'm a newbie in MEAN stack so I have no idea how to fix this.

monad98:
@rhutchison 
I still have problem. When I refresh the page at "/articles" or "/articles/create", I am automatically signed out.
The other pages like "/settings/profile" is ok.

rhutchison:
@monad98 this PR is pretty out dated from upstream, so not sure how reliable it would be.  I know @amoshaviv manually plucked these changes and put them in to the 0.4.0 branch (https://github.com/meanjs/mean/commit/ab81d61bd3cc629d55e0b9eeb8914f2254e191db#diff-90593c1684669b4a0cc185ec424f0a24R9) - have you had any luck there?

PR #204, opened 2014-10-13
add preloadProvider - for preloading data.

rhutchison:
I had a need to preload data to my angular app. After some research, I came up with this.

Credit: http://bitwit.ca/blog/regarding-angularjs-data-preloading/


app/views/layout.server.view.html
```diff
@@ -40,6 +40,13 @@
 	<!--[if lt IE 9]>
 		<script src="http://html5shim.googlecode.com/svn/trunk/html5.js"></script>
 	<![endif]-->
+
+	{% if user %}
+	<!--Embedding The User Object-->
+	<script type="text/preloaded" name="user">
```
ilanbiala: What is the `name` attribute for?

public/application.js
```diff
@@ -4,11 +4,24 @@
...
-]);
+angular.module(ApplicationConfiguration.applicationModuleName).config(['$locationProvider', function($locationProvider) {
+	$locationProvider.hashPrefix('!');
+}])
+.constant('$preloadProvider', {})
+.config(['$preloadProvider', function($preloadProvider) {
+    var scripts = document.getElementsByTagName('script');
+    for (var i = 0; i < scripts.length; i++) {
+        var script = scripts[i];
+        if (script.type === 'text/preloaded') {
```
ilanbiala: This part seems way to easy to break if someone changes a single attribute. Any other ideas on how to handle it?

app/views/layout.server.view.html
rhutchison: @ilanbiala, see below:
* name is not a valid attribute on a script tag.
* name serves as the array key of $preloadProvider constant.

public/application.js
rhutchison: What is the concern?  There are only two supported attributes, type & name.

ilanbiala:
@lirantal what do you think about this? It definitely needs to wait until 0.4.0 is released, but is this something we should consider?

lirantal:
@ilanbiala  @rhutchison  I don't like all the for loop handling there to parse the DOM, and having to neat pick data from an element's HTML value.

@rhutchison can you come up with other ways to work around the problem you are describing? there must be a more elegant way to do this.

rhutchison:
I found it to be pretty elegant.  How is this any different then how in-line templates are handled?

ilanbiala:
I don't like the for loop either, and I don't think many people will need the extra complexity.

lirantal:
@rhutchison I'll close this PR since I believe there is a more proper way of implementing this, and also this PR is against 'master' branch and not 0.4.0

We'll get back to it later if you wish, with a newer PR.

Issue #238, opened 2014-10-29
Validation is no longer working


jeromegv:
Hi

It seems that the client-side validation is not working on the article example 

If you try to edit an existing article, the form will let you submit the article even thought you have an empty title and empty content

Let me know if you need more details

Thanks

PR #247, opened 2014-11-03
fix clientside angular validation on article example, issue #238

jeromegv:


public/modules/articles/controllers/articles.client.controller.js
```diff
@@ -4,19 +4,23 @@ angular.module('articles').controller('ArticlesController', ['$scope', '$statePa
...
 
-		$scope.create = function() {
-			var article = new Articles({
-				title: this.title,
-				content: this.content
-			});
-			article.$save(function(response) {
-				$location.path('articles/' + response._id);
+		$scope.create = function(isValid) {
+			if (isValid) {
```
ilanbiala: Seems like a good idea to check, but it would probably make more sense to set the inputs as required and not have to do a check in JS if necessary. Thoughts?

ilanbiala:
@jeromegv let's keep dependency updates out of this PR because it is about something else. Otherwise I like validating it, but does doing form validation through HTML not work?

lirantal:
@jeromegv if it's still relevant for 0.4.0 branch can you please submit a new PR?

codydaig:
@lirantal There's an open PR #759 that addresses this.

lirantal:
Thanks @codydaig so we can close this.

PR #253, opened 2014-11-05
Correctly encode and decode password salt

rmuch:
The user password salt should be encoded with Base64 before being saved
to the database.

The current code adds an unecessary step of converting the result of
crypto.randomBytes() (which already returns a SlowBuffer) to a Base64
string and back again to a Buffer, and misses the final step of
converting the Buffer's bytes back to a Base64 string.

Because of this, the salt stored in the database is garbled. This is
inconvenient when manipulating the data in a terminal or text editor.

When generating the password hash, the crypto.pbkdf2Sync() method
creates a new Buffer directly from the data supplied. Due to the
incorrect encoding of the salt, entropy is lost at this step,
weakening the security of stored passwords against brute force attacks.

### DIFF
--- a/app/models/user.server.model.js
+++ b/app/models/user.server.model.js
@@ -96,7 +96,7 @@ var UserSchema = new Schema({
  */
 UserSchema.pre('save', function(next) {
 	if (this.password && this.password.length > 6) {
-		this.salt = new Buffer(crypto.randomBytes(16).toString('base64'), 'base64');
+		this.salt = crypto.randomBytes(16).toString('base64');
 		this.password = this.hashPassword(this.password);
 	}
 
@@ -108,7 +108,7 @@ UserSchema.pre('save', function(next) {
  */
 UserSchema.methods.hashPassword = function(password) {
 	if (this.salt && password) {
-		return crypto.pbkdf2Sync(password, this.salt, 10000, 64).toString('base64');
+		return crypto.pbkdf2Sync(password, new Buffer(this.salt, 'base64'), 10000, 64).toString('base64');
 	} else {
 		return password;
 	}


ilanbiala:
@lirantal @amoshaviv I think a password reset is a good idea, and just doing that to every user is a one-time call, and it migrates the users properly.

ilanbiala:
If we put together a password reset snippet and provide that to users in addition to this fix, I think it's sufficient to get their systems up-to-date with the changes.

ilanbiala:
Since we are still pre 1.0.0 and I don't think users of MEAN really download the new version and change out all their code, I think 0.4.0 is a good time to fix this. We can still publish a migration script or something similar to help others who want to fix this in their projects, but we really shouldn't hold this change back...

@lirantal @NeverOddOrEven @roieki @amoshaviv what do you guys think?

lirantal:
I agree, it's a painful change but it should be taken as soon as possible.
Unfortunately, there's no migration route to take as the passwords can't be deciphered, so the only options are:
1. Users would roll out their own way of forcing password resets to their users
2. They exchange the new password handling code with the old one to stay compatible, even though the old algorithm is not working properly.

ilanbiala:
```javascript
var mongoose = require('mongoose'),
  User = mongoose.model('User'),
  nodemailer = require('nodemailer'),
  config = require('./config/config');

var transport = nodemailer.createTransport(config.mailer.options);
var link = 'reset link here';

User.find({}, function(err, users) {
  if (err) {
    throw err;
  }

  var email = {
    from: 'noreply@xyz.com',
    subject: 'Security update'
  };

  for (var i = 0; i < users.length; i++) {
    var text = [
      'Dear ' + users[i].displayName,
      '\n',
      'We have updated our password storage systems to be more secure and more efficient, please click the link below to reset your password so you can login in the future.',
      link
      '\n',
      'Thanks,',
      'The Team'
    ].join('\n');

    email.text = text;
    email.html = text;

    transporter.sendMail(email, function(err, info) {
		if (err) {
        console.log(
			console.log(err);
		} else {
			this.isSent = true;
			p.resolve(info);
		}
	});
  }
});
```

WIP email script.

lirantal:
I didn't fully review but looks good enough. I think the password issue was the major thing that held back the 0.4 branch merging into master (including more tests on 0.4, conflicts resolutions etc), but I'm all up for that @ilanbiala.

ilanbiala:
@lirantal doesn't it actually make more sense to merge this in 0.4.0, because then this fix will be in future versions. If we merge into master, it might not properly make it into 0.4.0 and on.

lirantal:
@ilanbiala good point indeed.
@rmuch can you create a new PR to merge with the 0.4 branch?

rmuch:
@lirantal I've created a new pull request for the 0.4.0 branch at https://github.com/meanjs/mean/pull/376

zlatinejc:
Hi guys.
Just an observation:
You could put a checker in the login flow so you can grab the password and encrypt it using the proposed "better entropy" encription etc...
?

I am guessing you would need an additional field in the model to check that the pass has been moved to the newer version.

My humble opinion:
Better than notifying customers of "security upgrade" which they thought it was already secure. I am not into this black/white code cutting ways...

Best!

miguelcoba:
The issue about old users shouldn't stop this MR from merging to 0.4 (and from there to master).
As others pointed out:
- this is pre 1.0, so changes are expected by end users
- a reset link option is provided if needed

Also, if someone has in fact a lot of users, that is enough motivation to either:
- manually send them the reset link
- to patch his installation to use the old code
- to have the conditional logic that uses the old encoding if so he decides
- to add a new column to track which users are already migrated

So IMHO this is a nonissue and the actual benefits of having it merged are bigger that the supposed disadvantages

ilanbiala:
@lirantal I have to agree with @miguelcoba that we've been spending too much time trying to solve this for old users when they've probably fixed it themselves already. I think we should just merge this in and leave a word of caution in the docs when upgrading.

lirantal:
@ilanbiala sure, I agree as well. Let's merge it for master branch, as well as 0.4 please (I see there's a PR for that too).

Another thing @ilanbiala, we have a scripts/ folder, can you please add a PR to place the reset password gist you've attached above?

ilanbiala:
I'll create a PR for that, does this PR and #376 LGTY? If so, you can merge.

lirantal:
Great, thanks @ilanbiala

wesleyfsmith:
I've got three different production apps, all with users hahaha.

That being said, I'd rather get the issue fixed now, I would just like to know what is the recommended strategy for existing users in my apps. What would be the best way for me to force a password reset/do a migration. Also, is the consensus that we should just do a password reset? I wasn't quite clear reading the backlog.

lirantal:
Hey @wesleyfsmith, thanks for joining us on this.

ilanbiala added a gist to help deal with existing meanjs installations which use the current and flawed encryption: https://github.com/meanjs/mean/pull/458/files

ilanbiala:
@wesleyfsmith you'll need to provide the link somehow, but otherwise it should do everything else for you.

PR #267, opened 2014-11-19
Fix two issues:

AlexBoYang:
1. Visiting http://localhost:3000/abc/def (this path does not exist) will not display 404 page. That is because the css and javascript library path is relative directory not root directory of MEAN.
2. Do not remove the "removeRoot" string ("public/" in MEAN source code) only if it is at the beginning of the path


lirantal:
@AlexBoYang I tested on Linux and it looks ok.

ilanbiala:
@NeverOddOrEven does this conflict/overlap with the image path PR?

lirantal:
@AlexBoYang looks like this is till happening on 0.4.0, would you like to send a PR to fix it?

ilanbiala:
@AlexBoYang @lirantal is this still a problem?

AlexBoYang:
Let me check.

AlexBoYang:
Those issues have been solved in current master branch.

PR #275, opened 2014-11-21
Made changes related to remove functionality.

kuldipem:


public/modules/articles/views/view-article.client.view.html
```diff
@@ -6,7 +6,7 @@ <h1 data-ng-bind="article.title"></h1>
 		<a class="btn btn-primary" href="/#!/articles/{{article._id}}/edit">
 			<i class="glyphicon glyphicon-edit"></i>
 		</a>
-		<a class="btn btn-primary" data-ng-click="remove();">
+		<a class="btn btn-danger" onclick="return confirm('Are you sure want to remove ?');" data-ng-click="remove();">
```
ilanbiala: Should be `ng-click` not `onclick`, and why return it? Also, don't use an ugly confirm dialog.

ilanbiala:
@lirantal @NeverOddOrEven I think most people would be prompted before deleting something to confirm. I like the idea, but I think some things need to be fixed. Thoughts?

NeverOddOrEven:
There's a conversation in the generator here kulpidem/generator-meanjs#53.

NeverOddOrEven:
Here's the direct link https://github.com/meanjs/generator-meanjs/pull/53

lirantal:
@ilanbiala you're absolutely correct - the idea is good and I'm up for adding it but the coding convention there is primitive. @kuldipem please update to use proper angular/javascript code to confirm deletions.

kuldipem:
@lirantal @NeverOddOrEven  , I have just updated onclick => ng-click, is it ok ?

lirantal:
I'm not the authority for confirming that as my strong side isn't AngularJS atm but I think there are probably cleaner, and better ways to create a confirmation dialog than to do 'return confirm()' which is the default browser window function.

public/modules/articles/views/view-article.client.view.html
```diff
@@ -6,7 +6,7 @@ <h1 data-ng-bind="article.title"></h1>
 		<a class="btn btn-primary" href="/#!/articles/{{article._id}}/edit">
 			<i class="glyphicon glyphicon-edit"></i>
 		</a>
-		<a class="btn btn-primary" data-ng-click="remove();">
+		<a class="btn btn-danger" ng-click="return confirm('Are you sure want to remove ?');" data-ng-click="remove();">
```
ilanbiala: You can't have `ng-click` and `data-ng-click`...one will override the other. Also, stop using confirm(), it is not pretty...Keep all the logic in the controller.

lirantal:
@kuldipem since @ilanbiala's comments haven't been addressed I'm closing this PR.
You or anyone else are welcome to open a new PR for 0.4.0 branch to address this problem.

Issue #12, opened 2014-03-30
bower_components in .gitignore


joeyvmason:
Is there a reason bower_components isn't in the .gitignore file? It seems like it doesn't really need to be in the git repo.

PR #280, opened 2014-11-26
0.4.0 - 'username' deprecated in facebook api >=2.0

mattraby:
When creating a user account using facebook oauth the error "(#12) username is deprecated for versions v2.0 and higher" is returned. Removing 'username' from the passport facebook strategy's 'profileFields' will correct this issue. The user account will be created using the local part of the user's email address as the username.


ilanbiala:
@lirantal @amoshaviv we should probably move to Facebook API >=2.0 because the older versions will be deprecated at the end of April.

mattraby:
Closing PR, taken care of in #384

PR #284, opened 2014-12-01
Angular services should be loaded first

herveguetin:
Some Angular providers may be needed at config state. So the services files must be loaded before the other JS files, especially the config ones.

But the `assets.js` object in `config/env/all.js` makes `module.exports.getJavaScriptAssets()` in `config/config.js` load all JS assets alphabetically.

This PR fixes this and makes the JS files being loaded in this order :
1. Angular lib files
2. App config (`public/config.js`)
3. App init  (`public/application.js`)
4. Modules declarations  (`public/modules/*/*.module.js`)
5. Modules services (`public/modules/*/services/*.js`)
6. The remainder...

### DIFF
--- a/config/env/all.js
+++ b/config/env/all.js
@@ -61,6 +61,8 @@ module.exports = {
 		js: [
 			'public/config.js',
 			'public/application.js',
+            'public/modules/*/*.module.js',
+            'public/modules/*/services/*.js',
 			'public/modules/*/*.js',
 			'public/modules/*/*[!tests]*/*.js'
 		],
--- a/config/env/all.js
+++ b/config/env/all.js
@@ -1,74 +1,74 @@
 'use strict';
 
 module.exports = {
-	app: {
-		title: 'MEAN.JS',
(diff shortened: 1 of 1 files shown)


ilanbiala:
@herveguetin if you fix the pattern issue, then this PR should be good to go.

config/env/all.js
```diff
@@ -61,6 +61,8 @@ module.exports = {
 		js: [
 			'public/config.js',
 			'public/application.js',
+			'public/modules/*/*.module.js',
+			'public/modules/*/services/*.js',
 			'public/modules/*/*.js',
```
ilanbiala: This one is repetitive.

codydaig:
@herveguetin @ilanbiala 
I think this has been resolved already in the 0.4.0 branch, but not entirely sure....
https://github.com/meanjs/mean/blob/0.4.0/config/assets/default.js#L33

If so, can this be closed?

rhutchison:
Agreed

PR #297, opened 2014-12-09
Added controller init

jellekralt:
With some projects, I want to load some logic when the server starts. Because I don't want to modify the express.js file too heavily with project specific stuff, I've added globbing of the controllers and a check if that module has an 'init' function. If so, the init function will be executed when the server starts. In this way, you can load stuff when the server starts while the logic stays in the controllers.

### DIFF
--- a/config/express.js
+++ b/config/express.js
@@ -33,6 +33,14 @@ module.exports = function(db) {
 		require(path.resolve(modelPath));
 	});
 
+	// Globbing controllers files
+	config.getGlobbedFiles('./app/controllers/*.js').forEach(function(controllerPath) {
+		var controller = require(path.resolve(controllerPath));
+		if('init' in controller) {
+			controller.init(app);
+		}
+	});	
+
 	// Setting application local variables
 	app.locals.title = config.app.title;
 	app.locals.description = config.app.description;


roieki:
https://en.wikipedia.org/wiki/Bullet-point_engineering

sielay:
To init routing you load all controller modules already. Why you can't put you init code there? Meaning while you declare your controller. Just an idea? This way you will load only linked controller and (more important) you won't create all requires twice.

ilanbiala:
I would actually put all the init stuff in a file a return a promise so that you can do stuff with the info before the server starts. So if I have to check if any users exist, I would do anything necessary in the promise and handle that promise in either `server.js` or `express.js`.

config/express.js
```diff
@@ -33,6 +33,14 @@ module.exports = function(db) {
 		require(path.resolve(modelPath));
 	});
 
+	// Globbing controllers files
+	config.getGlobbedFiles('./app/controllers/*.js').forEach(function(controllerPath) {
+		var controller = require(path.resolve(controllerPath));
+		if('init' in controller) {
```
ilanbiala: Why pass in `app`? Can you show a case where it is necessary? Why not also just do `if (controller.init)`?

lirantal:
I'm closing as this didn't get any further action on.
@jellekralt you're welcome to re-open with a possible 0.4.0 PR if you think it's still relevant.

PR #303, opened 2014-12-16
Make the 404 and 500 error pages less scary for users

snez:
Displaying the server error to users is scary, make 500 and 404 pages more user friendly.

### DIFF
--- a/app/views/404.server.view.html
+++ b/app/views/404.server.view.html
@@ -1,8 +1,12 @@
 {% extends 'layout.server.view.html' %}
 
 {% block content %}
-<h1>Page Not Found</h1>
-<pre>
-	{{url}} is not a valid path.
-</pre>
+<div class="row">
+	<div class="col-xs-12">
+		<div class="container">
+			<h1>Page Not Found</h1>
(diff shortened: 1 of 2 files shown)


app/views/404.server.view.html
```diff
@@ -1,8 +1,12 @@
...
 {% block content %}
-<h1>Page Not Found</h1>
-<pre>
-	{{url}} is not a valid path.
-</pre>
+<div class="row">
+	<div class="col-xs-12">
+		<div class="container">
+			<h1>Page Not Found</h1>
+			You have entered an invalid URL.
```
ilanbiala: Can we get a "back to home" button? I think most users will click that pretty quickly.

ilanbiala:
@snez I like the idea for production, can you set up some conditional based on a flag set in `config/express.js` about the environment. That way, development environment will produce errors and be more helpful.

ilanbiala:
@snez once you make the conditional environment change, we will review it and merge it in if it's all good.

ilanbiala:
@snez are you going to reopen this in a new PR or...?

PR #305, opened 2014-12-16
Implement an application-wide error handler

snez:
Implemented and events module which can handle events emitted from anywhere in the application. Can be used to handle database errors or exceptions across the application.

### DIFF
--- a/app/modules/events.js
+++ b/app/modules/events.js
@@ -1,0 +1,25 @@
+'use strict';
+
+var formatMessage = function(message)
+{
+	if (typeof message === 'object')
+		message = '<pre>'+JSON.stringify(message, null, 4)+'</pre>';
+
+	return message;
+};
+
+// Handles node errors generated with:
+// process.emit('error', err);
+var handleError = function(err)
+{
+	console.trace(err);
+
+	// Log the error in the database?
+
+	// Email the developers?
+};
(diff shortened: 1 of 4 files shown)


app/modules/events.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+var formatMessage = function(message)
```
ilanbiala: why in a variable and not just naming the function? Don't stack traces come out with anonymous functions that way..?

app/modules/events.js
```diff
@@ -0,0 +1,26 @@
...
+{
+	if (typeof message === 'object')
```
ilanbiala: {} missing.

app/modules/events.js
```diff
@@ -0,0 +1,26 @@
...
+
+	// Log the error in the database?
+
+	// Email the developers?
+	// var emailBody = formatMessage(err);
+};
+
+module.exports = function()
+{
+	process.on('error', handleError);
```
ilanbiala: What about logging? Is this going to allow for users to add in logging easily?

app/modules/events.js
ilanbiala: Is this function used? Right now the call looks commented out.

PR #306, opened 2014-12-16
Allow the user to configure an instance of elastic search in the app

snez:
Optionally, if a config.elasticsearch is configured, we can instantiate a global object with a connection to Elastic Search to be used for indexing/querying.

### DIFF
--- a/config/env/all.js
+++ b/config/env/all.js
@@ -69,4 +69,9 @@ module.exports = {
 			'public/modules/*/tests/*.js'
 		]
 	}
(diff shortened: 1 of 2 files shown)


ilanbiala:
@snez I think elasticsearch is too specific for most people. This should probably be added to the generator @NeverOddOrEven, no?

SOSANA:
@ilanbiala generator would be great to have. Would this be also considered a "module" for 0.4.0?

ilanbiala:
@SOSANA this can be reposted as an issue on https://github.com/meanjs/generator-meanjs

PR #308, opened 2014-12-16
Implement a returnUrl service to allow users to return to pages that require login

snez:
Self explanatory:
ReturnUrl.set({ name: $state.current.name, params: $stateParams }) from a controller that requires the user to log in
ReturnUrl.return() right after signin/signup


codydaig:
This has been resolved in the 0.4.0 branch. #487 
This PR can be closed.

PR #322, opened 2014-12-23
Fix #321 image paths in social meta tags

### LINKED ISSUE #321: Path to social meta brand image is wrong
In `app/views/layout.server.view.html` the path to the social media images in meta tags are wrong (`img/brand/logo.png`), should be: `/modules/core/img/brand/logo.png`.

Already made a PR fixing this.

AndreasHassing:


ilanbiala:
Which Facebook spec are we compliant with right now? Will this need to change for API >=2.0?

NeverOddOrEven:
Shew, this is confusing. It looks like the "{{url}}" variable is present in the master branch server.view.template.html file. I'm surprised this is not showing up as a diff conflict. This can be closed those @ilanbiala. I think if there are further issues, a new issue can be posted.

PR #326, opened 2014-12-24
Update users.profile.server.controller.js

Edo78:
null isn't a valid json

### DIFF
--- a/app/controllers/users/users.profile.server.controller.js
+++ b/app/controllers/users/users.profile.server.controller.js
@@ -52,5 +52,5 @@ exports.update = function(req, res) {
  * Send User
  */
 exports.me = function(req, res) {
-	res.json(req.user || null);
+	res.json(req.user || {});
 };


Edo78:
@roieki "fail" usually doesn't mean "beaks everythin", errors and exceptions are manageable (think about the whole set of error code 4xx and 5xx) ...

@ilanbiala I'm sorry but I don't understand what you mean ... where do you want to put the check ?

roieki:
@Edo78 - I actually think it *should* break everything if a user isn't loaded... Either way, it makes little difference - you have to check whether or not you have a user loaded or not.

Edo78:
@roieki every API should return something valid even if it's called with parameters full of garbage.
The more I think about this PR the more I'm willing to change adding even an error code like 404
Returnig something invalid let the client wonder if the fault it's on itself or on the server side ...

ilanbiala:
@Edo78 on the client side (Angular) you have to check against `{}` (truthy) instead of `null` (falsy), which changes lots of things.

ilanbiala:
@Edo78 you'll also have to fix the commit message if we get everything changed accordingly and everyone thinks this is a good idea to merge.

ilanbiala:
@NeverOddOrEven do you mean `req.user = user;`? The other big comparison issue is in all the Angular code @Edo78 which like I said before:
> you have to check against {} (truthy) instead of null (falsy), which changes lots of things.

Edo78:
I'm sorry, this is my first PR on github and I'm not a mean.js expert so maybe I'm screwing something but I can't event find where in the frontend is called this code ...

lirantal:
@Edo78 would you like to re-consider this PR for 0.4.0?
@ilanbiala I think Edo has a good point in APIs returning an empty object (or error object) rather than null.

ilanbiala:
@lirantal I agreed before, but it has to be fixed on the frontend as well.

lirantal:
@ilanbiala actually, taking another look here this whole discussion is futile because I couldn't find any use in the route /api/users/me so maybe we just need to remove it as it's obsolete.

Can you confirm?

ilanbiala:
Yeah I'm not finding a use. Do you remember us using this at any point?

lirantal:
Yes, it's used in master AFAIR.
Ok well, I'll go ahead and remove it.

ilanbiala:
@lirantal we can close this then, right?

lirantal:
yeah, closing, thanks.

PR #338, opened 2014-12-31
Aligning ExpressJS 4.0 compatibility

lirantal:
replacing old expressjs 3.x built-in 'compress' middleware with standalone compression middleware

### DIFF
--- a/config/express.js
+++ b/config/express.js
@@ -11,7 +11,7 @@ var fs = require('fs'),
 	logger = require('./logger'),
 	bodyParser = require('body-parser'),
 	session = require('express-session'),
-	compress = require('compression'),
+	compression = require('compression'),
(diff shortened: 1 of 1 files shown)


lirantal:
I don't think we can make that distinction. I mean, I don't see any point in introducingthat kind of different behavior:)

ilanbiala:
`/\.min/` gives us minified files, right? Compressing .min files seems like more of a waste because they are already pretty compressed, whereas you could probably compress unminified files more because they aren't as compressed to begin with.

lirantal:
@ilanbiala I don't think it matters much if files are minified or not.
Thanks for reviewing, I'll go ahead and merge.

ilanbiala:
Sounds good, we can always optimize once everything is else as fast.

Issue #336, opened 2014-12-29
2 different views with forms for creating and updating articles - bad practice


ramusus:
I suppose it would be better to join them into one view with one article form. Otherwise there is violatation of DRY principe. And it's much more complicated handle of the same html form in 2 different places
I can make PR If you agree with my point of view.

androidfred:
I agree, this violates DRY and is bad practice, would be great if you could fix it!

ramusus:
just commited pull request

JuanCaicedo:
My post was about using a directive and controlling the method through an attribute, but this seems much neater!

lirantal:
Seems we've already done some refactoring there on the articles module.
Closing.

@ramusus @JuanCaicedo welcome to comment and add anything else if you'd like.

PR #339, opened 2015-01-01
Join 2 articles public form views into one for sake of DRY

ramusus:
Solves #336


andresmanz:
In my opinion it is extremely helpful to have best practices in demo code. Is there a reason this hasn't been merged yet?

codydaig:
@ramusus Would you mind updating this PR against the latest master branch? I'd really like to see this change myself. Unless there's any objections of moving to a dry structure? @lirantal @ilanbiala @mleanos @trainerbill @rhutchison

public/modules/articles/controllers/articles.client.controller.js
```diff
@@ -4,19 +4,21 @@ angular.module('articles').controller('ArticlesController', ['$scope', '$statePa
...
+			var success = function(response) {
 				$location.path('articles/' + response._id);
-
-				$scope.title = '';
-				$scope.content = '';
-			}, function(errorResponse) {
+				$scope.article.title = '';
+				$scope.article.content = '';
+			}
+			var error = function(errorResponse) {
```
ilanbiala: What is this doing now?

ilanbiala:
@ramusus this looks fine to me, albeit some changes that need to be made such as removing `data-` and fixing the indentation.

ilanbiala:
@ramusus would you be able to tidy up this PR so it can be merged?

ramusus:
Sorry, unfortunatly I'm very busy now

mleanos:
@ilanbiala @codydaig 

This looks like a good implementation. Since @ramusus doesn't have time for this, I'll take it over. This is related to #807 & #874 Once the latter is decided on, then I will work on this implementation.

ilanbiala:
@mleanos why do you need #874 to work on this?

mleanos:
@ilanbiala #874 is a refactor of the Article module. However, since that PR is set for 0.5.0 then this can probably move forward with consideration for both #807 and #874

codydaig:
@ilanbiala Can this be moved to 0.5.0?

ilanbiala:
mleanos do you need #874 to work on this?

mleanos:
@ilanbiala I'm not sure. I need to take another look at this implementation. I'd rather wait until the dust settles with #874 since it's a huge refactor, and this implementation might not be relevant once we implement John Papa's styleguide; I'm not exactly sure though.

Also, #807 will introduce a change in how we're allowing users to edit/create Articles, it may make sense to hold off on this; or just address these concerns in #807

rhutchison:
@ramusus I agree with the direction, but not completely with the implementation.  The controller should call $save and the service should work out whether it's an insert or update.

@codydaig @mleanos

mleanos:
@rhutchison I agree. I would like to reintroduce this with #807. Also, would prefer to add the angular styleguide with it, since the article refactor PR has stalled for so long. We need to get moving on the styleguide, so we can have it out with the 0.5.0 release.

Issue #348, opened 2015-01-08
response code 500 on accessing non-existing document?


FREEZX:
Hi,

I'm currently developing an application using mean, and something caught my attention:
https://github.com/meanjs/mean/blob/master/app/controllers/articles.server.controller.js#L93

This returns 500 internal server error if the document does not exist.
In my opinion this should send back 404 Not Found, without disclosing a stack trace.

ilanbiala:
@lirantal isn't the specific route [this one](https://github.com/meanjs/mean/blob/master/app/routes/articles.server.routes.js#L15)? That just calls articleById internally and sends the result.

lirantal:
@ilanbiala that's the route but it uses the :articleId parameter which is resolved using express's app.param() method.

ilanbiala:
@lirantal yep, that's what I said. So what's the consensus, should we at least send a 400 if the id is not a valid ObjectId? I think we should start with that and work forward.

lirantal:
merged the relevant PR so I'm closing this.

PR #349, opened 2015-01-09
fixing issue #348 - instead of returning a server error 500 on article loading which isnt found we'll throw a 404 with json message

lirantal:


ilanbiala:
@lirantal why not under `if (!article)`?

lirantal:
because the check for err is already there, and it wins if there's an error over the if (!article)

ilanbiala:
Yeah but there shouldn't be an error when no article is found, the error will only be present if there is an issue with the database...not a nonexistant document with that _id.

FREEZX:
I think that if there's really an error then it should 500, but it should 404 if there's no article in the database.
In your code:

```
if (err) {
	return res.status(404).send({
  		message: 'Article not found'
  	}); 
}
if (!article) return next(new Error('Failed to load article ' + id));
```

The opposite happens. It shows 404 if there's an error and returns an error if there's no article with the specified id in the database.

ilanbiala:
Yes, the two if statement blocks have to be switched.

lirantal:
what happens there behind the scenes is that an err object is returned because its trying to cast wrongly if I load /article/5 for example
```
Server Error

	CastError: Cast to ObjectId failed for value "5" at path "_id"
```
Also, you're quite missing the point of the two ifs, even if I switch their ordering, whichever fires first will ignore the next one. Meaning if there's an err object returned then there's definitely no article object available.

ilanbiala:
Why should there be an error object if an objectId is given but no article is found? In fact, it would make more sense to check beforehand if id can be casted to `ObjectId` to prevent unnecessary DB queries.

lirantal:
Looks like it's something internal to mongoose, here's the full stack trace:
```
Server Error

  CastError: Cast to ObjectId failed for value "5" at path "_id"
    at ObjectId.cast (fnode_modules/mongoose/lib/schema/objectid.js:116:13)
    at ObjectId.castForQuery (/lirantal-meanjs/node_modules/mongoose/lib/schema/objectid.js:165:17)
    at Query.cast (/lirantal-meanjs/node_modules/mongoose/lib/query.js:2336:32)
    at Query.findOne (/lirantal-meanjs/node_modules/mongoose/lib/query.js:1118:10)
    at Query.exec (/lirantal-meanjs/node_modules/mongoose/node_modules/mquery/lib/mquery.js:2213:16)
    at Query.exec (/lirantal-meanjs/node_modules/mongoose/lib/query.js:1797:19)
    at exports.articleByID (/lirantal-meanjs/app/controllers/articles.server.controller.js:91:55)
    at paramCallback (/lirantal-meanjs/node_modules/express/lib/router/index.js:386:7)
    at param (/lirantal-meanjs/node_modules/express/lib/router/index.js:366:5)
    at Function.proto.process_params (/lirantal-meanjs/node_modules/express/lib/router/index.js:392:3)
```

ilanbiala:
@lirantal we can check to see if it's valid before querying to save on DB queries.

lirantal:
are you referring to mongoose.Schema.Types.ObjectId() ?
it seems cumbersome to me

ilanbiala:
Not really, because it will prevent us from querying the db if the id isn't even valid, which minimizes error checking afterwards and reduces load on the db.

lirantal:
@ilanbiala updated handling for checking invalid objectids, let me know what you think.

ilanbiala:
@lirantal made some comments.

ilanbiala:
@lirantal LGTM other than the comment I made above.
> That makes sense, but we will have to fix this error for the generator, too in a different set of PRs. I think having a generic error handler ASAP will allow for good error logging, cleaner code, and better error output in the end.
Thoughts on how we do this cleanly for both the generator and this repo?

app/controllers/articles.server.controller.js
```diff
@@ -88,9 +88,20 @@ exports.list = function(req, res) {
...
+		return res.status(400).send({
+			message: 'Article is invalid'
+		});
+	}
+
 	Article.findById(id).populate('user', 'displayName').exec(function(err, article) {
 		if (err) return next(err);
-		if (!article) return next(new Error('Failed to load article ' + id));
+		if (!article) {
+			res.status(404).send({
```
ilanbiala: shouldn't we `return res.status.send` so that next() doesn't get called below? Or we could toss it into an `else {}`?

ilanbiala:
@lirantal still one issue where I think it might continue. We should probably also add tests for the invalid ObjectId and nonexistent article when given a valid ID.

FREEZX:
One thing to note:
ids in monogodb are not limited to being just ObjectId type, even though that's the most common use case.

lirantal:
@ilanbiala 
1. whether there's return or there isn't for the req.send() it won't go passed that point as it's returning the response there, but for the sake of consistency I added the return.
2. I indeed added tests last night for validating that, just didn't commit yet, good to see we're both worrying about quality and tests safety net :)

@FREEZX you're right, although we can't account for all the variations ahead.

I just want to state that while we're working on delivering this with good quality and coding convention but let's keep in mind that users aren't going to use the MEAN.JS freamework for it's amazing article management capability. The articles management is more of a PoC or boilerplate for getting started but I don't think anyone is planning on that exact capability being there.

lirantal:
I believe we addressed all the pending issues, so if no one has anything else to add I'll go ahead and merge this.

ilanbiala:
@lirantal LGTM. The next thing we should do is refactor error handling so it's cleaner.

Issue #325, opened 2014-12-24
Non-conventional use of ng-init


androidfred:
From Google group discussion: https://groups.google.com/forum/#!topic/meanjs/cO5AAAOgVAY

Angular docs for ngInit (https://docs.angularjs.org/api/ng/directive/ngInit#!) say:

"The only appropriate use of ngInit is for aliasing special properties of ngRepeat, as seen in the demo below. Besides this case, you should use controllers rather than ngInit to initialize values on a scope.

But e.g. in list-articles.client.view.html it is being used as below:

    <section data-ng-controller="ArticlesController" data-ng-init="find()">

This appears to be a part of the boilerplate that breaks convention and unless there is good reason for that it should arguably be fixed. I'd be happy to do a pull request and fix it myself, though I'm not sure if data-ng-init="find()" should simply be removed or replaced by something else.

androidfred:
In case anyone cares, take a look at the controllers, views and karma tests in this module to see how I changed things: https://github.com/androidfred/ideit/tree/master/public/modules/ideas

TLDR I split the controller, run the initialization methods after declaring them in the controllers, got rid of ng-init and changed the tests slightly.

I'd be happy to do a PR but feedback and agreement from you guys is required first.

ikhos:
I find the proposed changes better than using ng-init, but kind of cumbersome. Couldn't this be delegated to a function that checks the current $state? Will this affect performance in any way?

Like so:

```
$scope.$on('$stateChangeSuccess', function(event, toState){
  switch (toState) {
    case 'view':
      $scope.findOne();
      break;
    case 'list':
      $scope.findOne();
      break;
    default:
      //Default logic
  }
});
```

ilanbiala:
@ikhos I'm not really a fan of doing stuff in the event, because that makes more work for people who want to switch routing and if we decide to move to angular/router in the future.

androidfred:
Thanks for your attention.

As Angular docs say you should use controllers rather than ngInit to initialize values on a scope, and that a controller should contain only the business logic needed for a single view, I hope both these issues are addressed by some other means.

mleanos:
@androidfred Have you explored implementing a resolve on your ui routes? This could help keep your controllers clean of these types of data initializing. I've seen this approach advocated. Personally, I prefer having an init() function in my controllers. It's easier for me to keep track of what data is being loaded when the controller's initialize; without the need to track it down in my route configurations. 

[...]

PR #355, opened 2015-01-11
removed ng-init from views, split controller, modified karma tests to reflect the changes

androidfred:
This is my suggested refactor for this issue https://github.com/meanjs/mean/issues/325


codydaig:
Is this something that we want implemented in 0.4.0?
IMO, its a case by case basis and I think the for boiler plate, the controllers should be left in a single file.

lirantal:
@androidfred this indeed isn't going to master neither to core 0.4.0
maybe you want to reach out to @rhutchison on gitter chat to discuss more improvements to the UI.

androidfred:
Thanks for your attention.

As Angular docs say you should use controllers rather than ngInit to initialize values on a scope, and that a controller should contain only the business logic needed for a single view, I hope both these issues are addressed by some other means.

lirantal:
once we release 0.4.0 we can discuss about these kind of changes and you're welcome to offer a PR then.

PR #367, opened 2015-01-16
cleaning up css and js assets 

lirantal:
cleaning up css and js assets so they look indented and easy to read on browsers when viewing the HTML source of the page

before:
assets_clearview1

after:
assets_clearview2


ilanbiala:
LGTM.

PR #373, opened 2015-01-23
User ID middleware now uses .findById

ilanbiala:
Better use of syntactic sugar.

### DIFF
--- a/app/controllers/users/users.authorization.server.controller.js
+++ b/app/controllers/users/users.authorization.server.controller.js
@@ -11,9 +11,7 @@ var _ = require('lodash'),
  * User middleware
  */
 exports.userByID = function(req, res, next, id) {
-	User.findOne({
-		_id: id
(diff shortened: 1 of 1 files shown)


ilanbiala:
@lirantal @NeverOddOrEven simple change, let me know if I can merge.

ilanbiala:
@lirantal This is the only instance I found by the way, so this will make it more consistent and cleaner.

lirantal:
why are we replacing findById() with findOne() ?

ilanbiala:
Because it's in ID middleware, and findById is cleaner, more readable, and suited for this function.

lirantal:
It's internal to mongoose, and is just a wrapper for findOne anyway... I'm ok with the PR overall, don't have anything against.

ilanbiala:
It is a wrapper, but it's a useful one that makes it much easier to understand what the main purpose is at a glance.

PR #374, opened 2015-01-24
Infer first and last name from github and twitter response

igorauad:
I think `firstName` and `lastName` should be set when signing through providers. This fixes it for github and twitter.

### DIFF
--- a/config/strategies/github.js
+++ b/config/strategies/github.js
@@ -23,8 +23,15 @@ module.exports = function() {
 			providerData.refreshToken = refreshToken;
 
 			// Create the user OAuth profile
+			var displayName = profile.displayName.trim();
+			var iSpace = displayName.indexOf(' '); // index of the whitespace following the firstName
+			var firstName =  iSpace !== -1 ? displayName.substring(0, iSpace) : displayName;
+			var lastName = iSpace !== -1 ? displayName.substring(iSpace + 1) : '';
+
 			var providerUserProfile = {
-				displayName: profile.displayName,
+				firstName: firstName,
+				lastName: lastName,
+				displayName: displayName,
 				email: profile.emails[0].value,
 				username: profile.username,
 				provider: 'github',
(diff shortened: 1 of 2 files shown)


ilanbiala:
@igorauad once you make the changes I'll review again and merge if it is good.

igorauad:
Thanks, @ilanbiala . It supports now last name with multiple names, as you suggested.

config/strategies/github.js
```diff
@@ -23,8 +23,14 @@ module.exports = function() {
 			providerData.refreshToken = refreshToken;
 
 			// Create the user OAuth profile
+			var displayName = profile.displayName.trim();
+			var firstName =  displayName.indexOf(' ') !== -1 ? displayName.substring(0, displayName.indexOf(' ')) : displayName;
+			var lastName = displayName.indexOf(' ') !== -1 ? displayName.substring(displayName.indexOf(' ') + 1) : undefined;
```
ilanbiala: Since we're using indexOf more than once, why don't we cache that value in a variable?

ilanbiala:
Just one change to save on indexOf calls, I think for larger names it's more useful, but still 4 down to 1.

config/strategies/github.js
igorauad: You are absolutely right.

ilanbiala:
@igorauad looks great, can you squash the commits and reword the last commit? If you don't know how you can refer to this article http://davidwalsh.name/squash-commits-git.

ilanbiala:
LGTM, @lirantal any objections?

lirantal:
for lastName, isn't it preferred that we set to an empty string '', instead of undefined?

ilanbiala:
@lirantal yeah that's clearer and probably more compatible actually. @igorauad can you make the change @lirantal suggested and squash it to 1 commit again?

igorauad:
Sure, @ilanbiala @lirantal

ilanbiala:
Thanks @igorauad for the good PR. Keep 'em coming.

igorauad:
Thank you, @ilanbiala !

PR #376, opened 2015-01-25
Correctly encode and decode password salt (0.4.0)

rmuch:
**This pull request merges the fix proposed in [pull/253](https://github.com/meanjs/mean/pull/253) into the 0.4.0 branch.**

The user password salt should be encoded with Base64 before being saved
to the database.

The current code adds an unecessary step of converting the result of
crypto.randomBytes() (which already returns a SlowBuffer) to a Base64
string and back again to a Buffer, and misses the final step of
converting the Buffer's bytes back to a Base64 string.

Because of this, the salt stored in the database is garbled. This is
inconvenient when manipulating the data in a terminal or text editor.

When generating the password hash, the crypto.pbkdf2Sync() method
creates a new Buffer directly from the data supplied. Due to the
incorrect encoding of the salt, entropy is lost at this step,
weakening the security of stored passwords against brute force attacks.

### DIFF
--- a/modules/users/server/models/user.server.model.js
+++ b/modules/users/server/models/user.server.model.js
@@ -100,7 +100,7 @@ var UserSchema = new Schema({
  */
 UserSchema.pre('save', function(next) {
 	if (this.password && this.password.length > 6) {
-		this.salt = new Buffer(crypto.randomBytes(16).toString('base64'), 'base64');
+		this.salt = crypto.randomBytes(16).toString('base64');
 		this.password = this.hashPassword(this.password);
 	}
 
@@ -112,7 +112,7 @@ UserSchema.pre('save', function(next) {
  */
 UserSchema.methods.hashPassword = function(password) {
 	if (this.salt && password) {
-		return crypto.pbkdf2Sync(password, this.salt, 10000, 64).toString('base64');
+		return crypto.pbkdf2Sync(password, new Buffer(this.salt, 'base64'), 10000, 64).toString('base64');
 	} else {
 		return password;
 	}


modules/users/server/models/user.server.model.js
```diff
@@ -112,7 +112,7 @@ UserSchema.pre('save', function(next) {
  */
 UserSchema.methods.hashPassword = function(password) {
 	if (this.salt && password) {
-		return crypto.pbkdf2Sync(password, this.salt, 10000, 64).toString('base64');
+		return crypto.pbkdf2Sync(password, new Buffer(this.salt, 'base64'), 10000, 64).toString('base64');
```
ilanbiala: Re-instantiating the buffer seems unnecessary when we could pass in L183 without the `.toString('base64')`, no?

rmuch: @ilanbiala It is necessary to be explicit about encoding and decoding the buffer using base64 in both cases.

I'm not worried about the overhead of re-instantiating the buffer here, because crypto.pbkdf2Sync() will create a buffer anyway if you pass a string as the salt, but it will make the incorrect assumption that the string contains binary-encoded data.

The buffer will always need to be re-instantiated somehow, because the setter created by the Mongoose model would automatically convert the buffer to a string when it's assigned to this.salt on L103.

ilanbiala: The code could be:
```js
UserSchema.methods.hashPassword = function(password, saltBuffer) {
  if (saltBuffer) {
    return crypto.pbkdf2Sync(password, saltBuffer, 1000, 64).toString('base64');
  } else if (this.salt) {
     return crypto.pbkdf2Sync(password, new Buffer(this.salt, 'base64'), 1000, 64).toString('base64');
  } else {
    return password;
  }
};
```

Wouldn't not re-instantiating be much more performant for long passwords? Can you set up some sort of JSPerf or benchmark to figure out whether we should do it or not?

zlatinejc:
Hi guys.
Just an observation:
You could put a checker in the login flow so you can grab the password and encrypt it using the proposed "better entropy" encription etc...
?

I am guessing you would need an additional field in the model to check that the pass has been moved to the newer version.

My humble opinion:
Better than notifying customers of "security upgrade" which they thought it was already secure. I am not into this black/white code cutting ways...

Best!

ilanbiala:
@rmuch can you rebase this PR and squash the commits?

rmuch:
@ilanbiala Done, hope it's worked OK. Please let me know if anything is wrong with the PR.

ilanbiala:
@lirantal this one is the exact same content as the PR for master, so if it LGTY, you can merge.

PR #381, opened 2015-01-30
Fix .editorconfig file bugs

treyhunner:
The only commit that is absolutely necessary from this pull request is c3ffb6a55ee17274973881d6adfaef482c75d134.  The comma and space syntax that was included before for separating filename globs is invalid.

The other three commits are just suggested style improvements.


.editorconfig
```diff
@@ -1,18 +1,17 @@
...
-# Howto with your editor:
-#   Sublime: https://github.com/sindresorhus/editorconfig-sublime
+# How-to with your editor: http://editorconfig.org/#download
 
 # top-most EditorConfig file
 root = true
 
 # Unix-style newlines with a newline ending every file
-[**]
+[*]
```
ilanbiala: Why 1 asterisk instead of 2?

.editorconfig
```diff
@@ -23,14 +22,14 @@ space_after_anonymous_functions = false
...
 indent_style = tab
 
 # No standard.  Please document a standard if different from .js
-[**.md]
+[*.md]
 indent_style = tab
 
-# Standard at: 
+# Standard at:
 [Makefile]
```
ilanbiala: Can you clean this up, we don't have a Makefile. Maybe set some standards for the Dockerfile and Procfile.

.editorconfig
treyhunner: When no `/` is present in the glob, it will apply to files in all subdirectories, so `**` would be unecessary.

.editorconfig
treyhunner: I removed the `Makefile` rule.  I didn't see any indentation in the `Dockerfile` or `Procfile`.  It looks `tab` indentation is meant to be used in all files.  `indent_style = tab` could be added to the `*` rule to apply to these files and all others if that's the case.

.editorconfig
```diff
@@ -23,14 +22,10 @@ space_after_anonymous_functions = false
...
+[{*.yml,*.html,*.css}]
 trim_trailing_whitespace = true
 indent_style = tab
 
 # No standard.  Please document a standard if different from .js
-[**.md]
-indent_style = tab
-
-# Standard at: 
-[Makefile]
```
ilanbiala: Can you add in settings for the Dockerfile and Procfile?

.editorconfig
```diff
@@ -23,14 +22,10 @@ space_after_anonymous_functions = false
...
+[*.md]
 indent_style = tab
```
ilanbiala: Shouldn't there be a newline at the end of this file?

ilanbiala:
@treyhunner if you fix up this PR, we can get it merged in.

ilanbiala:
@treyhunner can you apply the fixes I commented about above?

.editorconfig
treyhunner: There is a newline.  If there wasn't it would indicate it with a little icon in the diff.

treyhunner:
@ilanbiala this should be ready for merging.  I fixed indentation and newlines in lots of files and consolidated `.editorconfig` rules.

.editorconfig
```diff
@@ -1,36 +1,25 @@
...
+# How-to with your editor: http://editorconfig.org/#download
 
 # top-most EditorConfig file
 root = true
 
 # Unix-style newlines with a newline ending every file
-[**]
+[*]
 end_of_line = lf
+indent_style = tab
```
ilanbiala: Does anyone object to tabs as the indent style?

ilanbiala:
@treyhunner can you squash the commits?

treyhunner:
I squashed everything into two commits: one to update the `.editorconfig` file and one to add missing newlines to ends of files

ilanbiala:
LGTM. @lirantal LGTY?

lirantal:
I'm not really thrilled with indent_style tabs, and would rather get spaces with a size of 2. I know we tried to vote on that but don't think there are conclusive answers. With this PR we still keep the original indentation style of tabs, so at least we're not breaking anything.

Other than that, if it LGTY go ahead and merge.

ilanbiala:
@lirantal I know about the indentation issue, but since it hasn't caused anything to break and no one really has swayed the vote, I don't see any reason to change the indention in every file and have lots of pointless commits right now.

PR #382, opened 2015-01-30
Adjust profile image dimensions for Fb and Twitter

igorauad:
@ilanbiala please, let me know where I should be suggesting these PRs. Shall I do it for 0.4.0, master or both?

Regarding image sizes, I just faced a need for larger dimensions, so I changed the profileImageUrl for FB and Twitter. Thought perhaps this should be the default. What do you think?

### DIFF
--- a/modules/users/server/config/strategies/facebook.js
+++ b/modules/users/server/config/strategies/facebook.js
@@ -29,7 +29,7 @@ module.exports = function(config) {
 				lastName: profile.name.familyName,
 				displayName: profile.displayName,
 				email: profile.emails[0].value,
-				profileImageURL: (profile.photos && profile.photos.length) ? profile.photos[0].value : undefined,
+				profileImageURL: (profile.id) ? '//graph.facebook.com/' + profile.id + '/picture?type=large' : undefined,
 				provider: 'facebook',
 				providerIdentifierField: 'id',
 				providerData: providerData
--- a/modules/users/server/config/strategies/twitter.js
+++ b/modules/users/server/config/strategies/twitter.js
@@ -26,7 +26,7 @@ module.exports = function(config) {
 			var providerUserProfile = {
 				displayName: profile.displayName,
 				username: profile.username,
-				profileImageURL: (profile.photos && profile.photos.length) ? profile.photos[0].value : undefined,
+				profileImageURL: profile.photos[0].value.replace('normal', 'bigger'),
 				provider: 'twitter',
 				providerIdentifierField: 'id_str',
 				providerData: providerData


ilanbiala:
@igorauad 0.4.0 is preferred so they get merged into the new version. I'm okay with these changes, but even though it's so little I'd prefer the username and image changes to be in 2 separate PRs just because they are pretty different things.

Also, is there any way to test these OAuth flows and configs? I couldn't think of a good way, but we should really be testing these just like we test our own API because changes may happen without us knowing and the tests would let us know.

igorauad:
You are right, @ilanbiala ! I will separate them. This one has only the image changes now.

Testing OAuth would be interesting, I don't know an easy way of testing either. I have tested for Fb here I realized it was not working due to `username`.

Maybe at least a standard `.env` file being loaded by `grunt-env` or a gulp correspondent would help.

ilanbiala:
@lirantal @NeverOddOrEven looks good?

modules/users/server/config/strategies/facebook.js
```diff
@@ -30,7 +30,7 @@ module.exports = function(config) {
 				displayName: profile.displayName,
 				email: profile.emails[0].value,
 				username: profile.username,
-				profileImageURL: (profile.photos && profile.photos.length) ? profile.photos[0].value : undefined,
+				profileImageURL: (profile.id) ? 'https:graph.facebook.com/' + profile.id + '/picture?type=large' : undefined,
```
lirantal: is it actually https:graph or is that a typo and should read really https://graph...?

lirantal:
@igorauad is the change there means that we won't be storing the profile image on the meanjs project but rather take it from facebook using the URL there? If so, then websites which are enabling http:// and not SSL might have browsers block SSL communication (and vice-versa) due to security issues

simison:
Please use `://graph...` instead of `https://graph...` and it'll work with both.

igorauad:
@lirantal we were not storing the profile image, but rather storing another URL from `https://scontent-b.xx.fbcdn.net`. I read somewhere that it is better to use the Graph API instead of relying on facebook CDN, so I changed it. Also because it is apparently easier to play with the image size through the Graph API.

@lirantal @simison , so, shall I change to //graph? This would work with `ng-src` also?

modules/users/server/config/strategies/facebook.js
igorauad: It seems to be a typo, but I am actually using it this way and it is working. I will change it as soon as we decide whether `//graph` should be used.

lirantal:
@igorauad I'm unable to test, but please do change it to //graph and test it to say if it works or not before committing.

igorauad:
@lirantal I've tested. The protocol-relative URL works.

modules/users/server/config/strategies/twitter.js
```diff
@@ -26,7 +26,7 @@ module.exports = function(config) {
 			var providerUserProfile = {
 				displayName: profile.displayName,
 				username: profile.username,
-				profileImageURL: (profile.photos && profile.photos.length) ? profile.photos[0].value : undefined,
+				profileImageURL: (profile.photos && profile.photos.length) ? profile.photos[0].value.replace('normal', 'bigger') : undefined,
```
ilanbiala: Is there a case where a user won't have a photo that you have to check that it exists first? I would think even the bird would be the first profile picture, no? And is [0] the latest or the first?

jloveland: I think FB and Twitter would return a default image if the user hasn't uploaded one. I haven't tested this though. 

https://developers.facebook.com/docs/graph-api/reference/v2.2/user/picture

Looks like is_silhouette can be used to detect the if the default picture been replaced or not.

ilanbiala: Right, but is that in the array? If it is, then we don't need to check the length and existence.

igorauad: Yes, it is in the array, @ilanbiala . I've just tested here. Twitter returns 
```
[ { value: 'https://abs.twimg.com/sticky/default_profile_images/default_profile_0_normal.png' } ]
```

ilanbiala:
LGTM, @lirantal any suggestions?

lirantal:
@ilanbiala I'm ok with it.

PR #384, opened 2015-01-30
Remove username from facebook strategy

igorauad:
Username is now deprecated on Facebook API 2.0, as discussed in #280

### DIFF
--- a/modules/users/server/config/strategies/facebook.js
+++ b/modules/users/server/config/strategies/facebook.js
@@ -14,7 +14,7 @@ module.exports = function(config) {
 			clientID: config.facebook.clientID,
 			clientSecret: config.facebook.clientSecret,
 			callbackURL: config.facebook.callbackURL,
-			profileFields: ['id', 'name', 'displayName', 'email', 'username', 'photos'],
+			profileFields: ['id', 'name', 'displayName', 'emails', 'photos'],
 			passReqToCallback: true
 		},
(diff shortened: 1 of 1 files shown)


ilanbiala:
@igorauad does this already work? Does the URL not have an API version or something like that?

igorauad:
This works already, @ilanbiala . I think it is possible to choose the API version with the URL, but I`m not sure which one is the default. I would have to check later with more time.

Perhaps this helps:
https://github.com/jaredhanson/passport-facebook/pull/86

ilanbiala:
@lirantal @NeverOddOrEven any objections?

lirantal:
@ilanbiala sure, but @igorauad needs to confirm this has gone through his testing.
@igorauad do we need to update the docs somehow about this? what about config URLs to use the new facebook URL for that?

I also noticed we changed 'email' field with 'emails' which is already used in the profile.emails[] code on the callback so it looks ok but please confirm.

igorauad:
Thanks, @lirantal . Yes, this has gone through testing. It works. 

I guess the facebook URL depends on `passport-facebook`, not on `meanjs`, if I have understood you right. I don't think the docs should be updated, it all works the same way. 

I changed to `emails` because it was being used in the callback, but I realized both way works after testing here.

lirantal:
Alright, thanks for the update.

ilanbiala:
@lirantal LGTY?

lirantal:
Yep.

Issue #390, opened 2015-02-04
Add more control fields for articles


miguelcoba:
Change the user and created fields in the article model for new fields that track who and when the article was created or updated.

Should add the following fields:

updatedAt, default value: Date.now
updatedBy, default value: req.user
createdAt, default value: Date.now
createdBy, default value: req.user

ilanbiala:
The PR for this issue was merged in, should be resolved.

britztopher:
@ilanbiala @miguelcoba Any idea why this ticket has open closed so many times?  Anything I can help with?

ilanbiala:
This issue has only been closed then reopened once, it references lots of other things though. I'd like to merge this in, but I think @miguelcoba had some issues with redoing this PR on 0.4.0 instead of master based on the latest comment in #392.

britztopher:
@ilanbiala i see now, I saw 'closed' above on #391, thinking it was this ticket.  My B on not looking more closely.  Thanks, I wont make this mistake twice

ilanbiala:
@miguelcoba since we haven't done much about it, I'm going to close this PR unless people express strong interest in bringing it back or you can get it into 0.4.x once we have released 0.4.0

PR #392, opened 2015-02-04
Adds more control fields to article model

miguelcoba:
Fixes https://github.com/meanjs/mean/issues/390

### DIFF
--- a/app/controllers/articles.server.controller.js
+++ b/app/controllers/articles.server.controller.js
@@ -12,8 +12,17 @@ var mongoose = require('mongoose'),
  * Create a article
  */
 exports.create = function(req, res) {
+
+	// Filters out control fields
+	delete req.body.updatedAt;
+	delete req.body.updatedBy;
+	delete req.body.createdAt;
+	delete req.body.createdBy;
+
 	var article = new Article(req.body);
-	article.user = req.user;
+
+	article.createdBy = req.user;
+	article.updatedBy = req.user;
 
 	article.save(function(err) {
 		if (err) {
@@ -37,9 +46,19 @@ exports.read = function(req, res) {
  * Update a article
  */
 exports.update = function(req, res) {
+
+	// Filters out control fields
+	delete req.body.updatedAt;
+	delete req.body.updatedBy;
+	delete req.body.createdAt;
+	delete req.body.createdBy;
+
 	var article = req.article;
 
 	article = _.extend(article, req.body);
+
+	article.updatedAt = Date.now();
+	article.updatedBy = req.user;
 
 	article.save(function(err) {
 		if (err) {
@@ -73,7 +92,7 @@ exports.delete = function(req, res) {
  * List of Articles
  */
 exports.list = function(req, res) {
-	Article.find().sort('-created').populate('user', 'displayName').exec(function(err, articles) {
+	Article.find().sort('-createdAt').populate('createdBy', 'displayName').populate('updatedBy', 'displayName').exec(function(err, articles) {
 		if (err) {
 			return res.status(400).send({
 				message: errorHandler.getErrorMessage(err)
@@ -95,7 +114,7 @@ exports.articleByID = function(req, res, next, id) {
 		});
 	}
 
-	Article.findById(id).populate('user', 'displayName').exec(function(err, article) {
+	Article.findById(id).populate('createdBy', 'displayName').populate('updatedBy', 'displayName').exec(function(err, article) {
 		if (err) return next(err);
 		if (!article) {
 			return res.status(404).send({
@@ -111,7 +130,7 @@ exports.articleByID = function(req, res, next, id) {
  * Article authorization middleware
  */
 exports.hasAuthorization = function(req, res, next) {
-	if (req.article.user.id !== req.user.id) {
+	if (req.article.createdBy.id !== req.user.id) {
 		return res.status(403).send({
 			message: 'User is not authorized'
 		});
--- a/app/models/article.server.model.js
+++ b/app/models/article.server.model.js
@@ -10,10 +10,6 @@ var mongoose = require('mongoose'),
  * Article Schema
  */
 var ArticleSchema = new Schema({
-	created: {
-		type: Date,
-		default: Date.now
-	},
 	title: {
 		type: String,
 		default: '',
@@ -25,7 +21,19 @@ var ArticleSchema = new Schema({
 		default: '',
 		trim: true
 	},
-	user: {
+	updatedAt: {
+		type: Date,
+		default: Date.now
+	},
+	updatedBy: {
+		type: Schema.ObjectId,
+		ref: 'User'
+	},
+	createdAt: {
+		type: Date,
+		default: Date.now
+	},
+	createdBy: {
 		type: Schema.ObjectId,
 		ref: 'User'
 	}
--- a/app/tests/article.server.model.test.js
+++ b/app/tests/article.server.model.test.js
@@ -31,7 +31,8 @@ describe('Article Model Unit Tests:', function() {
 			article = new Article({
 				title: 'Article Title',
 				content: 'Article Content',
-				user: user
+				updatedBy: user,
+				createdBy: user
 			});
 
 			done();
--- a/app/tests/article.server.routes.test.js
+++ b/app/tests/article.server.routes.test.js
@@ -75,7 +75,8 @@ describe('Article CRUD tests', function() {
 								var articles = articlesGetRes.body;
 
 								// Set assertions
-								(articles[0].user._id).should.equal(userId);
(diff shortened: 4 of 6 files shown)


app/controllers/articles.server.controller.js
```diff
@@ -12,8 +12,17 @@ var mongoose = require('mongoose'),
...
+	// Filters out control fields
+	delete req.body.updatedAt;
+	delete req.body.updatedBy;
+	delete req.body.createdAt;
+	delete req.body.createdBy;
+
 	var article = new Article(req.body);
-	article.user = req.user;
+
+	article.createdBy = req.user;
```
ilanbiala: Why `delete` when you can just set it and overwrite?

miguelcoba: Just in case some malicious user tries to send that info in the request payload trying to overwrite some private fields in db. If I understand correctly, the new Article(req.body) will try to use any property in body that matches a property in Article, and someone could use this to overwrite fields that are not meant to be overwritten.

lirantal:
Overall looks like a good idea. @ilanbiala what do you think?

app/controllers/articles.server.controller.js
ilanbiala: That's fine for the ones that we aren't setting, but I'd rather just set them to req.user to begin with than overwrite them.

app/models/article.server.model.js
```diff
@@ -25,7 +21,19 @@ var ArticleSchema = new Schema({
...
-	user: {
+	updatedAt: {
+		type: Date,
+		default: Date.now
+	},
+	updatedBy: {
+		type: Schema.ObjectId,
+		ref: 'User'
+	},
+	createdAt: {
```
ilanbiala: why `createdAt`/`updatedAt` and not just `created`/`updated`? Those seem pretty clear to me when you also have the `..By` fields.

ilanbiala:
@lirantal I think it's fine with some changes to be made, but this should probably merge into 0.4.0, no?

app/models/article.server.model.js
miguelcoba: I just went for uniformity, as the updated/created could also be used for booleans. The At suffix implies time:

article.updated: true
or 
if (article.updated) {}

vs.

article.updatedAt = Date.now();
or
if(article.updatedAt > someOtherDate) {}

If you want I can change them, but I plan to update the generators so that every new model has those fields, and in that case by using the At and By suffixes, we also avoid name collisions if the developer want to use created or updated fields for other things (as booleans for example).

What do you think?

app/controllers/articles.server.controller.js
miguelcoba: I'm afraid I don't understand. 
You mean, to state the values of updatedBy, updatedAt in the articleByID method, before setting the req.article?
I think that could have some problems, because, as I understand it, the intention of the code is to give the middleware a pristine copy of the article, as it is in the db, so that the middleware can do anything with it (like update it or just use it to get its info for other calculations). If we update the fields on loading, then the middleware won't have a pristine copy of the data in db and could not do some operations like, for example, checking if has not been modified.
Did I understand you correctly?

What I tried to do is to only discard dangerous or control fields from the request body, when they are used to update the db object in an automated way (like the _extend(article, req.body)) as that could have some unintended effects or security implications. Other than that, I don't modify the request payload at all.

ilanbiala: Change this:
```
delete req.body.updatedAt;
delete req.body.updatedBy;
delete req.body.createdAt;
delete req.body.createdBy;
var article = new Article(req.body);
article.user = req.user;
article.createdBy = req.user;
```
to this:
```
delete req.body.updatedAt;
delete req.body.updatedBy;
delete req.body.createdAt;
article.createdBy = req.user;
article.user = req.user;
var article = new Article(req.body);
```
Doesn't it make more sense to not delete then set, and also why not assign and set up everything first, then just save the new article, don't modify it.

app/models/article.server.model.js
ilanbiala: That's fine, I just wanted to make sure there was a valid reason to not do it.

app/controllers/articles.server.controller.js
miguelcoba: This will not work:

delete req.body.updatedAt;
delete req.body.updatedBy;
delete req.body.createdAt;
article.createdBy = req.user;
article.user = req.user;
var article = new Article(req.body);

as the new Article will create a new object that will substitute the old prepared article var. So I think you meant:
delete req.body.updatedAt;
delete req.body.updatedBy;
delete req.body.createdAt;
req.body.createdBy = req.user;
var article = new Article(req.body);

that will work indeed. My only hesitation about this change is that it kinda breaks the logic I tried to convey:

someFunction(inputObject) {
  // 1st part: some logic to clean the input, very generic, not specific to someFunction real work

  // 2nd part: the real work that someFunction must do:
}

In this case, the deletion on req.body properties is the 1st part, some generic logic to clean/check security:

        // Filters out control fields
	delete req.body.updatedAt;
	delete req.body.updatedBy;
	delete req.body.createdAt;
	delete req.body.createdBy;

then comes the second part, the one is specific to this function, the creation of a new Article:

	var article = new Article(req.body);

with the required logic to build a new Article, in this case, to ensure that we record the creator of the article:

	article.createdBy = req.user;
	article.updatedBy = req.user;

Also, I think adding data to the req.body object is going to have side effects if other middleware were to use the same req.body object later in the call chain.
I realize that the same could be said about the deletion of the fields I propose, I suppose, and I would agree on that and the only thing I could say is that those fields are internal to the app and could never be a valid parameter on the request payload or that we should be explicit in the properties of req.body that we'll use for the article creation ;). 
Something along the lines 
var article = new Article({
   firstName: req.body.firstName,
   lastName: req.body.lastName
});

would be the ultimate (but tedious) secure way to build an object using data from the user without fears of security implications or unintended side effects.

I hope this clarifies my implementation.

lirantal:
@ilanbiala yeah I didn't code review it but overall looks like a good addition.
we should probably merge with 0.4, but I have considerable less time on my hands now, and we need to get the 0.4 branch out...

miguelcoba:
Any update on merging this? I have rebased it to current master branch.

ilanbiala:
Can you keep some of the unrelated changes out of this PR like spacing, comments, script loading modification, etc.?

miguelcoba:
Done. Only changes related to control fields remain.

ilanbiala:
@lirantal LGTM, but we should have this in 0.4.0 right?

ilanbiala:
LGTM now, @lirantal LGTY? @miguelcoba can you rebase so the conflicts go away?

miguelcoba:
Done. Rebased to current mean/master branch

lirantal:
@miguelcoba can check if this cleanly gets merged into 0.4 branch? I hate to ask but otherwise we need this merged there without conflicts.

miguelcoba:
I created the branch from master and the PR is against master. Should the
PR be against 0.4 branch then?

lirantal:
We'd appreciate it, yeah.

ilanbiala:
@miguelcoba can you take this exact branch (issue-390) and make a PR for 0.4.0? Merging that in will make merging 0.4.0 in much easier.

miguelcoba:
@lirantal, @ilanbiala, I tried but there are a lot of conflicts when trying to cherry-pick this change to a new branch from 0.4.0 branch.
In fact, comparing master with 0.4.0 they are so much different that they appear two different apps.
What is the goal of 0.4.0? I see there a lot of commits from 2014 and even a chat module.
Also, the directory structure is very distinct and several key files are completely different to the point of not seeing the relation between 0.4.0 and master architectures.
Are you sure that 0.4.0 is going to be merged to master? Because a lot of changes in master are not in 0.4.0 and if master is more updated/current than 0.4 I don't see why there is a 0.4.0 branch.
IMHO the better code is in master already and a lot of issues and PR are in risk of getting abandoned if 0.4.0 is going to be used as a base for future development.
At least for me, looking in the 0.4.0 code, there is no interest in that code base. master, on the other hand, is a lot more clean, updated, and useful the way it is currently.

Just a suggestion: 0.4.0 should be discarded, and this project should focus on master code and PRs before the already diminishing contributions exhaust and this project lost the appeal it began with. At least from outside, it seems that the original developers have already abandoned this code.

lirantal:
@miguelcoba back on this - yes 0.4.0 is going to be merged into master, we're making good progress there already.

I am interested in merging this in, but we need a new PR against 0.4.0 branch. can you work it out? there's not a lot of work here to be done.

ilanbiala:
@miguelcoba since this isn't against 0.4.0 we're going to have to hold off. If you can create one for 0.4.x that would be great.

PR #396, opened 2015-02-07
Remove dist files 0.4.0

sylvainlap:
Dist files should not be in the repository, but should be created locally thanks to "grunt build".


ilanbiala:
LGTM, @lirantal @Shrizzy LGTY? Only thing is that .gitignore in 0.4.0 is worse than in master, so we'll have to keep that in mind.

lirantal:
LGTM

PR #408, opened 2015-02-13
Output one JS file per line

berendt:


lirantal:
well that's odd... because I issued a PR exactly for that a while ago.

app/views/layout.server.view.html
```diff
@@ -58,8 +58,9 @@
 	</script>
 
 	<!--Application JavaScript Files-->
-	{% for jsFile in jsFiles %}
-		<script type="text/javascript" src="{{jsFile}}"></script>{% endfor %} 
+	{%- for jsFile in jsFiles %}
+	<script type="text/javascript" src="{{jsFile}}"></script>
```
ilanbiala: I'd prefer to have the tag indented.

berendt:
Already fixed. I first worked with version `0.3` and proposed this PR without checking version `0.4` and `master.`

PR #411, opened 2015-02-16
Remove executable bit when not necessary

berendt:
The executable bit is set for a lot of files where it is not necessary
to have the executable bit set. This PR removes the executable bit from
those files.


ilanbiala:
Does this actually have any effect? Most of those can't be run anyway, right?

berendt:
This is a cleanup. It looks ugly and wrong to have a `robots.txt` file or a `package.json` file with a set executable script.

ilanbiala:
What about server.js, doesn't `node` need to be able to execute that?

berendt:
It is not necessary to set `+x` for the `server.js` file. Please have a look at  http://en.wikipedia.org/wiki/Modes_(Unix).

ilanbiala:
@lirantal LGTY?

lirantal:
Hey, 

Definitely agree with @berendt. I saw this coming in on e-mail earlier today and wanted to approve but I'm only getting to working from home now.

It's a good cleanup, I'll merge.

Issue #68, opened 2014-05-18
Proposal with working code:  separate core from boilerplate.  Adopt contrib module system.


MichaelJCole:
Hi,

This is a proposal to add a core and plugin system to MEAN.JS

My motivation for this is:

1) To get an awesome clean boilerplate for new projects
2) Get updates to core code easily through npm
3) Get new features through modules from an open and thriving community
# The TLDR;

Here is a MEAN.JS boilerplate app using a new core and contrib plugin system implemented via npm.

To see the code in action:

```
git clone git@github.com:MichaelJCole/mean.git
cd mean
git checkout meanjs-core
npm install
grunt build
grunt
```

Notice package.json has two modules: meanjs-core and meanjs-module

In action:
1. Signup a new user and you should get an email.
2. Click [a confirmation link](http://localhost:3000/auth/confirm/adsfasdfasdf) and check your server console.
# Proposal

The proposal is to adopt a framework for boilerplate, core, and contrib code.

To move forward, I propose we: 

1) start to refactor core code out of the boilerplate and into meanjs-core
2) add new features (email verification and lost password) to meanjs-core
3) refine the boilerplate for the best end-developer experience
4) experiment with contrib modules and configuration, propose some best practices
5) adjust core best practices as we go

Please send any suggestions or comments!  What do you think?
# The details

I wanted to add three features to my MEAN.JS app:
1. email verification
2. lost passwords
3. subscription to an email list at MailChimp on signup.

1) & 2) seemed like 'core' functionality.  Something MEAN.JS would probably want to maintain 

3) is very custom.  Probably not maintained by MEAN.JS, but something I could share with other developers.

I created a plugin system in the MEAN.JS boilerplate forked [here](https://github.com/MichaelJCole/mean) that uses it's package.json to import modules [meanjs-core](https://github.com/MichaelJCole/mean-core) and [meanjs-module](https://github.com/MichaelJCole/mean-module).
## In action:  Signup email

1) Sign up as a new user.  An email should be sent to the address you used.

This email came from Mandrill email service account (I'll change the keys shortly).
You can config your own email transport in config/env/development.js 'emailTransactional'

The core function sendConfirmEmail is called from the boilerplate in app/controller/users.server.controller.js:

```
/**
 * Signup
 */
exports.signup = function(req, res) {
  // ...  

  // Then save the user 
  user.save(function(err) {
    if (err) {
      // ...
    } else {
      // ...
      req.login(user, function(err) {
        // ...
      });

[...]

PR #412, opened 2015-02-16
Support publishing RabbitMQ msgs via global.publishEvent

ppotoplyak:
Hi,

This patch would allow publishing RabbitMQ msgs via global.publishEvent(msg) when USE_RABBIT = true.

Users who do not care about RabbitMQ would not have their workflow changed in any way.

I am not crazy about polluting the global namespace but I think the ability to publish msgs from anywhere is very attractive.

A msg is published when node starts:
global.publishEvent({
				type : 'node started',
				version : process.versions.v8,
				filename : __filename
			});

A msg subscriber for instance can check the node version and make sure that no known security vulnerabilities exist.

Other use cases include publishing all errors, state transitions, or anything else that is potentially interesting to other pieces in your macro system.

Python example of pulling out the msg:
#!/usr/bin/env python
import pika
RABBIT_HOST = 'localhost'
RABBIT_QUEUE = 'meanjs'

connection = pika.BlockingConnection(pika.ConnectionParameters(host = RABBIT_HOST))
channel = connection.channel()

channel.queue_declare(queue = RABBIT_QUEUE)

print ' [*] Waiting for messages. To exit press CTRL+C'

def callback(ch, method, properties, body):
    print " [x] Received %r" % (body,)

channel.basic_consume(callback,
                      queue=RABBIT_QUEUE,
                      no_ack=True)

channel.start_consuming()

For all this to work, RabbitMQ should be installed on the host.

I can add docs, examples, etc..

Thanks,
-pavel

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -8,6 +8,10 @@ module.exports = {
 			pass: ''
 		}
 	},
+	rabbit: {
+		hostname: 'localhost',
+		queue: 'meanjs'
+	},
 	log: {
 		// Can specify one of 'combined', 'common', 'dev', 'short', 'tiny'
 		format: 'dev',
--- a/config/express.js
+++ b/config/express.js
@@ -28,6 +28,22 @@ module.exports = function(db) {
 	// Initialize express app
 	var app = express();
 
+	if (process.env.USE_RABBIT === 'true') {
+		var amqp = require('amqp');
+		var rabbitConn = amqp.createConnection({ host: config.rabbit.hostname });
+		rabbitConn.on('ready', function() {
+			var exchange = rabbitConn.exchange('');
+			global.publishEvent = function(eventMsg) {
+				exchange.publish(config.rabbit.queue, eventMsg);
+			};
+			global.publishEvent({
+				type : 'node started',
+				version : process.versions.v8,
+				filename : __filename
+			});
+		});
+	}
+
 	// Globbing model files
 	config.getGlobbedFiles('./app/models/**/*.js').forEach(function(modelPath) {
 		require(path.resolve(modelPath));
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
 		"glob": "~4.0.5",
 		"async": "~0.9.0",
 		"nodemailer": "~1.3.0",
-		"chalk": "~0.5"
+		"chalk": "~0.5",
+		"amqp": "^0.2.0"
 	},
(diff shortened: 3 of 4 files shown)


ilanbiala:
@ppotoplyak this isn't something that needs to be in core MEAN.JS because not everyone will need it.

ppotoplyak:
@ilanbiala I agree that it does need to be in core MEAN.JS but is that the sole criteria for patches?

My only argument is that this small patch provides a lot of capability.  However, I do get your concern that many people like me might have lots of small patches that needlessly bloat the system.

ilanbiala:
@ppotoplyak but it pollutes the code and for 99.9% of people is just a waste of bytes on their computer.

ppotoplyak:
In my mind, it would be ideal if there was baked in publisher/subscriber support in whatever core framework I am using.  This would differentiate the framework as opposed to being a "waste of bytes".  Perhaps I am just one of the 00.1 % which you describe.

ilanbiala:
@ppotoplyak also most people want to use a specific pubsub and we can't include support for all of them...

jloveland:
I agree that this is too opinionated and not needed in the core. However, there should be a way people can add contrib modules to meanjs. There was a proposal in issue #68 to add a contrib module system, but the thread died. Does anyone know what happened to this? @amoshaviv @roieki? 

I do like the concept of having a message queue to be able to publish and subscribe to events. It would be nice to have a module to provide a pub/sub implementation. The concept I would lean towards is make it pluggable for [amqp](https://github.com/postwait/node-amqp), [redis](https://github.com/mranney/node_redis), [kafka](https://github.com/SOHU-Co/kafka-node), etc.

ilanbiala:
@jloveland a contrib module system would be nice, but how do we keep it clean yet pluggable? It would probably just be a plugin like passport or body-parser where you just do `app.use(passport.initialize())`, no?

max-mog:
Having RabbitMQ as a part of the code but turned off by default would make the app much more robust as now other services can be hooked up via pub/sub. +1

ilanbiala:
@max-mog but for the majority of people it's just unnecessary code like the socket setup in 0.4 will be for some people. If we added in every feature every person wanted directly into the code, it would make MEAN very bloated and opinionated and in the end it wouldn't suit anyone's needs.

ilanbiala:
Closing for now because there's been no progress and this is better implemented on a case-by-case basis.

PR #414, opened 2015-02-17
Use validator.js instead of regexp for validations in User Schema.

fauria:
The original regexp did not filter some invalid email addresses, such as "foo..bar@example.com". Replaced with [validator.js](https://github.com/chriso/validator.js)


lirantal:
@fauria I like this.
Can you submit another PR for 0.4.0 branch instead of this one?

fauria:
Sure, no problem!

fauria:
Sure! Just submitted, see: https://github.com/meanjs/mean/pull/656

lirantal:
Great, I merged it, thank you.

PR #418, opened 2015-02-18
Wait for async removing

vepasto:


ilanbiala:
Have you actually had an error where the documents were not removed? I would rather do the remove in parallel and not serial if this is truly necessary.

ilanbiala:
Can you combine all these PRs into 1 PR?

vepasto:
Combined in #421

Issue #118, opened 2014-07-19
Multiple users can register with the same email address with default app


crewshin:
While registering users with a default app, it's possible to register multiple users with the same email address. It does however check if the username exists and prompts the user to change.

ilanbiala:
@crewshin identities and providers are being discussed in #337. @crewshin you may want to subscribe to that issue and chip in with your 2c.

SOSANA:
Any updates?

ilanbiala:
@SOSANA the plan is to handle that through #343, which needs to be merged into 0.4.0 rather than master. @sielay needs to create a new PR for 0.4.0 to avoid having lots of merge issues later on.

codydaig:
Since this moved to #343, and #343 was closed, can we close this issue?

lirantal:
@codydaig #343 is closed until we get that PR against 0.4.0 so maybe we leave this thread open still.

codydaig:
I believe this is taken care of now in #439

lirantal:
right, I'll close this @crewshin unless you spot any other problems with the fix in the latest 0.4.0 branch then you're welcome to re-open.

Issue #337, opened 2014-12-29
Discuss: Thoughts about identity handling


sielay:
Hi

I do enjoy learning and doing on this MEAN stack. In my projects I heavily play with multiple identity providers (can say that your code works well with all passport strategies I found so far). I have listed few things you may consider to address:
- putting first non-email provider into `user.providerData` and all other to `user.additionalProviders` has no benefits and a lot of workarounds needed. Following snippet can be useful:
```
UserSchema.methods.getProviderData = function ( provider ) {
    if ( this.provider === provider ) {
        return this.providerData;
    }
    return this.additionalProvidersData[ provider ];
};
```
- many strategies and/or APIs aren't very consequent about user ids. Many of them (like Github or Trello) allow you to use user hash and unique user name. We often save hash as user field for authentication sake, but later API calls often prefer to get `username`. We may consider having multi-id per provider.
- previous point lead to one of biggest pitfalls - we assume user can have only one email connected to the app; that also limit hugely use cases

Why I think that should be in core:
Authentication is core. It's not modular. A lot of different parts of the stack depend on it. If I fork out with huge changes, my project (or anyone who would like to say contributor to the root repo) will lose chance to link close to origin.

I try to address some of the issues in my small lib (https://github.com/sielay/octopusidentity doc is not up to date, but `meanuser`plugin is in latest version (npm not updated yet)). It puts some of the issues abstract, but to achieve all I want I would have to interfere more in core.

sielay:
Ok I will do it together with #273 on top of already PR'ed #288. In fact I need such change now.

ilanbiala:
Closing in favor of handling everything on the PR.

PR #439, opened 2015-02-28
Make emails unique

igorauad:
Emails are made unique. When user attempts to sign in through a provider in which his email is one that is already registered, user is redirected to the signin page with an error passed as a query string parameter.

Is this a reasonable solution to https://github.com/meanjs/mean/issues/118?

Your feedback is welcome. Thanks in advance.

### DIFF
--- a/modules/core/server/controllers/errors.server.controller.js
+++ b/modules/core/server/controllers/errors.server.controller.js
@@ -8,10 +8,10 @@ var getUniqueErrorMessage = function(err) {
 
 	try {
 		var fieldName = err.err.substring(err.err.lastIndexOf('.$') + 2, err.err.lastIndexOf('_1'));
-		output = fieldName.charAt(0).toUpperCase() + fieldName.slice(1) + ' already exist';
+		output = fieldName.charAt(0).toUpperCase() + fieldName.slice(1) + ' already exists';
 
 	} catch(ex) {
-		output = 'Unique field already exist';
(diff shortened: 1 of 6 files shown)


ilanbiala:
@igorauad I think if we merge in #337 then this may have some issues working. It would be better to merge that in first IMO and then make the necessary changes. It would be great if you could provide your 2¢ on that.

igorauad:
Ok, @ilanbiala . Thank you!

lirantal:
this PR needs some conflicts fixing before we can merge but overall looks like we can go for it.
@igorauad ?

igorauad:
@lirantal I've just fixed the conflicts

lirantal:
thanks @igorauad 
@rhutchinson @codydaig @mleanos LGTY?

codydaig:
@igorauad I pulled your branch down to test and it works! But, can we make it so it says the email is already in use versus "Unique field already exists". The user would have no way to know if it's the username or email field that is already taken. (Or even which fields must be unique.)

igorauad:
@codydaig see if the commit that I've just added solves the problem for you

codydaig:
@igorauad Tested and works great! Thanks :)

LGTM

lirantal:
Thanks @igorauad , @codydaig.

igorauad:
Thanks @codydaig , @lirantal

PR #445, opened 2015-03-03
update gulp-sass to ensure node-0.12 compatibility

dotch:
node-sass < 2.0 which gulp-sass depends upon can't be installed on node-0.12 so doing an npm install gave weird errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
 		"gulp-mocha": "~1.1.1",
 		"gulp-karma": "~0.0.4",
 		"gulp-protractor": "~0.0.11",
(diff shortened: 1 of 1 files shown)


gsavoie:
This fix does work, but any reason why it's not applied when I use the generator-meanjs 0.4-dev branch ? Should there be another change in the generator project ?

ilanbiala:
@gsavoie until 0.4.0 has been finished and we update the generator, it will be out of date.

gsavoie:
Thanks @ilanbiala, not a big a problem anyway. I'm not yet familiar how the two projects work together!

Issue #450, opened 2015-03-06
0.4.0 'gulp test' hangs because mongoose connections are staying open


reblace:
The mongoose connection opened as part of the 'test' task in the gulp build is never closed. This causes gulp to hang at the completion of the tests.

ilanbiala:
Does the referenced commit and changes made in `gulpfile.js` and `config/lib/mongoose.js` fix the issue?.

reblace:
You commit only references the gulpfile. I think you want to pass a callback into the disconnect method on the mongoose connection so you can call the gulp done method when the disconnect is complete. Eg:

mongoose.disconnect(function(err) {
	done(err);
});

Then, you can have the config/lib/mongoose.js file do something like this:

module.exports.disconnect = function(cb) {
	db.disconnect(function(err) {
		cb(err);
	});
};

ilanbiala:
Fixed in the latest commit.

reblace:
You can't do this:
> mongoose.connect(done);
because gulp interprets any parameter to the callback (done) as an error. So, in this case, the build immedaitely fails on the connect when the connect function returns a refernece to the database.

So, you need to do this instead on the openMongoose task:

mongoose.connect(function(db) {
	done();
});

ilanbiala:
[The build](https://travis-ci.org/meanjs/mean/builds/53291197) passes.

reblace:
> meanjs@0.4.0 test /home/travis/build/meanjs/mean
> grunt test

It isn't running the gulp build.

ilanbiala:
Out of curiosity, what error does Gulp throw because of that?

reblace:
Gulp fails because the task is reporting an error:

[22:53:28] 'openMongoose' errored after 247 ms
[22:53:28] Error: [object Object]
    at formatError (/usr/local/lib/node_modules/gulp/bin/gulp.js:161:10)
    at Gulp.<anonymous> (/usr/local/lib/node_modules/gulp/bin/gulp.js:187:15)
    at Gulp.emit (events.js:129:20)

You can check out the branch and do a "gulp test" to reproduce.

ilanbiala:
Fixed in the latest commit.

reblace:
Fixed.

reblace:
Just discovered that this issue isn't entirely fixed. The remaining issue is that when the tests fail, the 'closeMongoose' task isn't called, so the build still hangs.

One solution is to suppress errors on both the mocha and karma build tasks... eg:

[...]

PR #453, opened 2015-03-06
#450 Fix for hanging gulp on mongoose connections being left open

reblace:
I updated the gulp build so that the mongoose connect/disconnect take place in the mocha task based on the lifecycle of the stream. This way, we can guarantee that the disconnect function is always called regardless of if the mocha tests fail or not. I'm also sending an error message to the task callback in the event that an error is generated by the tests. This way, gulp will finish in an error state if the tests fail and will finish in an ok state if they pass. I've tested this with passing and failing tests and the gulp build closes correctly now.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -119,7 +119,6 @@ gulp.task('mocha', function (done) {
 
 	// Connect mongoose
 	mongoose.connect(function() {
-
 		// Run the tests
 		gulp.src(testAssets.tests.server)
 			.pipe(plugins.mocha({
@@ -128,7 +127,8 @@ gulp.task('mocha', function (done) {
 			.on('error', function (err) {
 				// If an error occurs, save it
 				error = err;
-			}).on('end', function() {
+			})
+			.on('end', function() {
 				// When the tests are done, disconnect mongoose and pass the error state back to gulp
 				mongoose.disconnect(function() {
 					done(error);
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -111,29 +111,26 @@ gulp.task('less', function () {
 		.pipe(gulp.dest('./modules/'));
 });
 
-// Connect to MongoDB using the mongoose module
-gulp.task('openMongoose', function (done) {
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -111,29 +111,26 @@ gulp.task('less', function () {
...
 
+	var error;
 	mongoose.connect(function() {
-		done();
+		gulp.src(testAssets.tests.server)
+			.pipe(plugins.mocha({
+				reporter: 'spec'
+			}))
+			.on('error', function (err) {
+				error = new Error('Mocha tests failed');
```
ilanbiala: How come we aren't using `err` to create the `error` variable?

ilanbiala:
I'm fine with the change, it's actually what I've been doing in my projects before MEAN.js added Gulp support. What happens when you don't have the error listener?

gulpfile.js
reblace: Actually, it's better to use the err value. I'll update the PR.

reblace:
You want to report the error still so that a CI server will be able to see that the build failed if some of the tests failed. Karma already reports the error because in the karma task you return a handle to the stream. The mocha task needs to manually send the error back up to gulp since you are not returning the reference to the stream.

gulpfile.js
```diff
@@ -111,29 +111,31 @@ gulp.task('less', function () {
...
+		gulp.src(testAssets.tests.server)
+			.pipe(plugins.mocha({
+				reporter: 'spec'
+			}))
+			.on('error', function (err) {
+				// If an error occurs, save it
+				error = err;
+			}).on('end', function() {
+				// When the tests are done, disconnect mongoose and pass the error state back to gulp
+				mongoose.disconnect(function(){
```
ilanbiala: Fix spacing.

ilanbiala:
Can you make sure that the code is formatted according to our editorconfig and jshintrc? Spacing and tabs are biggest priorities.

ilanbiala:
LGTM.

PR #463, opened 2015-03-09
Bump glob to version 5.0

adamwalz:
In glob v4.5 the sync option was deprecated in favor of `glob.sync(patterns, opts)`

https://github.com/isaacs/node-glob/tree/v4.5.0#options


config/config.js
```diff
@@ -56,26 +51,20 @@ var getGlobbedPaths = function(globPatterns, excludes) {
...
-
-            process.env.NODE_ENV = 'development';
+    var environmentFiles = glob.sync('./config/env/' + process.env.NODE_ENV + '.js');
+    console.log();
+    if (!environmentFiles.length) {
+        if (process.env.NODE_ENV) {
+            console.error(chalk.red('No configuration file found for "' + process.env.NODE_ENV + '" environment using development instead'));
         } else {
-            console.log(chalk.bold('Application loaded using the "' + process.env.NODE_ENV + '" environment configuration'));
+            console.error(chalk.red('NODE_ENV is not defined! Using default development environment'));
```
ilanbiala: This isn't an error message, more like info or warn or just log.

config/config.js
```diff
@@ -27,25 +27,20 @@ var getGlobbedPaths = function(globPatterns, excludes) {
...
-                if (excludes) {
-                    files = files.map(function(file) {
-                        if (_.isArray(excludes)) {
-                            for (var i in excludes) {
-                                file = file.replace(excludes[i], '');
-                            }
-                        } else {
-                            file = file.replace(excludes, '');
+            var files = glob.sync(globPatterns);
+            if (excludes) {
```
ilanbiala: Can't we handle this with Glob's exclusion syntax or with `ignore`?

config/config.js
adamwalz: The github diff is a little messed up. It makes it look like I changed that line from console.log to console.error, when really all that changed is indentation. 

If you would like me to change it to either `warn` or `log` though I can.

ilanbiala: @lirantal what do you think? Personally not having a `NODE_ENV` doesn't seem like an error.

config/config.js
adamwalz: Isn’t the excludes array here being treated as a rename operation instead of an ignore pattern.

For instance. If the cwd contained
public/lib/a
public/lib/b

`glob.sync(‘**’, {ignore: ‘public’})` would return [‘public/lib/‘, ‘public/lib/a’, ‘public/lib/b’]

Whereas `getGlobbedPaths(‘**’, [‘public/‘])` would return [‘lib/‘, ‘lib/a’, ‘lib/b’]

In this case the glob `ignore` option couldn’t be used.

ilanbiala:
LGTM.

PR #464, opened 2015-03-10
Fixed build to work in production environment

FREEZX:


ilanbiala:
@FREEZX I'd rather refactor to get production environment to build properly.

FREEZX:
@ilanbiala What would be the difference, the point is to build the development files for production anyways.

ilanbiala:
The difference is that development should not be concatenated, minified, etc. while production should be. The files generated/used are indeed different.

adamwalz:
This is fixed in the 0.4.0 branch by using default assets for minify/uglify tasks. A similar change for the master branch could be changing `grunt.config.set('applicationCSSFiles', config.assets.css);` to 

    var defaultConfig = require(‘./config/env/all’);
    …
    grunt.config.set('applicationCSSFiles', defaultConfig.assets.css);

Although I prefer the changes made in 0.4.0

FREEZX:
@ilanbiala what's the point in building for development? In the development environment application.js and application.min.js are not being loaded.

neilhanekom:
Sorry maybe noob question.. But these changes you have made. Will they be
updated in the yo generator/Npm? or do we have to clone the git repo...?

On Tue, Mar 10, 2015 at 9:19 AM, Kristijan Trajkovski <
notifications@github.com> wrote:

> @ilanbiala <https://github.com/ilanbiala> what's the point in building
> for development? In the development environment application.js and
> application.min.js are not being loaded.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/meanjs/mean/pull/464#issuecomment-78006317>.
>



-- 

Neil Hanekom
082 098 1180 (C)
neilhanekom1@gmail.com

neilhanekom:
Please explain the loadConfig as I not getting any builds done with this change

wesleyfsmith:
@FREEZX we want non-minified files loaded under development so that when we use dev tools in the browser, we can read something that's intelligible.

Only the files from production.js and the minified files should be included when we use grunt build.

 There is strange behavior with the all.js file, which I have documented here #472

FREEZX:
@wesleyfsmith You don't need to build to have the non-minified files, that's what development.js is for. Including development files. When you run in dev mode, it's the development files that are included in your html. Build is only needed to minify development files and get it ready for production.

On the note of production.js, it only includes the minified development files, so what's the logic behind building a minified file onto itself? 
Check these, and you'll see what i mean:
https://github.com/meanjs/mean/blob/master/config/env/production.js#L36
https://github.com/meanjs/mean/blob/master/gruntfile.js#L78

So production.js only references the minified application, that is generated by doing `grunt build`.

ilanbiala:
@FREEZX this PR should also be directed towards 0.4.0. Please create a new PR and close this one.

FREEZX:
@ilanbiala Obviously fixed there:
https://github.com/meanjs/mean/blob/0.4.0/gruntfile.js#L220

ilanbiala:
Oh I didn't catch that. In that case, this PR doesn't matter as 0.4.0 will be merged into master.

PR #480, opened 2015-03-17
test it 

sergiregi88:
It is a test you must don't accept it


lirantal:
@sergiregi88 ?

SOSANA:
@lirantal shouldn't this be broken up into individual PR's?

ilanbiala:
This looks like a conglomeration of commits that have been made on master or 0.4.0 because I've seen most of these changes in commits and PRs before.

lirantal:
this overall looks like a mistaken, so unless @sergiregi88 can comment what's this about we can ignore for now.

PR #485, opened 2015-03-20
Remove Bower, rely on npm for all dependencies

dotch:
What do you think about removing bower and using npm for the frontend dependencies?

Don't merge this just yet, it has the following issues:
- Only the Gulpfile is updated so far.
- I removed angular-utils, because it seems to not be used anywhere and they don't seem to regularly put their releases on npm.
- This change might have implications for the generator.

This pull-request should mainly be basis for discussion.


config/config.js
```diff
@@ -81,6 +81,12 @@ var initGlobalConfigFolders = function(config, assets) {
     config.folders.client = getGlobbedPaths(path.join(process.cwd(), 'modules/*/client/'), process.cwd().replace(new RegExp(/\\/g),'/'));
 };
 
+var transformLibPaths = function(paths) {
```
ilanbiala: why not just `function transformLibPaths` instead of putting it in a variable? Having it in a variable makes using an IDE harder and less useful in addition to just being unnecessary, no?

package.json
```diff
@@ -14,10 +14,17 @@
 	},
 	"scripts": {
 		"start": "grunt",
-		"test": "grunt test",
-		"postinstall": "bower install --config.interactive=false"
+		"test": "grunt test"
 	},
 	"dependencies": {
+		"angular": "^1.3.15",
```
ilanbiala: These should be `~`.

simison:
My 2cns:

- Frontend packages aren't that often in NPM — and while these specific packages would be there, bower.json offers a boilerplate that's easy to expand.

- I'm serving meanjs `./public/` folder directly with Nginx without NodeJS — having frontend libs under `./public/lib/` is incredibly useful for this.

- Bower uses flat folder structure, in NPM requirements are installed under ./package/node_modules/)

Interesting read; http://blog.npmjs.org/post/101775448305/npm-and-front-end-packaging

dotch:
@simison I agree with the fact that many front-end packages are not distributed via bower yet, however the trend seems to lead them to npm eventually.
Serving `.public/lib/` would still be possible, because the modified gulp tasks actually copy the front-end npm packages there.
@ilanbiala thanks for the comments, I will update this.

so I guess it would be reasonable to wait a while with removing bower until npm becomes the de-facto standard for front-end package management, which I think it will eventually.

simison:
True, now that you mention it's a trend, I've also noticed how quite a few packages have indeed appeared there.

+1 for waiting for a few months.

lirantal:
doesn't look like we'll be making this change to npm-only mode, closing.

Issue #185, opened 2014-09-22
Has anybody implemented signin redirection to "previous page"?


igorauad:
In case anyone implemented a scheme for redirecting the user to his previous state after sign-in , would it be possible to share?

I was thinking I could some how add the previous state information to $state (as below) and inform this to the backend, which, in turn, would have to inform back after successful authentication. Perhaps this is more work than necessary. Thoughts? 

```javascript
app.run(function ($rootScope, $state) {
  $rootScope.$on('$stateChangeSuccess', function (event, toState, toParams, fromState) {
    $state.previous = fromState;
  });
});
```

Uter1007:
Are there any updates to this feature ? Is it possible to have it already in the current master branch ? Is it possible to use: https://github.com/jaredhanson/connect-ensure-login ? Sorry for my n00b questions but I am a total Javascript rookie ;)

lirantal:
closed, implemented in https://github.com/meanjs/mean/pull/487

PR #487, opened 2015-03-22
Enable redirection to previous page after login

igorauad:
In response to #185

Two different strategies are adopted, one for when the user authenticates locally and the other through providers. When authenticating locally, the signin function in the client controller redirects to the previous state (storing and using a state name) after successful login. When authenticating through a provider, the first call to provider stores the previous URL (not state, URL) in the session. Then, when provider actually calls the authentication callback, session `redirect_to` path is used for redirecting user.

Note: I originally did this to master, so there is one extra thing to correct. First, note on branch master sign-in and signup are not nested states. Given they are nested in 0.4.0, the controller instantiated in the nested state does not capture the previous state properly. Therefore, this commit is not redirecting when authenticating locally yet. Suggestions are welcome.

For example, it seems an ugly solution to instantiate `AuthenticationController` both at parent and nested states. What would be better?

### DIFF
--- a/modules/users/client/views/authentication/authentication.client.view.html
+++ b/modules/users/client/views/authentication/authentication.client.view.html
@@ -1,24 +1,12 @@
-<section class="row">
+<section class="row" data-ng-controller="AuthenticationController">
 	<h3 class="col-md-12 text-center">Sign in using your social accounts</h3>
 	<div class="col-md-12 text-center">
-		<a href="/api/auth/facebook" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/facebook.png">
-		</a>
-		<a href="/api/auth/twitter" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/twitter.png">
-		</a>
-		<a href="/api/auth/google" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/google.png">
-		</a>
-		<a href="/api/auth/linkedin" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/linkedin.png">
-		</a>
-		<a href="/api/auth/github" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/github.png">
-		</a>
-		<a href="/api/auth/paypal" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/paypal.png">
-		</a>
+		<img ng-click="callOauthProvider('/api/auth/facebook')" ng-src="/modules/users/img/buttons/facebook.png">
+		<img ng-click="callOauthProvider('/api/auth/twitter')" ng-src="/modules/users/img/buttons/twitter.png">
+		<img ng-click="callOauthProvider('/api/auth/google')" ng-src="/modules/users/img/buttons/google.png">
+		<img ng-click="callOauthProvider('/api/auth/linkedin')" ng-src="/modules/users/img/buttons/linkedin.png">
+		<img ng-click="callOauthProvider('/api/auth/github')" ng-src="/modules/users/img/buttons/github.png">
+		<img ng-click="callOauthProvider('/api/auth/paypal')"  ng-src="/modules/users/img/buttons/paypal.png">
 	</div>
 	<div ui-view></div>
 </section>
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -9,6 +9,12 @@ var path = require('path'),
 	passport = require('passport'),
 	User = mongoose.model('User');
 
+// URLs for which user can't be redirected on signin
+var noReturnUrls = [
+	'/authentication/signin',
+	'/authentication/signup'
+];
+
 /**
  * Signup
  */
@@ -78,10 +84,29 @@ exports.signout = function (req, res) {
 };
 
 /**
+ * OAuth provider call
+ */
+exports.oauthCall = function(strategy, scope) {
+	return function(req, res, next) {
+		// Set redirection path on session.
+		// Do not redirect to a signin or signup page
+		if (noReturnUrls.indexOf(req.query.redirect_to) === -1) {
+			req.session.redirect_to = req.query.redirect_to;
+		}
+		// Authenticate
+		passport.authenticate(strategy, scope)(req, res, next);
+	};
+};
+
+/**
  * OAuth callback
  */
 exports.oauthCallback = function (strategy) {
 	return function (req, res, next) {
+		// Pop redirect URL from session
+		var sessionRedirectURL = req.session.redirect_to;
+		delete req.session.redirect_to;
+
 		passport.authenticate(strategy, function (err, user, redirectURL) {
 			if (err) {
 				return res.redirect('/authentication/signin?err=' + encodeURIComponent(errorHandler.getErrorMessage(err)));
@@ -94,7 +119,7 @@ exports.oauthCallback = function (strategy) {
 					return res.redirect('/authentication/signin');
 				}
 
-				return res.redirect(redirectURL || '/');
+				return res.redirect(redirectURL || sessionRedirectURL || '/');
 			});
 		})(req, res, next);
 	};
--- a/modules/core/client/app/init.js
+++ b/modules/core/client/app/init.js
@@ -32,6 +32,14 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function($roo
             }
         }
     });
+    // Record previous state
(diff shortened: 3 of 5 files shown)


modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -79,10 +79,32 @@ exports.signout = function(req, res) {
...
+};
+
+/**
  * OAuth callback
  */
 exports.oauthCallback = function(strategy) {
 	return function(req, res, next) {
+		// Pop redirect URL from session
+		var sessionRedirectURL = req.session.redirect_to;
+		delete req.session.redirect_to;
```
ilanbiala: Why are we temporarily setting a property on the req session?

igorauad: @ilanbiala thanks for looking into this. I am setting a property in the first call, `oauthCall`. In the callback I retrieve it. Do you think there would be a better way to implement this?

ilanbiala: I think I'm missing something, but what about the `successRedirect` property?
http://passportjs.org/guide/google/

igorauad: @ilanbiala , the callback is called by the provider. As I understand, it is not possible to pass the previous URL to the callback. This is why a `oauthCall` was introduced, so we have a way of storing the previous state. If we use `successRedirect`, then we would again need to pass the value for `successRedirect` somehow, perhaps through a property on session.

Here is where the idea came from [http://stackoverflow.com/a/13336055/2859410](http://stackoverflow.com/a/13336055/2859410)

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -79,10 +79,32 @@ exports.signout = function(req, res) {
 };
 
 /**
+ * OAuth provider call
+ */
+exports.oauthCall = function(strategy, scope) {
+	return function(req, res, next) {
+		// Set redirection path on session.
+		// Do not redirect to a signin or signup page
+		if (!_.contains([
```
ilanbiala: Why not just do `[].indexOf`? Is lodash faster or any different?

modules/users/server/controllers/users/users.authentication.server.controller.js
ilanbiala: What is `req.url` in `oauthCall`? What about in `oauthCallback`?

igorauad: For example:

`oauthCall`: `/api/auth/twitter`
`oauthCallback`: `/api/auth/twitter/callback?oauth_token=xxx&oauth_verifier=xxx`

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -79,10 +79,29 @@ exports.signout = function(req, res) {
 };
 
 /**
+ * OAuth provider call
+ */
+exports.oauthCall = function(strategy, scope) {
+	return function(req, res, next) {
+		// Set redirection path on session.
+		// Do not redirect to a signin or signup page
+		if (['/authentication/signin', '/authentication/signup'].indexOf(req.query.redirect_to) === -1) {
```
ilanbiala: Instantiating this array every time will be costly, let's store a constant reference somewhere, no?

modules/users/server/controllers/users/users.authentication.server.controller.js
ilanbiala: So this PR uses a setter on the URL in Angular to tell the server which URL to redirect to? Is there no way for us to know what page the user was previously on before the going to the authentication route?

igorauad: Not that I could think of. As my understanding, this Information does not come in `req` in the `oauthCallback`

modules/users/server/controllers/users/users.authentication.server.controller.js
igorauad: Sure!

lirantal:
@igorauad what's the status here? we can't merge it to 0.4.0 due to conflicts.

igorauad:
Hi @lirantal , thanks for asking. There was something missing, which I just added. Have a look, if possible.

lirantal:
thanks @igorauad, I assume you tested it? as we don't have any frontend tests yet to confirm.
@rhutchinson @ilanbiala @simison can you review?

modules/users/client/views/authentication/authentication.client.view.html
```diff
@@ -1,24 +1,12 @@
...
-		<a href="/api/auth/linkedin" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/linkedin.png">
-		</a>
-		<a href="/api/auth/github" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/github.png">
-		</a>
-		<a href="/api/auth/paypal" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/paypal.png">
-		</a>
+		<img ng-click="callOauthProvider('/api/auth/facebook')" src="/modules/users/img/buttons/facebook.png">
```
ilanbiala: Shouldn't these use ng-src even though we know the resource?

igorauad: @ilanbiala let me know once you think the modifications are ready, so that I can squash the commits

rhutchison:
@lirantal Need more time to review.  I think this should wait for post-0.4.0

simison:
I blew up my test setup for oauth stuff, gotta wait a bit before I can test this stuff again.

PS. https://pagekite.net/ is pretty neat for testing this.

lirantal:
sure, I'm ok with it. @ilanbiala what do you think?

ilanbiala:
Looks good, but I haven't tested it.

igorauad:
I've just realized something specifically for Paypal authentication was missing, since it didn't exist when I first wrote this PR. I have included it and re-tested the modifications to double check. It is working for both local and external authentication. I think it is ready for a merge. Anybody can also confirm?

ilanbiala:
@igorauad are you familiar with adding tests for strategies? @lirantal I think we should push for that in the next few versions, because OAuth strategies can change quite a bit.

igorauad:
@ilanbiala tests would be very useful, but I haven't seen an approach for this. Do you want to share anything that I could take a look?

ilanbiala:
@igorauad I did it with a local strategy https://github.com/HTHS/hths-lunch/blob/master/test/server/mock-strategy.js, but maybe we can adapt it to use e2e tests to do OAuth?

lirantal:
@ilanbiala yeah I agree on the tests. I'm also working on adding some coverage to the code base,  and I'd love it if you can help out let's take it in another thread.

About this PR - need angular people to review. @rhutchison WDYT?

modules/users/client/views/authentication/authentication.client.view.html
```diff
@@ -1,24 +1,12 @@
...
-		<a href="/api/auth/linkedin" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/linkedin.png">
-		</a>
-		<a href="/api/auth/github" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/github.png">
-		</a>
-		<a href="/api/auth/paypal" target="_self" class="undecorated-link">
-			<img src="/modules/users/img/buttons/paypal.png">
-		</a>
+		<img ng-click="callOauthProvider('/api/auth/facebook')" ng-src="/modules/users/img/buttons/facebook.png">
```
pgrodrigues: There are some `data-` prefix missing both in `ng-click` and `ng-src`.

ilanbiala: data- isn't necessary, I'm not sure why we even started using it.

pgrodrigues:
@ilanbiala probably it was most likely to pass in w3c html validation tool

modules/users/client/views/authentication/authentication.client.view.html
simison: That was when meanjs was in xhtml era: https://github.com/meanjs/mean/blob/master/app/views/layout.server.view.html#L2

Now with html5 it's perfectly fine not to have them: https://github.com/meanjs/mean/blob/0.4.0/modules/core/server/views/layout.server.view.html#L1-L2

ilanbiala:
@igorauad is this ready to review?

igorauad:
Yes, @ilanbiala. This PR works. Is there anything I should change?

ilanbiala:
@lirantal LGTM.

lirantal:
Sure.
@igorauad can you please rebase to squash your commits?

igorauad:
Done, @lirantal

lirantal:
Thanks @igorauad

Issue #449, opened 2015-03-05
fig -> docker compose


tejasmanohar:
https://github.com/meanjs/mean/tree/0.4.0#development-and-deployment-with-docker

fig is deprecated, docker compose is the new thing.

ilanbiala:
Can you put together a PR to update it?

kbowerma:
Yes but you can just change 'fig' to 'docker-compose' and it will even look for the fig.yml file.

However in the yml file you set the NODE_ENV to development which is looking at development.js which has localhost hardcoded to the uri. ( ```		uri: 'mongodb://localhost/mean-dev',```  I think you either need to make a new docker.js config  file that use ```uri: 'mongodb://db_1/mean-dev',``` or ```mean_db_1``` then set the ```NODE_ENV:  docker``` in the yml file.  

Or you could just edit the development.js to be something like:

```
uri:   process.env.DOCKER_MONGO_URI   ||  'mongodb://localhost/mean-dev',   

```
and set that var from the yaml file.

SOSANA:
Once we fix this, can we update docs for others as well to get started with docker

codydaig:
Has this been fixed with pull request #611 ?

lirantal:
let's close, I think we're good here, otherwise @tejasmanohar can re-open.

PR #492, opened 2015-03-24
removed reference of fig and replaced #449 with docker-compose and chang...

kbowerma:
...ed the cofnig file to look at a env var which is set from the yaml file

I went down the path of adding logic to the development.js config file but could add a docker.js config file instead if you prefer

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -2,7 +2,7 @@
 
 module.exports = {
 	db: {
-		uri: 'mongodb://localhost/mean-dev',
+		uri: process.env.DOCKER_MONG_URL || 'mongodb://localhost/mean-dev',
 		options: {
 			user: '',
 			pass: ''
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,9 +4,10 @@ web:
    - db
   ports:
    - "3000:3000"
-  environment: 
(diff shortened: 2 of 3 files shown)


config/env/development.js
```diff
@@ -2,7 +2,7 @@
 
 module.exports = {
 	db: {
-		uri: 'mongodb://localhost/mean-dev',
+		uri: process.env.DOCKER_MONG_URL || 'mongodb://localhost/mean-dev',
```
ilanbiala: Missing an O.

docker-compose.yml
```diff
@@ -4,9 +4,10 @@ web:
    - db
   ports:
    - "3000:3000"
-  environment: 
+  environment:
    NODE_ENV: development
+   DOCKER_MONG_URL: mongodb://mean_db_1/mean-dev
```
ilanbiala: Same here, missing an O.

ilanbiala:
@kbowerma this PR should be against 0.4.0 so that we can avoid merge conflicts when releasing 0.4.0.

codydaig:
@kbowerma @lirantal I think this has been resolved in the latest branch.

lirantal:
Closing.

Issue #493, opened 2015-03-25
Authentication service is reset to an empty string when redirected


c1moore:
I noticed in my application, which has different levels of access for users (admin, user, ect.), that when a user is redirected from a page to which they should not have access (e.g.: an admin page when they are a normal user) that Authentication.user is reset to an empty string.

In my application, only registered users should see a page other than the sign in/sign up page.  So when a logged in user is redirected from the restricted page to the homepage, the homepage immediately redirects them to the sign in page and they are required to sign back in, which appears to be the expected behavior.  However, if the user refreshes the page, Authentication is reset to the value in window.user and is redirected to the homepage.  As log as the user does not attempt to access any data restricted to the user that was stored in window.user, the current user can access all information.  I believe window.user should also be set to null when setting Authentication.user to null to ensure that this cannot happen.

ilanbiala:
@c1moore is this still an issue for you?

c1moore:
I have fixed this in my own project, but, as I stated in the [pull request](https://github.com/meanjs/mean/pull/494), the code where the bug exists in the most recent version of mean is still the same.  I have not been able to run tests with the most recent version yet.  If you wish to close this issue before testing, go ahead, but it may be an overlooked security hole for other users.

ilanbiala:
@c1moore there was no movement on that PR. @lirantal @codydaig do you still see this as an issue?

jloveland:
@c1moore I would like to help reproduce what you are seeing. I can't seem to do this. If I try to access the following url without admin role http://localhost:3000/admin/users, I get a forbidden message in the ui
![screen shot 2015-11-03 at 10 18 37 pm](https://cloud.githubusercontent.com/assets/3456708/10928689/f428c982-8278-11e5-8387-659411d9be86.png)

I get redirectd to http://localhost:3000/forbidden. I do not get logged out.

Can you help me try to reproduce by pulling from master, reproducing, and posting very detailed instructions on the steps you took?

mleanos:
I'm experiencing the same behavior that @jloveland reported. 

Doesn't look like this is an issue anymore.

ilanbiala:
@c1moore can you just double check against current master?

lirantal:
Closing old issue.

PR #494, opened 2015-03-25
Fix authentication problem

c1moore:
Set window.user to null so one can not simply refresh the page and magically log back in without providing credentials as mentioned in issue #493.

### DIFF
--- a/public/modules/users/config/users.client.config.js
+++ b/public/modules/users/config/users.client.config.js
@@ -4,14 +4,15 @@
 angular.module('users').config(['$httpProvider',
 	function($httpProvider) {
 		// Set the httpProvider "not authorized" interceptor
-		$httpProvider.interceptors.push(['$q', '$location', 'Authentication',
-			function($q, $location, Authentication) {
+		$httpProvider.interceptors.push(['$q', '$location', 'Authentication', '$window',
+			function($q, $location, Authentication, $window) {
 				return {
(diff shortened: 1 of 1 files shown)


ilanbiala:
Doesn't it make more sense to `angular.copy` the object so we don't rely on it existing or not existing anymore?

c1moore:
I don't think that would fix the problem, if I understand `angular.copy` fully.  The problem exists because the `window.user` remains set after we unset `Authentication.user`.  If we use `angular.copy` the same problem would still exist.  For example, say a user logs out then somebody refreshes the app.  The app will use `angular.copy` to set `Authentication.user` using `window.user`.  Since `window.user` was not reset when the previous user logged out, the person that refreshed the page would now be logged in as the previous user.

We could use `$watch` on both `Authentication.user` and `window.user` and update one if the other changes, but this may have a noticeable affect on performance.

codydaig:
@c1moore Is this still an issue in the latest master branch?

lirantal:
Closing for now.

c1moore:
I would have to pull the most recent version and find a way to test it since it has changed so much since the version I use, but the code dealing with this problem appears to be the same so I would assume the problem still exists.  I will confirm when I have a chance.

lirantal:
@c1moore  We'd appreciaite if you can please quickly test that on a fresh clone of the `master` branch

PR #503, opened 2015-04-02
return a 404 for not found api, module and lib routes

### LINKED ISSUE #501: Page not found handler not working in 0.4.0
Steps to reproduce:
* clone the repository;
* run `npm install`;
* run `grunt`;
* load an URL for which there's no route, e.g. http://localhost:3000/abcdefghijklmnopqrstuvwxyz (why not ;) )

Expected result:
* a 404 status code and a redirection to http://localhost:3000/not-found , displaying an error message.

Actual result:
* a 200 status code and a redirection to http://localhost:3000 , effectively displaying the root page.

If I remove the "*" from the last route of modules/core/server/routes/core.server.routes.js, the 404 functionality is restored; however, sign up, sign in and pretty much everything else is broken, because of the fallback to the error routes defined in config/lib/express.js. I haven't yet found what the correct fix is.

I noticed the issue on my own testing application, when the server not only produced a 200 status code, but also the HTML content of the root page, when requested an image which wasn't there.

dotch:
fixes #501

### DIFF
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -177,7 +177,6 @@ module.exports.initModulesServerRoutes = function (app) {
  * Configure error handling
  */
 module.exports.initErrorRoutes = function (app) {
-	// Assume 'not found' in the error msgs is a 404. this is somewhat silly, but valid, you can do whatever you like, set properties, use instanceof etc.
 	app.use(function (err, req, res, next) {
 		// If the error object doesn't exists
 		if (!err) return next();
@@ -187,12 +186,6 @@ module.exports.initErrorRoutes = function (app) {
 
 		// Redirect to error page
 		res.redirect('/server-error');
-	});
-
-	// Assume 404 since no middleware responded
-	app.use(function (req, res) {
-		// Redirect to not found page
-		res.redirect('/not-found');
 	});
 };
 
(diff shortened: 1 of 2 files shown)


ilanbiala:
So what routes have been modified now?

dotch:
Sorry for the lacking explanation.

The server is responsible for the following routes:
`/api/*`
`/componens/*`
`/lib/*`
`/server-error`
The other routes (`/*`) are handled by Angular due to html5mode.

This PR shows the server 404 page for:
`/api/*`
`/componens/*`
`/lib/*`
when no route-handler kicks in due a non existing endpoint, component or library.

Angular route handling is not part of this PR.

simison:
BTW, would be great if non existing API routes would return 404 as a JSON.

Something like this http://stackoverflow.com/a/9802006

lirantal:
sounds like a good enhancement indeed, need to take a look at that further

simison:
@lirantal are you gonna look into 404 stuff for 0.4.0? (har har those digits). If not, I'm also fiddling with this stuff now, but I'm sure you have more perspective.

This still requires 404 handling at Angular side, now app just redirects to `/` (often confusing users). Redirecting to `/not-found` is kinda fine, but would be better not to change URL at the URL-bar. `location:false` option at [$state.go()](http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.$state#methods_go) could be useful for that.

Would be lovely to get real 404 headers working for `/*` stuff as well from Express (instead of giving the job for Angular), but that would require shared configs for URL routing between Express and Angular. Not sure how feasible is that...

lirantal:
@simison go ahead and look at it, it's ok with me.

simison:
@lirantal See https://github.com/meanjs/mean/pull/566

PR #509, opened 2015-04-08
Fix headers to HTML5 format

simison:
Headers were a weird mixture of html5 and xhtml. Now it's pure html5.

Use this as doctype if you really want it to be xhtml, but I don't see why would you want that with Angular:
```html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```


ilanbiala:
LGTM.

PR #516, opened 2015-04-13
Removed duplicate include for crypto in modules/users/server/controllers/users/users.password.server.controller.js

tcwd:


ilanbiala:
LGTM.

PR #520, opened 2015-04-19
Change the way Express gets hostname

simison:
Getting host via req.get('host') instead of req.hostname. 

get() methods adds also port if it's something else than 80.

Otherwise links that should be "localhost:3000" end up being just "localhost".

### DIFF
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -35,7 +35,7 @@ module.exports.initLocalVariables = function (app) {
 
 	// Passing the request url to environment locals
 	app.use(function (req, res, next) {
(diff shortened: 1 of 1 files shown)


simison:
On second thought, nope. 

This would've been great for stuff like:
`<link href="http://localhost:3000/modules/core/img/favicon.png" rel="shortcut icon" type="image/png" />`

But I overlooked this:
`<script type="text/javascript" src="http://localhost:3000:35729/livereload.js"></script>`

I think on my own version I'll just pass separate `hostPort` and `host` parameters, but dunno what's the best for the boilerplate.

Sorry noise. :-)

PR #521, opened 2015-04-19
Change Mongoose unique to boolean

simison:
As it should be: http://mongoosejs.com/docs/api.html#schematype_SchemaType-unique

### DIFF
--- a/app/models/user.server.model.js
+++ b/app/models/user.server.model.js
@@ -50,7 +50,7 @@ var UserSchema = new Schema({
 	},
 	username: {
 		type: String,
-		unique: 'Username already exists',
+		unique: true,
 		required: 'Please fill in a username',
 		trim: true
 	},


ilanbiala:
Was that changed in Mongoose 4.0? Is there a way to make sure these are the 3.8 docs?

simison:
Nope, it was boolean also with 3.x: http://mongoosejs.com/docs/3.7.x/docs/api.html#schematype_SchemaType-unique
Not sure when it has changed, if ever; http://mongoosejs.com/docs/prior.html

The irritating bit is that currently at signup controller, Mean.js doesn't shout out if it's the email that's unique or the username. It just shouts "[Unique field already exist](https://github.com/meanjs/mean/blob/0.4.0/modules/core/server/controllers/errors.server.controller.js#L14)", which isn't very helpful for the user.

[Mongoose-unique-validator](]https://www.npmjs.com/package/mongoose-unique-validator) tries to solve the issue, but I don't know if that's a great package.

ilanbiala:
@lirantal thoughts on why it was a string?

lirantal:
I don't recall on why this was set as string, but it seems a reasonable fix to update to boolean so I'll merge if it LTY

ilanbiala:
Lgtm.

simison:
Volá, now someone made a plugin that adds support to `unique: 'Error message'` -format. I might add this first to my project and then update this PL.

Read more: https://github.com/Automattic/mongoose/issues/2284#issuecomment-102785674

codydaig:
@simison Is this something that still needs to get changed in the 0.4 version? Or are we leaving it as a string to represent the error message?

lirantal:
@simison is that a bug though? it does work today because there's a validation attached to that specific field. Correct me if I'm wrong.

simison:
I think it "works" now with a string because it returns true, but it doesn't work as expected since that message won't be sent with the error. I recently implemented a solution elsewhere with that [mongoose-beautiful-unique-validation](https://www.npmjs.com/package/mongoose-beautiful-unique-validation) package so I think I could send a PL a bit later.

Let's leave the string there for now and it'll then just work with mongoose-beautiful-unique-validation. Closing this.

ilanbiala:
@simison I think we may also run into some issues when we upgrade to Mongoose Mongoose 4.1.x, which would definitely be awesome when we start using promises instead of callbacks.

PR #534, opened 2015-04-26
dockerfile/nodejs has been merged to official repo node

jvasallo:
I tried to get started with the docker file and kept getting Error: image dockerfile/nodejs:latest not found

After some digging, looks like it has been merged to: https://registry.hub.docker.com/_/node/

I can't even go to the old https://registry.hub.docker.com/u/dockerfiles/nodejs.

Sorry if this is wrong, first commit using docker in sometime.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM dockerfile/nodejs
+FROM node
 
 MAINTAINER Matthias Luebken, matthias@catalyst-zero.com
 


codydaig:
Looks like this has already been changed in 0.4.0
https://github.com/meanjs/mean/blob/0.4.0/Dockerfile#L1

@lirantal Should be changed in master as well?

lirantal:
@codydaig is right, we already have this change on 0.4.0 so we can close this one as it's not something mandatory for master.

PR #537, opened 2015-04-29
add backend support for Article voting

ppotoplyak:
Hi,

This patch introduces backend support for up/down voting an Article.

I don't expect it to be merged as is but my intent is to start a conversation on the design decisions and gauge interest for up/down voting.

thanks

### DIFF
--- a/app/controllers/users/users.profile.server.controller.js
+++ b/app/controllers/users/users.profile.server.controller.js
@@ -54,3 +54,66 @@ exports.update = function(req, res) {
 exports.me = function(req, res) {
 	res.json(req.user || null);
 };
+
+/**
+ * Record User's vote and update Article vote counts
+ */
+exports.vote = function(req, res) {
+	var article = req.article;
+	var user = req.user;
+	var newVoteDirection = req.params.direction;
+	var oldVoteDirection;
+	var upVotesAddend = 0;
+	var downVotesAddend = 0;
+	var voteIndex;
+
+	// look for existing vote
+	for (voteIndex = 0; voteIndex < user.votes.length; voteIndex++) {
+		if (article._id.toString() === user.votes[voteIndex]._id.toString()) {
+			break;
+		}
+	}
+
+	// if no vote exists, add one
+	if (voteIndex === user.votes.length) {
+		user.votes.push({
+			_id: article._id,
+			direction:'none'
+		});
+	}
+
+	oldVoteDirection = user.votes[voteIndex].direction;
+	user.votes[voteIndex].direction = newVoteDirection;
+	if (oldVoteDirection === 'none') {
+		upVotesAddend = newVoteDirection === 'up' ? 1 : 0;
+		downVotesAddend = newVoteDirection === 'down' ? 1 : 0;
+	} else if (newVoteDirection === 'none') {
+		upVotesAddend = oldVoteDirection === 'up' ? -1 : 0;
+		downVotesAddend = oldVoteDirection === 'down' ? -1 : 0;
+	} else if (oldVoteDirection !== newVoteDirection) {
+		upVotesAddend = newVoteDirection === 'up' ? 1 : -1;
+		downVotesAddend = upVotesAddend * -1;
+	}
+
+	user.save(function(err) {
+		if (err) {
+			return res.status(400).send({
+				message: errorHandler.getErrorMessage(err)
+			});
+		} else {
+			article.update({ $inc: { upVotes: upVotesAddend, downVotes: downVotesAddend }}, {}, function(err) {
+				if (err) {
+					return res.status(500).send({
+						message: errorHandler.getErrorMessage(err)
+					});
+				} else {
+					res.json({
+						oldVoteDirection: oldVoteDirection,
+						newVoteDirection: newVoteDirection
+					});
+				}
+			});
+		}
+	});
+
+};
--- a/app/models/article.server.model.js
+++ b/app/models/article.server.model.js
@@ -28,6 +28,14 @@ var ArticleSchema = new Schema({
 	user: {
 		type: Schema.ObjectId,
 		ref: 'User'
+	},
+	upVotes: {
+		type: Number,
+		default: 0
+	},
+	downVotes: {
+		type: Number,
+		default: 0
 	}
 });
 
--- a/app/models/user.server.model.js
+++ b/app/models/user.server.model.js
@@ -82,6 +82,17 @@ var UserSchema = new Schema({
 		type: Date,
 		default: Date.now
 	},
+	votes: [{
+		_id: {
+			type: Schema.ObjectId,
+			ref: 'Article',
+			index: true
+		},
+		direction: {
+			type: String,
+			enum: ['up', 'down', 'none']
+		}
+	}],
 	/* For reset password */
 	resetPasswordToken: {
 		type: String
--- a/app/routes/users.server.routes.js
+++ b/app/routes/users.server.routes.js
@@ -52,6 +52,9 @@ module.exports = function(app) {
 	app.route('/auth/github').get(passport.authenticate('github'));
 	app.route('/auth/github/callback').get(users.oauthCallback('github'));
 
+	// up/down voting on Articles
+	app.route('/users/vote/:direction/:articleId').put(users.requiresLogin, users.vote);
+
 	// Finish by binding the user middleware
 	app.param('userId', users.userByID);
 };
--- a/app/tests/user.votes.routes.test.js
+++ b/app/tests/user.votes.routes.test.js
@@ -1,0 +1,130 @@
+'use strict';
+
+var should = require('should'),
+	request = require('supertest'),
+	app = require('../../server'),
+	mongoose = require('mongoose'),
+	User = mongoose.model('User'),
+	Article = mongoose.model('Article'),
+	agent = request.agent(app);
+
+/**
+ * Globals
+ */
+var credentials, user, article;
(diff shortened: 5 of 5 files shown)


SOSANA:
Why not? Any additional features is 1+

simison:
_Any additional features is 1+_

Sounds like Linux settings panel. :-)

I'd add; _if it serves a purpose for a general purpose boilerplate_.

app/controllers/users/users.profile.server.controller.js
```diff
@@ -54,3 +54,49 @@ exports.update = function(req, res) {
...
+	// prevent User cross voting
+	if(req.user.id !== req.profile.id) {
+		var err = new Error('User ' + req.user.id + ' trying to vote as user ' + req.profile.id);
+		return res.status(400).send({
+			message: errorHandler.getErrorMessage(err)
+		});
+	}
+
+	// look for existing vote
+	for(voteIndex = 0; voteIndex<req.profile.votes.length; voteIndex++) {
```
ilanbiala: spacing

app/controllers/users/users.profile.server.controller.js
```diff
@@ -54,3 +54,49 @@ exports.update = function(req, res) {
...
+		if(req.article._id.equals(req.profile.votes[voteIndex]._id)) {
```
ilanbiala: spacing

app/controllers/users/users.profile.server.controller.js
```diff
@@ -54,3 +54,49 @@ exports.update = function(req, res) {
...
+			_id: req.article._id,
+			direction:'none'
+		});
+	}
+
+	oldVoteDirection = req.profile.votes[voteIndex].direction;
+	req.profile.votes[voteIndex].direction = req.params.direction;
+	req.profile.save(function(err) {
+		if (err) {
+			return res.status(400).send({
```
ilanbiala: In which case is this a client error and not a 5xx of sorts?

ilanbiala:
Added some comments. As a whole it seems like you are keeping track of direction. Why not just keep track of the number itself, i.e. 5, -6, 8. So the model would have:
```
votes: {
  type: Number,
  default: 0
}
```

Also, why not just check if `req.user` is defined and avoid having to keep track of `userId`. Then you would just pass the article and up/down, and then do `article.update({ $inc: { votes: direction === 'up' ? 1 : -1 }}, {}, callback);`?

ppotoplyak:
@ilanbiala , note that this patch does not accumulate the total votes for an Article, the accumulation would be a follow on patch.  You are correct, the userId in the route is superfluous and req.user should be used.  The vote direction could indeed be a [-1|0|1] instead of a string.
 
In my mind the very first step is to record a Users vote somewhere, i.e. the User Id, Article Id, Direction triple.  This patch accomplishes this by adding a nested Document to the User model.  I am not convinced that this is the best solution and I think this is the first critical design decision.  What do you guys think?

ilanbiala:
Why make a followup PR? This is all 1 feature (voting), so it should be encompassed in one PR. I would skip directly to counting votes on the article, because that's what more people probably want. Obviously the user needs to know which articles he has voted on, I think the array of voted article IDs and direction is probably the best in terms of MongoDB implementations.

ppotoplyak:
Sure, I can roll Article vote totals and vote recording into a single PR.  However, I would like to agree on the model changes.  

It sounds like you are okay with the current vote recording design? That is an array of Article IDs and direction added to the User model.

For the Articles model, the following would need to be added to support totals:
```
upVotes: {
  type: Number,
  default: 0
},
downVotes: {
  type: Number,
  default: 0
}
```

Let me know if this sounds reasonable.

thanks

ilanbiala:
Why do we need to keep track of the votes separately in the article model itself? Why not just have a votes property and increment or decrement it from 0?

ppotoplyak:
I am assuming you are asking why have upVotes and downVotes as opposed to just votes on the Article model?

In my mind the single vote property would not reflect the number of votes which took place, which is an important dimension.  I am very used to seeing Thumbs Up and Thumbs Down Icons with counts next to them. e.g. youtube

I think a single vote count would work but it's just not as flexible as two distinct fields.  Albeit, the state transitions will be more complicated.

ppotoplyak:
PR updated to record up/down vote count totals on Articles as well as record votes.

Hopefully the spacing issues have been taken care of.

@ilanbiala , I have not changed the error codes and need a little more info as to what they should be.

thanks

app/controllers/users/users.profile.server.controller.js
```diff
@@ -54,3 +54,66 @@ exports.update = function(req, res) {
...
+	var article = req.article;
+	var user = req.user;
+	var newVoteDirection = req.params.direction;
+	var oldVoteDirection;
+	var upVotesAddend = 0;
+	var downVotesAddend = 0;
+	var voteIndex;
+
+	// look for existing vote
+	for (voteIndex = 0; voteIndex<user.votes.length; voteIndex++) {
```
ilanbiala: Spacing on second condition

app/controllers/users/users.profile.server.controller.js
```diff
@@ -54,3 +54,66 @@ exports.update = function(req, res) {
...
+		if (article._id.equals(user.votes[voteIndex]._id)) {
```
ilanbiala: Why not just `==`?

ppotoplyak: I originally had == or === but was not getting the results which I expected as the type of what's being compared is ObjectId.  I believe that I got the equals solution from here:

http://stackoverflow.com/questions/11060213/mongoose-objectid-comparisons-fail-inconsistently

A straight == (or ===) comparison will compare the two objects by reference, not value. So that will only evaluate to true if they both reference the very same instance.

Instead, you should be using the equals method of ObjectID to compare their values

ilanbiala: You can compare the String values using `.toString()` though, which I think is simpler and more JS style.

ppotoplyak:
PR updated with:
1) hopefully last spacing fix
2) id comparison changed to use ```.toString()```

app/controllers/users/users.profile.server.controller.js
```diff
@@ -54,3 +54,66 @@ exports.update = function(req, res) {
...
+
+	user.save(function(err) {
+		if (err) {
+			return res.status(400).send({
+				message: errorHandler.getErrorMessage(err)
+			});
+		} else {
+			article.update({ $inc: { upVotes: upVotesAddend, downVotes: downVotesAddend }}, {}, function(err) {
+				if (err) {
+					return res.status(400).send({
```
ilanbiala: This should be at least a 500, but preferably something more specific if you can find a 5xx code that is more applicable.

ilanbiala:
Other than the HTTP code, it looks good. @lirantal what do you think?

lirantal:
@ilanbiala I'm all up for it.
@ppotoplyak any chance on adding some companion tests for the voting?

ppotoplyak:
@lirantal absolutely, latest PR adds ```app/tests/user.votes.routes.test.js```

There are 11 tests:
1) should not be able to vote if not logged in
2) should not be able to vote if direction is invalid
3-11) test all 9 possible two vote sequences and verify return values & Article vote counts

I tried using an existing file but the beforeEach clauses did not seem compatible.  This is my first time using SuperTest, Should.js, and Mocha so please forgive any rookie mistakes.

@ilanbiala return code changed to 500, I could not find anything more fitting.  Let me know if I introduced any more spacing issues.  Are you doing a visual check or using a style checker?

thanks

lirantal:
Looks ok, thanks.
@ilanbiala I'm ok with merging.

app/controllers/users/users.profile.server.controller.js
```diff
@@ -54,3 +54,66 @@ exports.update = function(req, res) {
...
+		upVotesAddend = oldVoteDirection === 'up' ? -1 : 0;
+		downVotesAddend = oldVoteDirection === 'down' ? -1 : 0;
+	} else if (oldVoteDirection !== newVoteDirection) {
+		upVotesAddend = newVoteDirection === 'up' ? 1 : -1;
+		downVotesAddend = upVotesAddend * -1;
+	}
+
+	user.save(function(err) {
+		if (err) {
+			return res.status(400).send({
```
ilanbiala: Can we change this to `500` too?

ilanbiala:
@ppotoplyak one last change, but this PR should actually be against the 0.4.0 branch and not against master. Can you create a new one and I'll merge that one in?

mleanos:
Does this really need to be implemented into the User model? It seems that the most common use case would be to use this voting feature in the context of an Article.

What I mean by this is that the relevancy of the article vote makes more sense when interacting with the actual Article model itself.

If we want to keep track of the User's that have voted, then just track that in the Article model. I'm not a fan of this feature separating the up & down votes. Shouldn't this be a cumulative value?

This is my proposal for the Article model implementation... 

    votes: [{
        user: {
            type: Schema.ObjectId,
            ref: 'User'
        },
        value: {
            type: Number,
            default: 0
        }        
    }]

We could add date fields to this schema as well. What worries me about this pull request is that it seems very specific to one use case, and there's unnecessary updates to the User model, and User server controller. My proposal keeps things simple, and can be modified to meet individual needs.

lirantal:
I will close this one since we already have another PR for 0.4.0 that is waiting @ppotoplyak feedback for some adjustments before merging.

PR #538, opened 2015-04-29
Update grunt-nodemon, livereload only when server has restarted.

richardlt:
When a change is detected on server JS and HTML files, grunt-watch livereload the site but nodemon also restart the server and cause white screen on web browser. 
The grunt-nodemon readme contains an exemple to create a ".rebooted" file that change when the server has restarted. Then livereload will be executed by watching only the ".rebooted" file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ public/lib/
 app/tests/coverage/
 .bower-*/
 .idea/
+.rebooted
 
 # MEAN.js app and assets
 # ======================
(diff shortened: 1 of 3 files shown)


ilanbiala:
This PR should be against 0.4.

richardlt:
You mean on the 0.4.0 branch of Mean. Why ?

ilanbiala:
Create a PR against the 0.4.0 branch, because we are releasing 0.4.0 soon.

PR #541, opened 2015-04-30
Update grunt-nodemon and livereload only when server has restarted

richardlt:
When a change is detected on server JS and HTML files, grunt-watch livereload the site but nodemon also restart the server and cause white screen on web browser. 
The grunt-nodemon readme contains an exemple to create a ".rebooted" file that change when the server has restarted. Then livereload will be executed by watching only the ".rebooted" file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ public/dist/
 .idea/
 uploads
 modules/users/client/img/profile/uploads
+.rebooted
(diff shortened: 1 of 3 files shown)


ilanbiala:
@lirantal initial feedback?

lirantal:
I'm not much of a frontend person so I didn't bump into this situation but sounds like this fix would make it much easier and quicker to develop for frontenders. Can someone else test it beside @richardlt ?

codydaig:
I have not tested this. but if @richardlt submits another pull request against the latest master, Id be willing to test and as long as there are no bugs, id like to see this get merged in.

PR #549, opened 2015-05-06
Hide email address

vepasto:
Email address should not be shown to client


modules/users/server/controllers/users/users.password.server.controller.js
```diff
@@ -74,7 +74,7 @@ exports.forgot = function(req, res, next) {
 			smtpTransport.sendMail(mailOptions, function(err) {
 				if (!err) {
 					res.send({
-						message: 'An email has been sent to ' + user.email + ' with further instructions.'
+						message: 'An email has been sent to users email with further instructions.'
```
ilanbiala: Why not just `the provided email`?

ilanbiala:
LGTM.

PR #565, opened 2015-05-16
Add fonts and svgs to gzip compression pattern

simison:
I think it's good to have these here "just in case", even the seed project wouldn't have any fonts/svgs. These are just so common filetypes at any project.

Thoughts?


lirantal:
Looks ok I guess.
@roieki @ilanbiala ?

ilanbiala:
LGTM.

PR #566, opened 2015-05-18
#501 Handle 404 errors at Express backend and at Angular frontend

simison:
- `/{api|modules|lib}/*` returns an error page when path doesn’t exist
(from Express).
- `/*` always returns index (from Express), but if `$state` doesn’t
exist, Angular redirects to `/not-found` (no 404 status in that case
though!)
- If `Accept: application/json` header is present without `Accept:
text/html`, return error as json. Hence looking at non existing /api/*
paths with browser would show html error, but querying them with script
would return json.
- Slightly prettier 404 error

Test:
```bash
curl http://localhost:3000/api/notfound -4 -H "Accept: application/json"
```
=> json error.

```bash
curl http://localhost:3000/api/notfound -4 -H "Accept:
text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0
.8"
```
=> html error (imitates Chrome’s Accept header).

Starting point was @dotch’s PL: https://github.com/meanjs/mean/pull/503

### DIFF
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -177,7 +177,6 @@ module.exports.initModulesServerRoutes = function (app) {
  * Configure error handling
  */
 module.exports.initErrorRoutes = function (app) {
-	// Assume 'not found' in the error msgs is a 404. this is somewhat silly, but valid, you can do whatever you like, set properties, use instanceof etc.
 	app.use(function (err, req, res, next) {
 		// If the error object doesn't exists
 		if (!err) return next();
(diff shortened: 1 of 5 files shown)


lirantal:
looks valid, need more eyes.
@jloveland @ilanbiala @miguelcoba @mleanos

lirantal:
LGTY @codydaig @rhutchison ?

rhutchison:
LGTM

lirantal:
Thanks @rhutchison , @simison, merging.

mleanos:
+1 LGTM

PR #568, opened 2015-05-20
Fix the Google OAuth Redirect Bug - #535

### LINKED ISSUE #535: google authentication callback url bug
Hi,
I get URL like localhost:3000/# when I use google authentication with passport.

I fixed the problem using the same way which has been used to fix the Facebook bug with redirect (public/modules/application.js):

```
//Then define the init function for starting up the application
angular.element(document).ready(function() {
	// Fixing facebook bug with redirect
	if (window.location.hash === '#_=_') window.location.hash = '#!';

	// Then init the app
	angular.bootstrap(document, [ApplicationConfiguration.applicationModuleName]);
});
```

I added this part:

```
// Fixing google bug with redirect
	if (window.location.href[window.location.href.length - 1] === '#' &&
		// for just the error url (origin + /#)
		(window.location.href.length - window.location.origin.length) === 2) {
		window.location.href = window.location.origin + '/#!';
	}
```

Does you know something about it?

Thanks.

floydpink:
Correct the issue of Google OAuth2 redirect missing the hashbang after enabling GooglePlus API - Fixes #535

Thanks to the discussion on #535 and the fix from @junit38

### DIFF
--- a/public/application.js
+++ b/public/application.js
@@ -15,6 +15,13 @@ angular.element(document).ready(function() {
 	//Fixing facebook bug with redirect
 	if (window.location.hash === '#_=_') window.location.hash = '#!';
 
+	// Fixing google bug with redirect
+	if (window.location.href[window.location.href.length - 1] === '#' &&
+			// for just the error url (origin + /#)
+			(window.location.href.length - window.location.origin.length) === 2) {
+			window.location.href = window.location.origin + '/#!';
+	}
+
 	//Then init the app
 	angular.bootstrap(document, [ApplicationConfiguration.applicationModuleName]);
 });


floydpink:
Without this change, G+ auth is broken now. What else do we need for this to get merged?

Who can give the +1 indicated by the label?

pinny87:
:thumbsup:
Implemented your changes and fixes the redirect issue for my app.

Thanks

ilanbiala:
@floydpink this should be against 0.4.0 if it applies in that branch.

lirantal:
@ilanbiala I actually merged this into master unintentionally (too many open tabs) although I'm not sure we need to revert if we'll anyway;
1. release 0.4.0 soon
2. @pinny87 confirming this works already in master

PR #575, opened 2015-05-28
Add some abstraction to local strategy login error

pgrodrigues:
This is implemented in master but it is not in 0.4.0.

This basically prevents username enumeration by providing the same error message when the username doesn't exist or when the username/password combination is wrong.

### DIFF
--- a/modules/users/server/config/strategies/local.js
+++ b/modules/users/server/config/strategies/local.js
@@ -20,14 +20,9 @@ module.exports = function() {
 				if (err) {
 					return done(err);
 				}
-				if (!user) {
+				if (!user || !user.authenticate(password)) {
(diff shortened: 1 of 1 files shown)


ilanbiala:
I would rather say that the account doesn't exist, and then say that it's an invalid email or password if it does.

pgrodrigues:
From a user point of view, ofc it looks nicer to know in case if you did a mistake while logging in wether your username or password is wrong.

However for a user with bad intentions this can be seen as a simple and effective way to prevent him from building a dataset with registered usernames.

It's an option each developer can make.

ilanbiala:
@lirantal thoughts?

lirantal:
Looks ok

Issue #352, opened 2015-01-10
Swagger integration


quocvu:
I would like to have Swagger to document the server API integrated to MEAN.JS

https://helloreverb.com/developers/swagger

Really missed it (I used to be on Loopback)

jloveland:
+1

frederik:
+1

Have you looked at raml? I've only used Swagger so far but RAML looks really nice (especially the notebooks that are showcased in the video). This could be really great to get started with the article api from meanjs.

http://raml.org/

ilanbiala:
If anyone wants to create a PR for Swagger documentation, I can get it merged in.

ilanbiala:
This seems like extra that most users won't need, I'm going to close for now because there are also better features we should spend more time on.

PR #578, opened 2015-05-29
RAML checker

sielay:
This change introduce first stage of integration with RAML to document and validate API.

 * adds task `gulp raml` that will check Express routes definitions against RAML and other way around
 * I left two invalid RAML rules to show you output

In further steps we can introduce tests with [Abao](https://github.com/cybertk/abao/)

Benefits:
 * Keeps you with up to date API documentation and tells you that you need to update it
 * RAML allow you to generate quite powerful docs and testing consoles (it's far more mature than Swagger)
 * Allows you to use testing like Abao which can run schema validation against responses

`Note`: I use tabs as expected. Anyhow my dumb InteliJ don't allow tabs in YAML files. Maybe there is reason for it.


lirantal:
Hi @sielay 
What's RAML and why do we need it? what does this PR affect except for adding some documented API and updating gulp to support it?

sielay:
@lirantal RAML is most used standard to document RAML apis. For my projects it allows to introduce tools like Abao to gain close to 100% server side code coverage in tests.

codydaig:
IMO, it seems to fit a specific type of implementation and just adds another layer of "documentation" that would need to be supported.

codydaig:
There was also an initial discussion on this in #352

lirantal:
@sielay I'm ok with looking into this but let's do that as part of 0.5.0 which will take more focus on documentation, testing, etc.

PR #559, opened 2015-05-12
0.4.0 - refactor_user

sielay:
Hi. It took time, but this version __IS__ aligned to 0.4.0 and it would be quite clever to merge/agree on it before any further changes to users. I don't remember such a nasty merge so far ;)

This changes introduce major changes:
 * there is no providerData or additionalProviderData in user objet
 * each identity including email(s) and username(s) are stored in one array, and contain own data
 * tokens are being updated on next successful login

This includes:
 * user explorer for users with `admin` role in `/admin/users`
 * updates to oauth accounts and emails

This pass `grunt test` for me and work in currently developer project.

This was tested against actual integration with facebook, google+, twitter, linkedin and github.

PR #580, opened 2015-05-31
Oauth2config (depends on #559)

sielay:
Change dependant on https://github.com/meanjs/mean/pull/559

Aim is to move oauth switch out of routing. It should allow to create dynamic oauth setup and allow to re-use some controllers in case you connect to various endpoints, i.e. GitHub and GitHub Enterprise.


lirantal:
@sielay  Let's review this after #559 is merged.

PR #581, opened 2015-06-01
Patch 1

hougie:
1. Fig is obsolete now and has been replaced with Docker Compose.

    The DOCKER_MONGO_URI enviroment variable was added into the docker-compose.yml. 
     It has the value of "mongodb://db/mean-dev'".

2. When the containers are created and run in "development" mode, the "web" container cant connect  to the "db" container.

Currently, development mode mongo uri is hard coated to "mongodb://localhost/mean-dev". However, the web container is trying to connect to mongo @ mongodb://db/mean-dev',

If we change the mongo uri in the development.js to the following, the web container will connect to the right path. It will also run fine if you just run the grunt command locally.

uri:   process.env.DOCKER_MONGO_URI   ||  'mongodb://localhost/mean-dev',


ilanbiala:
@hougie there are already PRs that fix this that I haven't merged yet because I wanted to verify, but it seems like everyone is reaching the same consensus. You should search for that PR and just make sure everything there is what you have here and I'll merge that in since it has more subscribers to the PR.

hougie:
Ok, sounds great.

Thanks,
Peter

PR #586, opened 2015-06-04
Remove targets on signin and signup anchors

cdriscol:
Removing target on signin and signup anchors to prevent a complete page reload when changing to those states.

### DIFF
--- a/modules/core/client/views/header.client.view.html
+++ b/modules/core/client/views/header.client.view.html
@@ -25,11 +25,11 @@
 		</ul>
 		<ul class="nav navbar-nav navbar-right" data-ng-hide="authentication.user">
 			<li data-ui-sref-active="active">
-				<a data-ui-sref="authentication.signup" target="_self">Sign Up</a>
+				<a data-ui-sref="authentication.signup">Sign Up</a>
 			</li>
 			<li class="divider-vertical"></li>
 			<li data-ui-sref-active="active">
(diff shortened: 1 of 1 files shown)


ilanbiala:
@cdriscol can you provide a source for what this fixes in Angular/in general?

cdriscol:
@ilanbiala, I noticed that on the header the signin and signup were causing page refreshes because they were using a target="_self".  Since those links just change state using data-ui-sref, they do not need to cause this page reload.

I looked at meanjs/master branch - and the behavior there is to change state and not reload.  So I'm not sure how/why the target's were added in 0.4.0 - but they're not needed (from what I can tell) and it negatively impacts the user experience by causing a page reload rather than a smooth state transition to the signin/signup partials.

cdriscol:
@ilanbiala, I should also note that I was going to create an issue for this, and still can, if that's what we should be doing.  With the size of this change, I wasn't sure if creating the issue was just extra overhead..

lirantal:
@ilanbiala  if @cdriscol can confirm that this doesn't break anything then I see no harm in merging. what do you think?
@cdriscol I assume you can run some tests locally and confirm please?

cdriscol:
@lirantal, everything works as expected after this change.

lirantal:
@ilanbiala LGTY?

ilanbiala:
Yep, you can merge.

PR #587, opened 2015-06-04
removed /#! in user controllers

ycmjason:
Well in 0.4.0 obviously we dont want /#!/ :dancer:

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -84,11 +84,11 @@ exports.oauthCallback = function(strategy) {
 	return function(req, res, next) {
 		passport.authenticate(strategy, function(err, user, redirectURL) {
 			if (err || !user) {
-				return res.redirect('/#!/signin');
+				return res.redirect('/signin');
 			}
 			req.login(user, function(err) {
 				if (err) {
-					return res.redirect('/#!/signin');
+					return res.redirect('/signin');
 				}
 
 				return res.redirect(redirectURL || '/');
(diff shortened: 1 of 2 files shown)


ilanbiala:
LGTM, but are there any other places that need to be fixed? What about SEO? Does this affect that at all?

lirantal:
@ycmjason looks ok, can you fix your PR?

rhutchison:
imho, this should not be merged.  This will break routes when html5mode is disabled or not supported.

lirantal:
@rhutchison but if that were the case then wouldn't we need to add the #! everywhere else on other URLs?

ilanbiala:
@rhutchison I'm not sure why someone would disable html5mode and most browsers support it today anyway.

rhutchison:
@ilanbiala I'm not sure why someone would disable it, but I am identifying it as a concern.

@lirantal in most cases, our href's are generated through ui-sref which considers pushState support and the html5mode flag.

Since this is a change on the server-side, and there are no good methods to detect browser capabilities, it's best to leave as-is, unless we decide to drop support for unsupported browser (http://caniuse.com/#feat=history).

ilanbiala:
I don't mind dropping IE8 and IE9, anyway Angular 1.5/2.0 is probably going to drop support there..

lirantal:
you won't hear me complaining about IE9 :-)
if this cleans stuff up then I'm forward with this because those messy URLs look awful in code and doesn't look like they are supposed to be there.

unless this is truly breaking something I'm ok with this PR.

rhutchison:
I am ok with dropping support for IE9, but I think we should update documentation to note that we only support [evergreen browsers](http://eisenbergeffect.bluespire.com/evergreen-browsers/).

ilanbiala:
@rhutchison I don't think we need to do just Evergeen, but I would say IE10+ and latest of Safari.

codydaig:
The percentage of IE users are slowly dropping. I think it's reasonable to say IE10+.

lirantal:
So if we're ok with cleaning up the URLs then I've got the updated PR for it: https://github.com/meanjs/mean/pull/693

Issue #612, opened 2015-07-02
Update to grunt-nodemon 0.4.0


trainerbill:
Can we update grunt-nodemon to 0.4.0 on the mean 0.4.0 branch?  It will fix the issue of watching too many files:

https://github.com/ChrisWren/grunt-nodemon/issues/46

Happens to us Ubuntu users.

ilanbiala:
@trainerbill create a PR after 0.4 is officially released. This will probably have to wait about a week and it will be released as part of 0.4.1 or slightly after.

lirantal:
right, it's there and we can probably close this issue thanks to @rhutchison PR.

PR #622, opened 2015-07-05
update dependencies

rhutchison:


ilanbiala:
@rhutchison this should wait until 0.4.0 is released because we don't need this to release. In fact, I'd rather try to sort out whether we are going to support Grunt or Gulp before doing this, because there is no point in having support for both as we just need to maintain more.

rhutchison:
@ilanbiala are you in the gitter chat?  This PR is against your 4.0 branch.  Should update dependencies and then figure out gulp vs grunt.  Some of these dependencies are very old and if you are releasing 4.0, it should be tested against newer libs.

rhutchison:
Resolves https://github.com/meanjs/mean/issues/612

lirantal:
@rhutchison we'll get to review this PR soon, no worries.

lirantal:
hi @rhutchison I want to merge this but there's a bit of a mess there with the commit messages like reverts and some other "mistake commits" as well as some space indentation that is wrong (in the package.json file for example) and some spaces that were added but aren't really a change.

I don't want to merge with the current commit history, can you please fix this up quickly and squash your commits to logical commits that make sense with the changes?

Please hurry doing so as I want to merge this in to master_to_0.4.0 and quickly after merge that one into 0.4.0.

modules/core/client/views/header.client.view.html
```diff
@@ -33,12 +30,12 @@
 			</li>
 		</ul>
 		<ul class="nav navbar-nav navbar-right" data-ng-show="authentication.user">
-			<li class="dropdown">
-				<a href="#" class="dropdown-toggle user-header-dropdown-toggle" data-toggle="dropdown">
+			<li class="dropdown" dropdown>
```
lirantal: is that dropdown text suppose to be there like that?

rhutchison: Yes, it's how it works with ui-bootstrap.

rhutchison:
@lirantal package.json is written to by npm --save and --save-dev

Please see https://github.com/npm/npm/issues/4718

I updated .editorconfig - https://github.com/Gym/mean/commit/73be7943b169db26c83e62956c06fdd32c2684ac

lirantal:
@rhutchison with regards to package.json, you re-ordered all of the packages so the diff looks like this now:

image

and it's hard to figure out which package you added, or removed, or just updated the version for because they are all re-organized again in the new commit.
Also, was there a specific reason for updating the dependencies?

@ilanbiala WDYT?

simison:
@rhutchison @lirantal I think we should let the package.json re-organising and spacing go trough now and it'll be easier in the future. When working with package.json trough command line, that's what happens. If we don't let this ugly diff go trough here now, people will have it ahead of them when they'll fork and add/remove packages at their version.

ilanbiala:
@lirantal the package.json reordering makes sense, but right now it definitely makes it harder to diff. I say we update deps using https://david-dm.org/ and then allow npm to format the package.json in a different PR.

rhutchison:
npm is responsible for sorting the packages alphabetically.  Here is the list of updates prior to applying this commit.

2015-07-08 10_22_18-administrator_ command prompt

ilanbiala:
@rhutchison I'm not saying that npm isn't doing something logical, I'm saying that we should manually increment the versions based on david-dm and let npm format the package.json properly in a different commit or even PR, because that will simplify reviewing.

lirantal:
Ok then, I'll go ahead ahead and merge this in, we can indeed resolve the package.json later.

PR #632, opened 2015-07-08
Remove duplicate assets definition

danaronoff:
Looks like this sneaked in during the merge with master. Not sure why it would be needed since we have the assets file.


lirantal:
indeed, thanks!

PR #634, opened 2015-07-09
[Bug]: Remove social account.

rhutchison:
The settings controller was moved under settings directory, so I removed the duplicate controller.

There is a bug with removing social accounts due to the undefined "params" on the request.  This is changed to use "query".  While debugging this, I noticed if there are no params passed, the request just times out due to there being no response.  I added some exception handling to throw 401 for unauthenticated user and 400 for no params.


lirantal:
need someone with more angular experience to review this

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -175,33 +175,39 @@ exports.saveOAuthUserProfile = function(req, providerUserProfile, done) {
...
+			req.login(user, function (err) {
+				if (err) {
+					return res.status(400).send(err);
+				} else {
+					return res.json(user);
+				}
+			});
+		}
+	});
+};
```
ilanbiala: Newline missing

rhutchison: What's the issue with newline?  Why is it required?

ilanbiala:
@lirantal LGTM. The file that is deleted was just left as a remnant when @amoshaviv split it up into smaller chunks in a folder called settings.

lirantal:
thanks @ilanbiala, merged.

Issue #490, opened 2015-03-24
Mean JS Angular App to Use Style Guide


britztopher:
I am wondering if we could start following the john papa style guide for the angular application.  It proves to serve a lot of great ideas and fundamentals.  

https://github.com/johnpapa/angular-styleguide

ilanbiala:
@britztopher Any initial changes that you think should be made?

SOSANA:
Maybe I am miss understanding but we are doing this already for v4.0 we changed the file structure similar to what he recommends...

around 4:49
![image](https://cloud.githubusercontent.com/assets/2634185/6794567/f807e876-d198-11e4-8121-4fa49bb1bfac.png)

John Papa - 10 AngularJS Patterns - Code on the Beach 2014
https://www.youtube.com/watch?v=UlvCbnKAH3g

ilanbiala:
I'm sure there are other things that could be changed, but I don't think everyone will necessarily agree with everything he recommends.

britztopher:
@ilanbiala I would say these would be a good start:
1. Use IIFE's - https://github.com/johnpapa/angular-styleguide#iife  
2. Controllers - https://github.com/johnpapa/angular-styleguide#controllers
  a.  the whole section is really good and would be that hard to change some files
3. services - https://github.com/johnpapa/angular-styleguide#services

I mean I seem to follow the whole guide with a little deviation from folder/file structure, which makes code a lot easier to read and maintainable.  Also, the style guide does point out some good reasons why to do certain things.  Like why to use IIFEs.

@SOSANA the structure is on the ball, however, internals are not.

Shrizzy:
Great idea @britztopher
One of the first things I do with a new meanjs app is split up the controllers into smaller components, remove the ng-init's and change the controllers to 'Controller as'. 
I'd really like to see the Controller As syntax, and removed references to $scope as we edge closer to Angular 2.0.

aarosil:
This not entirely related but I think an awesome feature would be to be
able to provide your own templates to the generator.  Then one can follow
whatever style one likes or needs.

I often scaffold an app, then customize one module for specific needs of
that app, and after that I use `yo` to generate a new vertical module, and
replace the autogenerated files with what I customized and do a global
search and replace for the object name.

wormyy:
+1 for using John Papa's style guide.

simison:
Sounds like we could finally get rid of pre written annotations as well:

write:
```
[...]

PR #636, opened 2015-07-10
[WIP] ngBestPractices

rhutchison:
This PR is to refactor the angular application to conform with the [angular style guidelines](https://github.com/johnpapa/angular-styleguide).

This is up for discussion - please provide feedback in comments.

### TODO
- [x] Articles
 - [ ] once core is complete, change articles.menus dependency to core.menus
- [ ] Chat
- [ ] Core
- [ ] Users

After the articles module is reviewed, and if the decision is made to enforce these guidlines, I will complete the remaining modules (core probably should have been first).


simison:
I skimmed this trough just now. Looks really neat and pretty much along the lines how I would've done it. Great!

---
PS. Here's a list for your initial message if you'd prefer checkboxes:

- [x] Articles
 - once core is complete, change articles.menus dependency to core.menus
- [ ] Chat
- [ ] Core
- Users
 - [ ] Edit profile
 - [ ] Change profile picture (there's a lot of non-view-controller logic going on here right now...)
 - [ ] Manage social
 - [ ] Settings
 - [ ] Auth: login/signup
 - [ ] Password stuff

rhutchison:
@simison thanks for the feedback and github tip

simison:
Related to #490.

rhutchison:
@jloveland

modules/articles/client/controllers/articles.client.controller.js
```diff
@@ -1,68 +1,52 @@
...
+		// Save Article
+		function save() {
+			var article = vm.article;
 
-				for (var i in $scope.articles) {
-					if ($scope.articles[i] === article) {
-						$scope.articles.splice(i, 1);
-					}
-				}
+			if (article._id)
```
danaronoff: @rhutchison Following up from Gitter, this was the line I was referring to. Here we check if the article has an ID to then decide to save or update. I might be misinterpreting the Style guide, but one could see this as being too tied to the data service - the controller shouldn't know what having an _id means. It could be refactored so that the controller's save() calls a method on a service named saveOrCreate() which then examines the object to determine what path to take. I understand this is probably a matter of opinion but wanted to let you know my thoughts.

rhutchison: You are correct - I will make a change.

ilanbiala:
If this won't change functionality, there is no reason we need to hold up 0.4.0, so either we can make it part of 0.4.x or if not, we'll release a quick 0.5.0 with less changes, which would probably be better anyway.

ilanbiala:
I'm for a more consistent styleguide, but I don't like some of the rules in John Papa's. @lirantal do you feel the same way?

lirantal:
@ilanbiala what kind of styles you don't like specifically?

ilanbiala:
IIFEs, and some other smaller things that I can't find under a specific section.

modules/articles/client/articles.client.module.js
```diff
@@ -1,4 +1,10 @@
-'use strict';
+(function (app) {
```
ilanbiala: I don't really prefer using IIFEs. It seems unnecessary and just extra for very little benefit.

modules/articles/client/config/articles.client.menus.js
```diff
@@ -1,8 +1,17 @@
-'use strict';
+/**
```
ilanbiala: Unnecessary IMO.

ilanbiala:
@rhutchison the indentation seems off, no? This hasn't been updated for that. Also, I don't really think the comments at the top of each file are necessary. I'm not a fan of IIFEs and $inject. What do you think about using ngAnnotate?

simison:
+1 for ngAnnotate. Makes life easier but in Meanjs case it would require bundling all the Angular code into one file even in development mode.

modules/articles/client/config/articles.client.menus.js
simison: What do you think about ngDoc or jsDoc for these? I believe this project should encourage documenting code more.

ilanbiala:
@simison I actually use 1 file for development and production and just have sourcemaps work it out for me. I think it works fine, is there any reason not to do that? It would also simplify the asset management in my opinion.

simison:
@ilanbiala I also think it's the way to go. Just pointing out it requires some work with configs/all that jazz.

modules/articles/client/articles.client.module.js
mleanos: I agree.. I'd prefer we don't use IIFE's like this. The way we're loading our client js files, is essentially the same behavior. This makes it harder to read, without a recognizable benefit.

I'm a huge fan of John Papa's angular style guide, but I don't think we have to take everything that he recommends.

modules/articles/client/config/articles.client.menus.js
ilanbiala: Comments are good, but this is unnecessary. If at any time you want to know what the file is, you can just look at the title, but the comment only helps if you can see it and anyway it doesn't add anything to the code.

modules/articles/client/articles.client.module.js
ilanbiala: @rhutchison can you remove all IIFEs?

modules/articles/client/config/articles.client.menus.js
ilanbiala: @rhutchison as long as the file names are good, this is just extra. Let's try to keep it as clean as possible.

modules/articles/client/controllers/articles.client.controller.js
ilanbiala: @rhutchison has this been fixed?

ilanbiala:
@rhutchison is the indentation also outdated?

codydaig:
What was decided about IIFE's?

ilanbiala:
@codydaig unless anyone can provide compelling evidence in this framework for the need of IIFEs, they are unnecessary and I'd prefer to avoid the complexity.

mleanos:
:+1: on @ilanbiala assessment of IIFE's in this project.

ilanbiala:
Final decision: no IIFEs. We can put a page or something on the wiki, but it's unnecessary in the project itself.

PR #638, opened 2015-07-10
Gulp updates

rhutchison:
These are updates to gulp build automation.

### Gulp Template Cache

### Auto Prefixer

### Path Separator
There is an issue on windows with path separator when compiling less/sass

https://github.com/Gym/mean/commit/43fade02d2bfde802666450e3e37f644b289d670#diff-b9e12334e9eafd8341a6107dd98510c9R126

### Grunt vs Gulp
I have localized automation configuration changes to their respective config so that gruntfile.js changes do not restart gulp process and vice versus.

https://github.com/Gym/mean/commit/43fade02d2bfde802666450e3e37f644b289d670#diff-595bf3fa2348192244b0319be33066b8


.jshintrc
```diff
@@ -9,7 +9,7 @@
 	"curly": false, // Require {} for every new block or scope.
 	"eqeqeq": true, // Require triple equals i.e. `===`.
 	"immed": true, // Require immediate invocations to be wrapped in parens e.g. `( function(){}() );`
-	"latedef": true, // Prohibit variable use before definition.
+	"latedef": "nofunc", // Prohibit variable use before definition.
```
rhutchison: This change is to support angular best practices (as implemented in templates.js).  Taken from [angular-styleguide](https://github.com/johnpapa/angular-styleguide#style-y230)

rhutchison:
github was having issues, reopening to trigger clean build.

config/env/production.js
```diff
@@ -2,7 +2,7 @@
 
 module.exports = {
 	secure: true,
-    port: process.env.PORT || 8443,
+	port: process.env.PORT || 8443,
```
ilanbiala: Already handled in another PR

ilanbiala:
Grunt is being changed here as well. Why?

rhutchison:
@ilanbiala To ellaborate further on "Grunt vs Gulp", since allJS (assets definition) contained gruntfile.js as one of the src files, any changes to gruntfile.js would trigger a reload of gulp.  This PR addresses this by separating the two.

See the following:

 - https://github.com/Gym/mean/commit/43fade02d2bfde802666450e3e37f644b289d670#diff-e46cd636275276017cc9fdf3097ea0b6L39
 - https://github.com/Gym/mean/commit/43fade02d2bfde802666450e3e37f644b289d670#diff-595bf3fa2348192244b0319be33066b8R81

ilanbiala:
@rhutchison the path separator should probably be a separate PR. Also the template cache. Mixing all this into one PR makes it hard to analyze both to see if there's anything that can be optimized and any bugs that may come up.

PR #640, opened 2015-07-10
cleaning up application startup information

lirantal:
On starting up the server:
1. Remove the 'application started' suffix to the application title
2. Align the database name tab length
3. Remove the initial 'Application loaded using <NODENEV> environment' as it is anyway reported in the startup info.

Startup information after the fix:
image

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -61,8 +61,6 @@ var validateEnvironmentVariable = function() {
             console.error(chalk.red('NODE_ENV is not defined! Using default development environment'));
         }
         process.env.NODE_ENV = 'development';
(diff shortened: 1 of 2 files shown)


danaronoff:
Is mean-test now the DB for the dev environment? I noticed that changed during the merge. It used to be mean-dev.

lirantal:
you're right, I submitted another PR for that: https://github.com/meanjs/mean/pull/641

lirantal:
I guess we can merge that?

ilanbiala:
Yep, looks good.

PR #641, opened 2015-07-10
updated the development.js environment configuration to use mean-dev database

lirantal:
updated the development.js environment configuration file to use the mean-dev database instead of mean-test which is used for testing/stages environment


rhutchison:
Good to merge - thanks

ilanbiala:
LGTM.

Issue #295, opened 2014-12-08
User passwords keep changing


JoaoPintoM:
Hello,

I'm playing with meanjs 0.3 and I have to recreate new users because the passwords seems to change all the time.. Sometimes it works sometimes not and works again.. I have no idea what is happening.

Thx in advance if you have any idea about that issue.

Joao

smirnovigor:
@yesnault Thank you, it's helped me.

ilanbiala:
@smirnovigor @yesnault does #253 cover the same thing that you guys are talking about?

smirnovigor:
@ilanbiala No, this issue happens on update user with already encrypted password field. Then encrypted password pass again through pre-save model cycle and encrypt encrypted password.

ilanbiala:
@smirnovigor the issue seems like the if-else has to be reversed in [this method](https://github.com/meanjs/mean/blob/master/app/models/user.server.model.js#L109), right?

smirnovigor:
No, it's more complicated. You shouldn't re-encrypt password and salt on every save. Take a look to the @yesnault [commit](https://github.com/docktor/docktor/commit/1f00332b4dfbdd8de992c8cd94156676171e869e)

ilanbiala:
@smirnovigor I think it makes more sense to check before saving the document instead of deciding when to check.

The issue is actually that we keep generating a new password because we are using a new salt. Also, it's not really clear to me why we are passing in a password that may have already been hashed (in the case of an update to the document).

@lirantal maybe you can clear up some of these questions?

lirantal:
I've no idea what password re-encryption is being discussed here.
@smirnovigor :
1. can you provide the steps to reproduce the problem? I saved the user profile and the password isn't changing.
2. that commit you referred to is out of the date from meanjs perspective as the changes there are not reflecting the current state of meanjs 'master' branch.

ilanbiala:
@lirantal it's because we do a save on the `_.extend`ed User document that the pre save middleware gets called and the password will be re-hashed it seems.

lirantal:
@ilanbiala tell me how to reproduce it?

ilanbiala:
If the pre save middleware really is kicking in, I think you just have to update a user and change anything but the password. Keep track of the password before and after.

lirantal:
@ilanbiala read my reply to smirnovigor from yesterday - I tried that and couldn't reproduce the problem. Unless someone can reproduce it I'm closing.

[...]

PR #642, opened 2015-07-11
Fixing user's passwords keep changing when saving the user model

lirantal:
Addresses issue: https://github.com/meanjs/mean/issues/295
updating the schema save pre hook so that it checks for a modified version of the password field before it tries to re-calculate the new password to save for the user model

### DIFF
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -67,6 +67,16 @@ describe('User Model Unit Tests:', function() {
 				done();
 			});
 		});
+
+		it('should confirm that saving user model doesnt change the password', function(done) {
+			user.firstName = 'test';
+			var passwordBefore = user.password;
+			return user.save(function(err) {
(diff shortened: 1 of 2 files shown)


lirantal:
This PR also adds a test to check for this problem in the future:
image

modules/users/tests/server/user.server.model.tests.js
```diff
@@ -67,6 +67,16 @@ describe('User Model Unit Tests:', function() {
 				done();
 			});
 		});
+
+		it('should confirm that saving user password doesnt change the password', function(done) {
```
rhutchison: Should this read 'should confirm that saving user *model* doesnt change the password'.  Also, should we have a test for change password now?  I know these tests need to be built out more, in general, but these are a good start.

lirantal: thanks @rhutchison, my typo :)
I'll definitely add more in the future.

Issue #443, opened 2015-03-02
Documentation for 0.4 branch


darko-mijic:
Hello,

Does documentation for 0.4 branch exist? 
I would like to see the draft version and I would like to contribute.

Regards,
Darko

Shrizzy:
There's no specific documentation for 0.4 yet, but you can find the current documentation on the gh-pages branch. If you're keen, perhaps start with moving across the reusable parts of the current doco into the docs/0.4 folder.

I've noticed a lot of repos using Jekyll, and Kramdown for documentation lately. An example is express.js: https://github.com/strongloop/expressjs.com

codydaig:
Has there been any work on the docs for 0.4.0 or shall I get started setting up a Jekyll solution and populating it with content that applies from the previous version?

codydaig:
Has anybody started on the 0.4.0 docs yet? Can I set-up a Jekyll solution and get rolling on moving the old documentation that still applies over?

Shrizzy:
I haven't seen much happening with the docs. If you're able to kick things off, submit a PR and we'll go from there. 
Similar to Issue #294

codydaig:
See Pull Request #649

lirantal:
@codydaig yes, looks good.
please keep up the 0.4.0 documentation there and we also need to get more reviewers on that PR.

codydaig:
Should the 0.4.0 documentation follow the same format and structure as the current documentation? Does anybody have any changes they would like to see happen?

lirantal:
specifically I don't have any objections

ilanbiala:
Condensed a few issues about 0.4 into this issue. For anyone who would like to contribute documentation to 0.4.0 @codydaig has helped us set up a Jekyll gh-pages site, so help us tie up loose ends and contribute any documentation you can!

codydaig:
I submitted a pull request #652 which demonstrates how documentation can be added. I also restructured the permalinks which I outlined and explained in the pull request.

My current question for discussion: I have not used the Yo Generator very much, so I would like some guidance. I'm assuming that the current version of the generator will still be accessible once 0.4.0 is released. With this assumption, I would structure the Yo Generator section the same as the documentation.
I'd appreciate some feedback and guidance on the structure and on the yo generator section. 
Thanks!

ilanbiala:
[...]

Issue #294, opened 2014-12-08
Update docs to v0.4


JuanCaicedo:
I get the impression that the docs still reflect v0.3, as this is still what is in the master branch. They will have to be updated for the new vertical structure, including the yeoman commands which have changed in the generator-meanjs project. 

I'm willing to start looking at this and make some PRs if that would be helpful!

ilanbiala:
@JuanCaicedo any docs that you make for 0.4.0 would be extremely helpful, as we are planning to finalize 0.4.0 very soon and formally release it.

JuanCaicedo:
@ilanbiala For sure, I can put something out for that. For clarification, should I send PRs here to generator-mean-js?

ilanbiala:
Here would be better.

codydaig:
@ilanbiala Can we close this Issue and move 0.4.0 documentation discussions to #443 ?

PR #649, opened 2015-07-15
Migrated the Website to Jekyll and set-up for 0.4.0 documentation

codydaig:
Reference Issues #294 #443


codydaig:
I didn't change the content of the website, just migrated it into a modular format around jekyll and set-up the folder structure for the docs for 0.4.0. I'm completely open for feedback if a different structure or method is desired. This was just the quickest way to get it up and running to match it's current state and easily add new documentation in a modular form.

codydaig:
You can view how it will render on GitHub Pages here:
http://codydaig.github.io/mean/

CNAME
```diff
@@ -1 +0,0 @@
-meanjs.org
```
lirantal: why is the CNAME entry removed?

lirantal:
I guess it looks ok, but probably need another set of eyes of someone who had commited in the past to gh-pages

CNAME
codydaig: I made that change after I submitted the pull request to make it render on my GitHub Pages for demonstration. I forgot that it would have added it too the pull request. My apologies. Once it's approved too be merged let me know and I'll change it back.

lirantal: Cool

ilanbiala:
@codydaig what's the difference, functionally, between the Jekyll setup and the current gh-pages branch we have set up?

codydaig:
The main difference is the ability to use templates which allows areas such as the documentation to be broken up into sections. Makes the code cleaner and easier to maintain. The current website has the same header and footer elements copied into each individual html page, so if you need to make one change, you have to change every page. With jekyll, you just change it once in the template and it's changed on all pages. 
As for breaking up the code into sections, I implemented this on the documentation page. So instead of it being around 3.,000 lines of code, it's broken down into 17 pages which is easier to find, add, and modify content. Every section (designated with a link on the sidebar) is it's own .html page but is rendered to the user as one solid page just like it was before.

ilanbiala:
@lirantal I think it's a good idea. WDYT?

lirantal:
Sure, @codydaig can you please have another walk-through on the rendered page, some general sanity to see that nothing is broken and I'll be happy to merge.

ilanbiala:
@lirantal do we have any 0.4.0 documentation? There are a few issues relating to that.

codydaig:
I added the CNAME back in as well as put the relative links back to work from the root directory. (So the demo at codydaig.github.io/mean is broken but will work on gh-pages for meanjs.org) I went through and tested this thoroughly on my local machine by modifying my local hosts file. 

I also removed the sample page, post, and sass files that Jekyll comes with by default. 

This commit is good to go and ready to be merged.

lirantal:
thanks @codydaig , applied.
@ilanbiala I don't think there's any "proper" 0.4.0 documentation. That branch originally came from Amos a long time ago and made considerable changes there to structure, etc...

PR #651, opened 2015-07-15
do not use aes when creating ssl certs to prevent password requirement

grantgeorge:
As a developer, I should not use SSL passwords for testing the production environment locally.


lirantal:
thanks @grantgeorge !

PR #655, opened 2015-07-16
Use validator.js for email validation

codydaig:
Takes pull request #414 and makes it compatible with 0.4.0. 

@fauria - If you want to create your own pull request to be merged in so the commit shows up as you, your more than welcome too! I just thought this was cool and started playing around with it ad figured I'd submit a pull request for the functionality you suggested while I had it implemented.


fauria:
Thank you, I already submitted it.

PR #656, opened 2015-07-17
Use validator.js for email and password length validation.

fauria:
Rely on https://github.com/chriso/validator.js for validations.

### DIFF
--- a/modules/users/server/models/user.server.model.js
+++ b/modules/users/server/models/user.server.model.js
@@ -5,7 +5,8 @@
  */
 var mongoose = require('mongoose'),
 	Schema = mongoose.Schema,
-	crypto = require('crypto');
+	crypto = require('crypto'),
+	validator = require('validator');
 
 /**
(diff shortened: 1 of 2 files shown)


lirantal:
@fauria thanks looks good.
would be great if you can add tests for the validator here, but for now I'll merge it in.

modules/users/server/models/user.server.model.js
```diff
@@ -18,7 +19,14 @@ var validateLocalStrategyProperty = function(property) {
  * A Validation function for local strategy password
  */
 var validateLocalStrategyPassword = function(password) {
-	return (this.provider !== 'local' || (password && password.length > 6));
+	return (this.provider !== 'local' || validator.isLength(password, 6));
```
rhutchison: @fauria Was there an issue with password.length?

@lirantal

fauria: No, its just that I thought it was good for consistency to use validator where possible.

PR #659, opened 2015-07-17
Fixing up SSL support

lirantal:
1.  On startup fallback to non-SSL mode if no cert/key is present to start in secure mode
2. Consistency for the error messages printed out


config/config.js
```diff
@@ -56,9 +56,9 @@ var validateEnvironmentVariable = function() {
     console.log();
     if (!environmentFiles.length) {
         if (process.env.NODE_ENV) {
-            console.error(chalk.red('No configuration file found for "' + process.env.NODE_ENV + '" environment using development instead'));
+            console.error(chalk.red('+ Error: No configuration file found for "' + process.env.NODE_ENV + '" environment using development instead'));
```
ilanbiala: Why is there a plus sign at the beginning?

lirantal: To distinguish the rest of the startup text information on the screen from actual errors. I couldn't find any other better method for that. I can remove, it's just styling on the console.

jloveland:
LGTM

Should we have an option for the user to have the application exit out if the it could not be started with https? If I deliberately set `config.secure = true`, the thought is that I may never want my application to start using http. We could add a config called `config.https_only = true`. We could default it to false.

I understand if you all think this would be too opinionated or make the config too complicated, I just wanted to see what people think.

lirantal:
@jloveland I think wrapping too much logic around it is an overkill. If this is production or anything of real use case, the application will simply start with no SSL mode and accessing it via https:// will fail so it is straightforward and easy to understand that something went wrong there + there are the logs.

jloveland:
Agreed, thanks!

lirantal:
Thanks for the review guys, merging.

Issue #660, opened 2015-07-18
description & keywords env var not display content in markup #0.4.0


lukekhamilton:
The app.description and app.keywords content from ./config/env/defaults.js is not being rendered in the markup. 

Have tried to workout why this, however I cant seam to work it out.

lirantal:
this seems very relevant and related to https://github.com/meanjs/mean/pull/536 so maybe we should close #536 and instead submit a new PR on the same topic for 0.4.0 branch

codydaig:
@lukekhamilton @lirantal 

I actually just got an email this morning about this and the correct solution from @jdalton. I will submit the pull request shortly.

lukekhamilton:
Confirmed #665 fixes this can be closed once merged. :)

lirantal:
great, thanks @codydaig and @lukekhamilton

PR #665, opened 2015-07-19
change _.extend to _.merge

codydaig:
Fixes issue #660 

The _.merge properly merges two objects together. I'm not sure on the exact differences between _.extend and _.merge but they act differently, and _.merge produces the desired results.


lukekhamilton:
Sweet that fixed it. :)

rhutchison:
merge recursively tries to map properties, where-as extend will copy the object value.

codydaig:
@rhutchison Thanks for the clarification!!

PR #673, opened 2015-07-20
Remove console.log from an article test

codydaig:
There seems to be an extra console.log laying around in one of the Article Tests. Probably got left behind by someone during debugging.


rhutchison:
Thanks

lirantal:
Good catch @codydaig, thanks.

Issue #525, opened 2015-04-20
before and after on user module unit tests


gilesbradshaw:
Wouldn't you be better with beforeEach and afterEach?

ilanbiala:
Which parts/files specifically?

codydaig:
I'm assuming @gilesbradshaw is taking about changing before and after ti beforeEach and afterEach. So instead of creating a two users as the beginning for all tests to share, it would create new users for each test to share. Then after each test it would reset the database before moving onto the next test.

https://github.com/meanjs/mean/blob/0.4.0/modules/users/tests/server/user.server.model.tests.js

If this is something we want, we could discuss it with #684 otherwise, I think we should close this issue.

lirantal:
I don't think it's an all or nothing. Some tests require other tests to run before them and are based on the previous test.

I do however suggest @gilesbradshaw that if you want to add more tests (and there are many places to add them, whether it's the backend, frontend, or e2e) then just go ahead and create a PR.

PR #684, opened 2015-07-23
Enhancing users model tests

lirantal:
Adding couple more tests to confirm users model works as expected.
Also confirms PR https://github.com/meanjs/mean/pull/439 is working as expected.

### DIFF
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -10,7 +10,7 @@ var should = require('should'),
 /**
  * Globals
  */
-var user, user2;
+var user, user2, user3;
 
 /**
  * Unit tests
@@ -33,6 +33,15 @@ describe('User Model Unit Tests:', function() {
 			email: 'test@test.com',
 			username: 'username',
 			password: 'password',
+			provider: 'local'
+		});
+		user3 = new User({
+			firstName: 'Different',
+			lastName: 'User',
+			displayName: 'Full Different Name',
+			email: 'test3@test.com',
+			username: 'different_username',
+			password: 'different_password',
 			provider: 'local'
 		});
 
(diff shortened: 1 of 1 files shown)


modules/users/tests/server/user.server.model.tests.js
```diff
@@ -77,6 +86,37 @@ describe('User Model Unit Tests:', function() {
...
+					});	
+				});
+			});
+		});
+
+		it('should not be able to save different user with the same email address', function(done) {
+			user.remove(function(err) {
+				should.not.exist(err);
+				user.save(function(err) {
+					user3.email = 'test@test.com';
```
mleanos: Should this line be using the email property of user? If for some reason the value of user.email changes, it would be easier to manage by referencing whatever email is being used by the user object.

`user3.email = user.email;`

mleanos:
These tests are looking really good. I left a comment suggesting a minor change.

ilanbiala:
@lirantal can we already make the indentation here 2 spaces with the expectation that we'll fix all other files soon?

codydaig:
LGTM

mleanos:
LGTM as well. Anyone have any thoughts on my line comment?

codydaig:
@mleanos Personally, I write every test as if had to run on its own. I would do a beforeEach and afterEach so I don't rely on other tests to prove what the current test is doing. But for a middle ground, your solution would work nicely just in case another test further up ends up changing the email.

mleanos:
@codydaig I agree with you on the strategy of using beforeEach and afterEach. It would ensure each test will be completely independent. 

I saw there was a discussion on moving this way, in some other issue. Can't seem to find it now. Thanks for providing your perspective. If we are moving to using this strategy, then my solution wouldn't be necessary; even as a temporary middle-ground. 

I'm definitely fine with this PR being merged, in it's current state. @lirantal @ilanbiala

codydaig:
@mleanos #525?

mleanos:
@codydaig Thanks. I actually just saw that you referenced it in this conversation. Too many tabs open, and it was right under my nose.

codydaig:
@mleanos I go through the issues and pull requests when I'm on the train every morning and on the way home in attempt to better understand the structure and concepts (since It's hard to code on a tablet while standing, I read the issues and pull requests instead), so sadly I've got too many of these memorized. :P

modules/users/tests/server/user.server.model.tests.js
lirantal: yeah, good note, I'll update.

lirantal:
fixed, and squashed commits.
thanks for the review guys!

PR #689, opened 2015-07-24
[WIP] User refactor

rhutchison:
Up for discussion.

This is a work in progress.  There is incomplete code and more to be done.  I am looking for general direction feedback, not a code-review.


modules/users/server/models/user.server.model.js
```diff
@@ -3,30 +3,24 @@
...
-	return (this.provider !== 'local' || validator.isLength(password, 6));
-};
-
-/**
- * A Validation function for local strategy email
- */
-var validateLocalStrategyEmail = function(email) {
-	return ((this.provider !== 'local' && !this.updated) || validator.isEmail(email));
+var validateLocalStrategyPassword = function (password) {
+	return ((password && password.length > 6) || !_.isEmpty(this.providers));
```
ilanbiala: Shouldn't this just be validatePassword?

mleanos:
@rhutchison I've taken a good look at the code. It's a pretty clean implementation.

What were you thinking about on how to handle the callback handling? What type of flow were you considering, and what limitations do you see with the current working implementation? For me, the callbacks seem fine, and they shouldn't affect what you're doing here. Although, I'm not sure I like how the callback routes are spread out in the env/config settings.

Were you planning on storing the 'local' provider in the providers field of the user model, as well? That seems like the right place for it.

Overall, this is really good work. The use of the EmailSchema is nice and clean.

lirantal:
So we now have competing PRs: #689 (this one) vs #559, and not only that but also we have the Admin Mode PR by @rhutchison to merge in which I wonder if when we merge either of user refactoring PRs will they just mess up the Admin Mode PR.

trainerbill:
@lirantal Luckily the admin module really only touches roles, username, firstName, and lastName.  I don't think this refactor touches any of those.  I will try and test it out but am just about ready to leave on a 4 day vacation.  Looking forward to see 0.4.0 when I return!

Issue #678, opened 2015-07-22
Error with sslcerts/key.pem running production 


jkelleher:
Running with 0.4.0 branch all is well for grunt (dev). When I set NODE_ENV to production and run I get:
```
/Users/john/Documents/dev/dmd/node_modules/mongoose/node_modules/mongodb/lib/mongodb/connection/base.js:246
        throw message;
              ^
Error: ENOENT, no such file or directory './config/sslcerts/key.pem'
```

Any ideas?

jloveland:
See "Running with TLS (SSL)" in README. You need to generate your keys. 

sh ./scripts/generate-ssl-certs.sh

lirantal:
@jkelleher 
1. can you update to the latest 0.4.0 version please? you are missing one of the features we added to detect this problem and let you know about it
2. secondly you should run dev and prod environments like this: `grunt dev` and `grunt prod`

@jloveland you've made a good point there without noticing :-) we can correct the behavior today to *tell the user* to do exactly that - to run the scripts/generate-ssl-certs.sh and not just quit abruptly and leave the user hanging. I'll definitely push that little feature.

jkelleher:
Thanks @jloveland (duh) - worked fine - @lirantal also thanks - suggest docs mention need to create the **sslcerts** folder for shell script to work.
`grunt prod` now works with `MEAN.JS application started on port 8443` but I see nothing when I go to `https://localhost:8443' (ditto when I use http). I'll keep digging.

lirantal:
yep, go ahead and update us on how this goes.

jkelleher:
@lirantal not much progress. All seems well with `grunt prod` but there's no reply from https://localhost:8443. Console shows:
```
Uncaught Error: [$injector:modulerr] http://errors.angularjs.org/1.2.28/$injector/modulerr?p0=mean&p1=Error%3A%2…https%3A%2F%2Flocalhost%3A8443%2Flib%2Fangular%2Fangular.min.js%3A18%3A170)
```
I suspect it has to do with minification of the angular.module() call but struggling to unravel what exactly is broke.

jloveland:
the script to generate the certs [creates](https://github.com/meanjs/mean/blob/0.4.0/scripts/generate-ssl-certs.sh#L12) the ./config/sslcerts directory. However, for people that want to generate their certs without the script, we can have that instruction in the docs.

@jkelleher I was not able to reproduce the error. maybe someone else can try?

jkelleher:
[...]

PR #691, opened 2015-07-24
Enhancing SSL error message

lirantal:
# Before 
When running grunt prod with `secured` mode enabled for SSL support, and there were missing key/cert files then the server would fall back to HTTP-only and would notify of the missing SSL files.

# After
Enhancing error message with an actual instruction on how to fix the missing ssl problem when running in prod mode and secured enabled

Addresses concern reported in #678 

image


codydaig:
Tested and LGTM.

PR #698, opened 2015-07-24
Putting the head in order

simison:
- Moving a few meta tags+base tag before the title. I've noticed Bootstrap [recommends this](http://getbootstrap.com/getting-started/#template) (wasn't sure why) so I did some quick googling and [IE seems to be the reason](http://blogs.msdn.com/b/ieinternals/archive/2011/07/18/optimal-html-head-ordering-to-avoid-parser-restarts-redownloads-and-improve-performance.aspx) but there were some other speculations, too. Feel free to search more.

- ...in any case, [charset tag should come before the title](http://www.w3.org/wiki/The_HTML_head_element#Stop_right_there.21_Inline_CSS_and_JavaScript_is_not_too_clever.21).

- Removing keyword tag since it [isn't really used anymore](https://chrisedwards.me/seo/keyword-meta-tag-google/).

- Removing duplicate Content-type/Encoding tag

- Adding respond.min.js so that media queries would work with old IEs.

- Using a proper CDN for shim JS. That previous one was over http (problems when using https) and googlecode.com doesn't send proper cache headers; it's not smart to use it as a CDN.

### DIFF
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -1,24 +1,17 @@
 <!DOCTYPE html>
 <html lang="en">
-
 <head>
-	<title>{{title}}</title>
-
-	<!-- General META -->
 	<meta charset="utf-8">
-	<meta http-equiv="Content-type" content="text/html;charset=UTF-8">
 	<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
-	<meta name="fragment" content="!">
-	<base href="/">
-
-	<!-- Responsive META -->
 	<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1">
+	<base href="/">
+	<title>{{title}}</title>
+	<meta name="description" content="{{description}}">
+	<meta name="fragment" content="!">
+
+	<!-- Apple META -->
 	<meta name="apple-mobile-web-app-capable" content="yes">
 	<meta name="apple-mobile-web-app-status-bar-style" content="black">
-
-	<!-- Semantic META -->
-	<meta name="keywords" content="{{keywords}}">
-	<meta name="description" content="{{description}}">
 
 	<!-- Facebook META -->
 	<meta property="fb:app_id" content="{{facebookAppId}}">
@@ -38,13 +31,8 @@
 	<!-- Fav Icon -->
 	<link href="/modules/core/img/brand/favicon.ico" rel="shortcut icon" type="image/x-icon">
 
-	<!--Application CSS Files-->
+	<!-- Application CSS Files -->
 	{% for cssFile in cssFiles %}<link rel="stylesheet" href="{{cssFile}}">{% endfor %}
-
-	<!-- HTML5 Shim -->
-	<!--[if lt IE 9]>
-		<script src="http://html5shim.googlecode.com/svn/trunk/html5.js"></script>
-	<![endif]-->
 </head>
 
 <body class="ng-cloak">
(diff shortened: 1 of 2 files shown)


simison:
PS, you'll probably want to see the diff in [split mode](https://github.com/meanjs/mean/commit/d764faebcd9d0aace4f3445c328df46494b4f205?diff=split)...

modules/core/server/views/layout.server.view.html
```diff
@@ -38,12 +31,13 @@
...
-	<!--Application CSS Files-->
+	<!-- Application CSS Files -->
 	{% for cssFile in cssFiles %}<link rel="stylesheet" href="{{cssFile}}">{% endfor %}
 
-	<!-- HTML5 Shim -->
+	<!-- HTML5 shim and Respond.js for IE8 support of HTML5 elements and media queries -->
 	<!--[if lt IE 9]>
-		<script src="http://html5shim.googlecode.com/svn/trunk/html5.js"></script>
+		<script src="//oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script>
+		<script src="//oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
```
rhutchison: If 0.4.0 is only supporting IE10+ (https://github.com/meanjs/mean/pull/587#issuecomment-123313817), is this needed?

lirantal:
LGTM. Thanks for cleaning this up.

modules/core/server/views/layout.server.view.html
lirantal: I don't mind leaving this there, we can clean it up in a later version when we'll also probably upgrade AngularJS versions, etc. IIRC IE9 is deprecating on January 2016 or something like that.

lirantal: if we're going all the way with removing html5mode support as @rhutchison mentions then I'm ok with cleaning this up as well.
Question is whether removing this will completely make the framework unusable on IE9 or not?

simison: That's the stuff currently used Bootstrap recommends to have — I'd say remove them when BS4 comes out?

rhutchison: I'd assume bootstrap recommends that only if you are trying to support those browsers.

simison: Yup, removed them now.

https://github.com/simison/mean/commit/7f486cfd8956f4c0b7e9552f70eb300f4c5eeadf

lirantal:
Great, thanks. Squash and I'll merge it in.

simison:
Hmm, not getting this stuff very fresh out of my head at 2am... :-) After Squashing should I make a new PL or?

(Going to be offline pretty much until next week some day after this comment, leaving traveling)

lirantal:
@simison you can rebase your commits on this branch and when you push it use --force.
This should probably work for you:
```
#because you have 4 commits then
git rebase -i HEAD~4
#in the editor leave the top one commit as pick, then change the prefix of the 3 commits to s which is shorthand for squash
#exit the editor
#push your changes 
git push origin patch-11 --force
```

rhutchison:
LGTM - thanks @simison

simison:
@lirantal thanks, so I just forgot to --force earlier. Looks better?

lirantal:
@simison I merged though in the future you should squash all commits to one to clean it up

simison:
:+1: Will practise git more. ;-)

Issue #694, opened 2015-07-24
Minor issue: Unused local variable passport in express.js branch 0.4.0


vaucouleur:
Hi,

A minor issue: Unused local variable passport in express.js branch 0.4.0, probably left over after a recent refactoring.

https://github.com/meanjs/mean/blob/00a4c0603aa6a2c785361a37cff6e5ae153b5688/config/lib/express.js#L18

lirantal:
Thanks @vaucouleur , I have a PR for that.

lirantal:
closing this due to PR https://github.com/meanjs/mean/pull/699 to fix it

PR #699, opened 2015-07-24
Removing unused passport variable in express.js library file

lirantal:
Addresses issue reported in https://github.com/meanjs/mean/issues/694


rhutchison:
LGTM

Issue #482, opened 2015-03-19
First login socket messages not going through


karan-kapoor90:
I copied the socket implementation from the 0.4 branch to 0.3 and it works fine. However instead of calling the `this.Socket=io()` in the chatController like it is, in the 0.4 app, I call it in the AuthenticationController, as follows 

    app.controller('AuthenticationController', ['$scope', '$http', '$location', '$state', 'Authentication','Socket','socketFactory',
    function($scope, $http, $location, $state, Authentication,Socket,socketFactory) {

        $scope.authentication = Authentication;
        if (Authentication.user) 
            this.Socket = io();
            //this.Socket = io('http://vesmed-karan-kapoor-2.c9.io/');
            //this.Socket = socketFactory({ioSocket: io.connect('http://vesmed-karan-kapoor-2.c9.io/')});
        $scope.userType = ['doctor', 'patient','sponsor'];
        Socket.on('refreshSocket',function(data){
            $scope.authentication.user.socketid = data.myId;
            Authentication.user.socketid = data.myId;
            console.log('####### Refresh from auth cont');
        });

so it's called right as soon as the user logs in. On the server side, in the socket.io.js

    io.on('connection', function(socket) {
        //to resolve the location of chat.server.socket.config.js along with other possible files
        console.log('Connection event triggered on server');
        io.to(socket.id).emit('refreshSocket',{myId : socket.id});
        socket.broadcast.emit('isAlive',{user : socket.request.user._id, socketid : socket.id});
        config.files.server.sockets.forEach(function(socketConfiguration) {
            require (path.resolve(socketConfiguration))(io, socket);
        });
    });

always prints 

    Connection event triggered on server

The server is set to listens to 'chatMessage' events as 

    socket.on('chatMessage', function(message) {
        console.log('%%%%%%%%%%\n chat event emitted \n %%%%%%%%%%%%%%%%');
        message.source = socket.request.user._id;
        User.findOne({_id:message.target._id},'socketid', function(err,user){
            if (err){
                message.type = 'status';
                message.text = 'Oops! Something went wrong. Please try again after a while.';
                message.username = '';
                io.to(socket.id).emit('chatMessage', message);
                //do nothing, pass message to be inserted into users list of messages
                return err;
            } else if (user.socketid){
                //checking if the target user has an active connection
                var socketList = io.sockets.server.eio.clients;
                if (socketList[user.socketid] === undefined){
                    //socketid did not get removed correctly, the user is not online
[...]

PR #701, opened 2015-07-24
Socket IO client enhancement with connect() method

mleanos:
Updated the Socket client service, with a connect() method. Moved state
redirect out of from Socket service.  Added the Authentication.user
check to the Chat client controller, and added a check to make sure the
Socket client service is connected to the server; if not, then connect
using the new connect() method.

I'm just putting this up here, since I think the Socket client service could be refactored. There is a loose relation to #686, and perhaps #482.

### DIFF
--- a/modules/chat/client/controllers/chat.client.controller.js
+++ b/modules/chat/client/controllers/chat.client.controller.js
@@ -6,6 +6,14 @@ angular.module('chat').controller('ChatController', ['$scope', 'Socket',
     	// Create a messages array
         $scope.messages = [];
         
+        // If user is not signed in then redirect back home
+        if (!Authentication.user) $location.path('/');
+        
+        // Make sure the Socket is connected
+        if (!Socket.socket) {
+            Socket.connect();
+        }
+
         // Add an event listener to the 'chatMessage' event
         Socket.on('chatMessage', function(message) {
             $scope.messages.unshift(message);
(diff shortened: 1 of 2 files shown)


modules/chat/client/controllers/chat.client.controller.js
```diff
@@ -1,11 +1,21 @@
 'use strict';
 
 // Create the 'chat' controller
-angular.module('chat').controller('ChatController', ['$scope', 'Socket',
-    function($scope, Socket) {
+angular.module('chat').controller('ChatController', ['$scope', '$location', 'Authentication', 'Socket',
```
lirantal: I assume @rhutchison will tell you something about $location :p

mleanos: Yea... Once we decide to move completely away from $location, we can update this to use $state.go(). I just wasn't sure if this would be necessary to do atm; for the purposes of this fix.

lirantal: @rhutchison mentioned we probably want to refactor that in a larger scale so definitely after 0.4.0 but would be good to have couple more eyes on this PR before we merge.

mleanos:
@lirantal I had to do a hard reset of this branch from 0.4.0. There were some merge conflicts that caused this branches history to look pretty bad. I think it's straightened out now. 

Can you confirm?

lirantal:
@mleanos yeah looks good in terms of commits squashing, good job :)
From your side, are we ok to merge this in?

mleanos:
Phew. Twas a couple of stressful hours with this one. Not as simple as the gitingore PR :)

From my end, everything is good. I addressed @simison comments. And I've tested this thoroughly.

lirantal:
Cool, merging in, thanks for squashing :)

PR #704, opened 2015-07-25
Gulp task enhancements - template cache

rhutchison:
These are updates to gulp build automation.

### Gulp Template Cache

### Auto Prefixer


rhutchison:
@ilanbiala please review

mleanos:
LGTM.

.jshintrc
```diff
@@ -9,7 +9,7 @@
 	"curly": false, // Require {} for every new block or scope.
 	"eqeqeq": true, // Require triple equals i.e. `===`.
 	"immed": true, // Require immediate invocations to be wrapped in parens e.g. `( function(){}() );`
-	"latedef": true, // Prohibit variable use before definition.
+	"latedef": "nofunc", // Prohibit variable use before definition.
```
ilanbiala: What does "nofunc" mean?

rhutchison: to ignore function declarations - https://github.com/jshint/jshint/issues/941

Taken from [angular-styleguide](https://github.com/johnpapa/angular-styleguide#style-y230)

ilanbiala:
@rhutchison squash?

rhutchison:
Wanted to give you time to review.  You want it down to 1 commit?

ilanbiala:
Yes please.

rhutchison:
@ilanbiala ready for final review

ilanbiala:
@rhutchison I like the idea and the implementation seems fine, but I'm trying to figure out how much the performance actually increases. Is there any way for you to maybe test a couple pages and see if it is indeed faster and by how much?

rhutchison:
it will avoid the html download on every page load, so the rendering is much quicker.  The performance will be based on the client internet connection/speed.

ilanbiala:
Ok, until we remove Grunt, will this cause issues in the JS?

ilanbiala:
@rhutchison since this only adds features and doesn't break anything, I'm moving this to 0.4.x and if it's ready for 0.4.0 in the next few days I'll move it back and merge, but there really isn't much rush to shove it out.

rhutchison:
@lirantal @ilanbiala I tested 'grunt prod' with these changes.  I didn't notice any issues.

grunt prod

codydaig:
LGTM

lirantal:
@rhutchison yeah of course it will be pretty much ok since most of the change is on gulp.
On that - what does the template cache do and how does it work? I'm curious

Also, why do we need `"latedef": "nofunc"` ?

rhutchison:
@lirantal https://github.com/meanjs/mean/pull/704#discussion_r35483344

[template cache](https://docs.angularjs.org/api/ng/service/$templateCache) will look for all of the html files in the application and then compile them in to javascript statements and minify with the app, so all of the html is downloaded with the app.  If you look at network traffic when you are using the app, whenever you change state, you will see the html is fetched, which will slow down the initial state load.  What angular does is when the html is fetched from the server, it is put in template cache.  This does that step at build time (on production).

lirantal:
so basically template cache means all the HTML templates for Angular actually downloaded when the app first loads so that it doesn't need to go and download them later on?
I'm not against but it sounds to me like simply a trafe-off between when to download the templates, at first app load or later on one by one, and when the templates grow then this will mean more time to load the app on first load.

rhutchison:
The concept follows the same idea of combining all of your application scripts and css.  It might take slightly longer to download the initial app, but here are a couple things to consider when not using template cache:
* Initial app load is going to pull down angular app, then start app, figure out what html, go back to server for html templates, then render. This may be slower than downloading the entire app with cached templates.
* If you have many templates accessed (for the first time) in a single layout, you can have blocking requests.

lirantal:
cool, thanks for the elaborate explanation @rhutchison - I think you can basically copy&paste parts of this thread to put it on the documentation for 0.4.0 :)

rhutchison:
Reopening against master

PR #143, opened 2014-08-17
Debug mongoose queries and Add angular-translate

rubenjgarcia:
Option to debug mongoose queries on development environment

PR #707, opened 2015-07-25
Enabling mongoose debug mode

lirantal:
Enhancing support for development environment to enable mongoose debug mode so that we can debug db queries

PR submitted out of original idea from https://github.com/meanjs/mean/pull/143

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -6,7 +6,9 @@ module.exports = {
 		options: {
 			user: '',
 			pass: ''
-		}
+		},
+		// Enable mongoose debug mode
+		debug: process.env.MONGODB_DEBUG || false
 	},
 	log: {
 		// Can specify one of 'combined', 'common', 'dev', 'short', 'tiny'
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -8,7 +8,9 @@ module.exports = {
 		options: {
 			user: '',
 			pass: ''
-		}
+		},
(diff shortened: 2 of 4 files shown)


mleanos:
+1 LGTM. I'm excited to try this out.

ilanbiala:
@lirantal doesn't it print a lot more stuff? Shouldn't we check against an environment variable to see because I wouldn't want debug to always run..

lirantal:
@ilanbiala you're right I don't think it should be enabled by default, I'll update the PR.

trainerbill:
I honestly don't mind debug mode running in dev mode...  But an env variable would be nice for production boxes.  so you don't have to change code to go into debug mode.  So I would add it to production.js with

production.js
process.env.MEAN_MONGOOSE_DEBUG || false

development.js
process.env.MEAN_MONGOOSE_DEBUG || true

lirantal:
I actually agree with @ilanbiala that user wouldn't want mongodb debug on, even on development environment, and I do like the environment variable idea so I'm combining it to this
```
```

ilanbiala:
What about secure?

lirantal:
@ilanbiala you mean the environment config 'secure'? we don't have that on 0.4.0

ilanbiala:
Oh yeah right, then this is fine.

lirantal:
Thanks guys.

PR #708, opened 2015-07-25
Heroku Deployment Batch

almegdad:
Add Heroku Deployment Batch Script File

### DIFF
--- a/herokuDeployment.bat
+++ b/herokuDeployment.bat
@@ -1,0 +1,13 @@
+@echo off
+
+rem usage:1) just execute the batch by double click
+rem usage:2) cmd>herokuDeployment.bat "your commit message here"
+rem note: add >null to the end of command lines to hide output messages (ex: git status >null)
+
+git status 
+git add . 
+git status
+set message="%~1 committed at %date%_%time%"
+git commit -m %message% 
+git push heroku master
+pause 


lirantal:
Thansk @almegdad for helping out.
My overall comments:
1. Would be good to also update the README with a section on Heroku (and also work with @codydaig about updating the 0.4.0 documentation about it, probably another PR is required for that)
2. I'm not a fan of windows/batch files so personally I think converting it to a shell script is a better choice and also more consistent because we have shell scripts on the projects and not batch scripts
3. If we do decide on adding this script I'll put it in scripts/ folder and not on the package root level

codydaig:
@almegdad @lirantal - When I've used Heroku in the past, I always have two different instances on heroku. Staging for testing and Production for actual production. This script assumes your always pushing to one machine. I'd recommend going with a grunt (or gulp) task. 
https://www.npmjs.com/package/grunt-heroku-deploy
(I haven't personally used this one, but I think it would be more efficient and along the lines of rest of the project to do deployment as a grunt/gulp task than a separate script.

trainerbill:
Yeah Grunt or Gulp definitely

lirantal:
@codydaig yeah I'm ok with that. @almegdad want to give it a try?

almegdad:
Thanks for your quick response , 
maybe script needs to support more features , 
I,ll work on it as a gulp/grunt/batch .

codydaig:
Close this PR?

@almegdad If you make it work with grunt, open a new PR.

Issue #221, opened 2014-10-19
Suggestion: Removing unused CSS


shubhamVerma:
Hey guys, I was checking out the sample app on Heroku. I ran an audit with the Chrome dev tools. 

![image](https://cloud.githubusercontent.com/assets/2177282/4693951/7cac5030-57b4-11e4-9fd2-e09028d80f89.png)

I was wondering if we can add the [grunt-uncss](https://github.com/addyosmani/grunt-uncss) plugin for removing unused css.

Here's the link to the article - http://addyosmani.com/blog/removing-unused-css/

Cheers,
Shubham

ilanbiala:
@shubhamVerma the 0.4.0 branch has Gulp. The only issue with using {{build tool here}}-uncss is that some classes may be applied through directive JS, so how can the tool really know. And some might not even be in code, like Angular-bootstrap requires their JS and bootstrap's CSS. How will it not remove bootstrap's CSS?

codydaig:
@shubhamVerma Any update on this?

lirantal:
well there's still unused CSS, but that's obvious, because we are including for example bootstrap.css and the MEAN.JS project isn't about implementing every piece of that.

So that's not necessarily a defect, but rather we should look into using grunt-uncss, possibly.

lirantal:
So @codydaig are we going for grunt-uncss or puritycss?

codydaig:
I'm unfamiliar with grunt-uncss. Personally, I'd pick purifycss. I submitted a new PR against master #819. It's open for discussion.

lirantal:
Quite old of a thread, closing it and anyone is welcome to try and introduce postcss + uncss for this feature.

PR #709, opened 2015-07-25
purifycss

codydaig:
Removes excess CSS from the production css. 

Project can be seen here:
https://github.com/purifycss/purifycss

They also have a gulp plugin, just wanted to check too see what people thought before adding gulp as well. 

Brought up a while ago in #221 but I came across this project today and fell in love because it checks the HTML and javascript files. :-D


lirantal:
I wonder though if there's any risk of false positives and this would result in bad layout due to the relevant CSS being removed?

codydaig:
There's always the chance for that. From what I've heard from people using this in production, it does a great job. However, since there is always the risk I made it so they can still easily use the minified version without any CSS removed here: 'public/dist/application.all.min.css' .

simison:
Is there a way to get this warn about these instead of just bluntly removing stuff? Would be better practise imho.

codydaig:
@simison I made it so running grunt prod does what it always does, but if you want to have purified css, then run grunt prod:pure and it will purify the css. Does that address your concerns?

lirantal:
I prefer having this on 0.4.x so we can test/discuss it further. @ilanbiala ?

ilanbiala:
Yeah 0.4.x please so we can have more time to review.

codydaig:
@lirantal @ilanbiala Sounds good to me!

lirantal:
@codydaig maybe re-submit for master branch now?
also, is there a reason for choosing purifycss over grunt-uncss?

codydaig:
@lirantal I'll have to redo that pull request against master since there's conflicts and I was going to wait until we decided whether or not we actually wanted this feature. 

I picked purifycss because that is what I've used, I'm not opposed to grunt-uncss.

codydaig:
Moving conversation to #221

PR #713, opened 2015-07-26
BUG: fix non-admin user edit route.  Broke with admin feature

rhutchison:


mleanos:
LGTM. I can confirm this fixes the issue.

mleanos:
The issue still exists, when trying to update the user's profile picture. @rhutchison and I are looking into the issue right now. But the problem seems to be with the `app.param('userId');` existing in both the admin and user routes. However, none of the user routes are using the `userId` param. 

Does anyone know what is passing that into the routes? My initial thought was that the instance of the User resource was passing the `userId` along with the request. I'll try to confirm whether that's the case or not.

mleanos:
Confirmed, or at least I believe I have, that `app.param('userId')` isn't being used by any of the user routes.. a little confusing as to why it's there.

rhutchison:
This PR addresses issues with non-admin routes getting caught up in the admin routes due to /api/users/picture matching /api/users/:userId.  This has been addressed by validating the value of userId against ObjectId type.

The first round of tests are started, but we should consider adding more.

@mleanos is completing a final round of testing and will provide the final sign-off.

mleanos:
This seems to be resolved with these fixes. I've tested this thoroughly, with different variations of use cases. There is a bigger issue at play here. Because I'm not sure if these fixes are necessary, if our routes are defined more accurately; but they are definitely necessary in order to get this fixed asap.

Thanks @rhutchison for getting these fixes out.

mleanos:
@rhutchison And I had a discussion about this, and we had different opinions on the matter. 

I'm a little worried about these fixes disrupting other dev user's experience with this project. Some users might have custom user routes that depend on the `app.param('userId');` being defined in the user routes configuration. If they're keeping their app synced with the 0.4.0, they will run into issues. 

My suggestion is to change the admin api to use `api/admin/users`. I'm not sure if my concerns are valid, but just wanted to put it out there.

mleanos:
Another thought.. can this issue be completely avoided by implementing ACL on the users routes?

rhutchison:
I am absolutely opposed to /api/admin/users.  I can not think of a reason why an authenticated user, editing its own record would depend on app.param('userId').  If someone is relying on it in their project, it's likely because they have built an admin feature or a public profile page, at which point the current users.userById exposes too much information.  Introducing a temporary solution like /api/admin/users will further kick the can down the road.  At some point you would say we can never make the change, because this route is also now embedded in some developers project.  At some point, we need to do what is right for the project.

GET /api/users/me follows RESTful practice, but PUT /api/users does not.  I have been set on changing these for a while now.
 * GET /api/account
 * PUT /api/account
 * PUT /api/account/password
 * POST /api/account/picture
 * DELETE /api/account/providers/:providerId

The following is absolutely valid and should have been there from the beginning.  What other ways are there to validate the value of userId? (I have little express/node experience)

```javascript
	if (!mongoose.Types.ObjectId.isValid(id)) {
		return next('route');
	}
```

The only alternative I can think of is to define your routes in the order that you want them to execute.  Should we have a single (server-side) route definition per module?  I can still see this conflicting across modules.

I am open to discussion, and I have not included the route changes to this PR because I do not want to hold back the fix, and neither should this discussion.

mleanos:
I agree that the user routes should be refactored. I'm not debating that.

What I'm trying to get across is that this particular fix introduces an unnecessary change to the users routes configuration. I don't think this is the best approach, in order to fix a bug that a new feature has introduced. In my experience, it's usually best to modify the incoming feature, rather than one that already existed. 

Making potentially intrusive changes like this, might not be good at this point in the 0.4.0 branch; right before it's about to be made a release branch. Once 0.4.0 is out, then I think we should discuss changes to the user routes, and sort this out at that time.

I don't want to hold this PR up, but I wanted to express my concerns & get feedback from other members of this community.

rhutchison:
> What I'm trying to get across is that this particular fix introduces an unnecessary change to the users routes configuration. I don't think this is the best approach, in order to fix a bug that a new feature has introduced.

In my opinion, the bug was always there.  It was just never exposed because the parameter was never used.

If userId param is not a valid ObjectId, the db query will throw an invalid cast exception.  This will result in a 500 error.  You can argue that this is your intention, but I think it's a weak argument.  The other option is to see if there are any other route definitions that match your request.  If there are no other route definitions, then the user will receive a 404.

A 404 will tell the client that the given route does not exist - a 500, that there is a server error (try again later).

I am open to hearing from others in the community about the proposed solution.  There may be better ways to define the routes without changing the route prefixes.

codydaig:
In every API that I have worked with, every "table" in the database has it's own API route. So, the way I've seen it done is the users would have /api/users/:userID and if you wanted to change the profile picture of the userID, you sent a PUT request to /api/users/userIdHere with the updated User object. The server would then take care of the authentication too ensure the user actually had permission to perform that PUT operation, and if it didn't it would return 401 (Unauthorized). 

That's the way I've seen every RESTful API work, and that's the way I've understood the RESTful concept to be structured. 

* I used "Table" above in quotes because that could be a table, collection, column family, or whatever the database model your working with titles it. Also, Table does not include join tables in a relational database model.

mleanos:
@codydaig That's what I was driving at, with my comment about avoiding these issues by implementing ACL on the users routes. That way we can send each user (admin, user, etc) to the same routes, and the ACL would handle the authentication. Since a user list wouldn't need to be accessed by non-Admin user's then we can just lock that one down to only allow the Admin role.

I think we can clean-up the user routes, but I guess my concern was that we don't introduce any changes at this moment. And my suggested fix, to change the Admin routes to `/api/admin/users` was intended to be temporary until we sort out these issues with the routes; and to expedite a fix to this bug.

mleanos:
Although I do agree with @rhutchison that it would be weird for the requesting user to send their own `:userId`, if they are authenticated. We are sending the user object with each request.

mleanos:
@rhutchison Would you be willing to compromise, and submit a PR that fixes this by changing the admin routes to `/api/admin/users` as a temp fix? This issue is only relevant to the Users Admin module. Since the Articles routes are normalized. 

We need to make sure this fix goes out asap. The 0.4.0 branch, in it's current state, won't allow a user to update their profile. 

Afterward, we can continue this discussion. And will have bought ourselves some time, to sort all this out.

rhutchison:
@mleanos I am not willing to submit a PR for a temporary solution and something I feel strongly against.

mleanos:
@lirantal @ilanbiala  Any thoughts? 

We definitely need more eyes on this.

lirantal:
Maybe I'm missing the problem. Can someone explain it in simple terms?

* I don't see 2 declarations of the userById parameter
* The user API routes *anyway* didn't have the userById parameter there, they are just /api/users/picture style routes so if there was an app.param() there then it was redundant there.

mleanos:
@lirantal With the addition of the Admin module, new `/api/users` routes were introduced; defined in the admin.server.routes config. The problem is occurring since there's now a conflict between these routes. If a user tries to update their profile, it is trying hit the admin route, that has ACL enforced on it. Thus, if the user doesn't have the 'admin' role, they can't update their own profile.

Furthermore, the `/api/users/picture` route is conflicting with the new admin route `/api/users/:userId`.

mleanos:
@lirantal I guess my issue with this PR, is that in order to fix a bug that a new feature introduced, a pre-existing feature is being fixed with what I feel is a band-aid. Using the `return next('route');` is probably not the best way to handle an invalid `:userId`. I would think we'd want to handle those, in a very specific way. I prefer what's being done in the Articles.articleByID method. 

I think the `/api/users` routes can be better defined, and shared with the Admin features. That's what the ACL is for. However, in order to get this fix out asap, with zero impact on other modules, My suggestion is to change the Admin Users routes to `/api/admin/users` until we sort out the users routes. Perhaps, putting it in the 0.5.0 pipeline? 

Introducing potentially breaking changes to existing modules, could disrupt user's of this project; especially in this stage of the 0.4.0.

rhutchison:
@mleanos what is the point of putting something in that you are going to take away in 0.5.0 as you are suggesting?  I feel like we are beating a dead horse here.  If there is enough concern that a invalid objectid should throw a 401 exception like articles instead of trying to match another route, then lets agree to remove the admin feature until we can fix the user routes.

mleanos:
@rhutchison The point is that the Admin feature doesn't impact many user's experience. The potential downside of going back and modifying the Admin feature, in the future, is very minimal. 

I don't understand your concern, with "taking" something away. We wouldn't be taking anything away. We would be improving the User's routes, and then hooking the Admin feature up to those new routes. 

I have the Admin feature working in my project. I merely had to make 3 changes.. to use `/api/admin/users`

`users.client.service` (which probably could have been put into it's own service file anyways.)
`admin.server.policies`
`admin.server.routes`

If in the future we move to change the Admin Users routes to using the same routes, that the Users use, it won't impact anything. Developers, using this project, can keep the `/api/admin/users` implementation, or they could move to using the same routes as the existing Users module. I don't see an issue with that strategy. 

What I do see as an issue, is that current users of this project, can (and have) been impacted by the new Admin feature. Removing the Admin feature at this point, would be a bad decision, as there's no need.

mleanos:
We probably are beating a dead-horse here. Something that we can all take from this, is that we should make sure we thoroughly test our PR's (and others'). And to not make changes without considering how they will affect other features; which test coverage should help with. I've been guilty of this myself, and I gave my blessing on the Admin PR before it was merged.

rhutchison:
@mleanos I've pushed some changes this evening.  The major difference is https://github.com/meanjs/mean/pull/713/files#diff-2bfba1dc361c727303e80bc84ccf08c7R11.  Please review, test, and provide feedback.  Additionally, I set the invalid objectid parameters to return 400's.

mleanos:
@rhutchison This looks really good. I'll do some testing tomorrow afternoon, and report back. Thanks for handling this. I don't see any impact on the existing Users routes; do you know of any?

@codydaig How does this look to you? Anyone else?

modules/users/server/routes/admin.server.routes.js
```diff
@@ -7,15 +7,18 @@ var adminPolicy = require('../policies/admin.server.policies'),
 	admin = require('../controllers/admin.server.controller');
 
 module.exports = function (app) {
+	// User route registration first
+	require('./users.server.routes.js')(app);
```
mleanos: This seems like the glue, that holds this solution together. If that's the case, I would comment that a bit more; possibly explaining why this is being done.

mleanos:
@rhutchison I've tested this thoroughly, and can confirm the issues are resolved. I very much prefer this solution over the previous one. The `/api/users` routes present a unique scenario, so we shouldn't run into the same sort of issues with other modules, since they will be normalized.

One question, is why the admin policy is named with `.policies`? Shouldn't it be `.policy` to be consistent? I didn't notice that before, from your original Admin PR. 

Other than that, I think you did a great job. Thank you.

Definitely, LGTM :)

lirantal:
That latest change indeed looks cleaner @rhutchison.

Please also fix the policies name to policy naming convention as the articles entry has, and if @mleanos tested and confirms then we're good to go here.

rhutchison:
@mleanos @lirantal @ilanbiala Give it a final glance

modules/users/server/routes/admin.server.routes.js
```diff
@@ -7,15 +7,18 @@ var adminPolicy = require('../policies/admin.server.policies'),
 	admin = require('../controllers/admin.server.controller');
```
mleanos: The admin.server.policy is still referencing `.policies`.

mleanos:
@rhutchison The adminPolicy require in the admin server routes, is still referencing `.policies`

mleanos:
LGTM. I haven't tested this latest version. I might have some time in a couple hours; but since the only thing that seems to have changed was the admin policy file being renamed, and the reference to it in the `require`, I don't see any issues.

ilanbiala:
@lirantal LGTM, if you don't get a chance to merge by tomorrow I'll take care of it unless you leave a comment saying otherwise.

lirantal:
Thanks guys for all the hard work, merging.
@rhutchison I especially appreciate the contribution of tests along with this PR, thanks for that and let's try to make it a habit :)

atomfried:
Hey there,
I just started using 0.4.2 and am a little confused about this.
Shouldn't 
* /user/:userId be allowed for admins generally and for users iff they own the id
* /user/me forward be an alias for /user/idOfRequestingUser
?

M

lirantal:
@atomfried ca you open a new issue to focus on the problem? please provide background and more details on current vs expected behavior if you think it's a bug.

atomfried:
@lirantal not a bug, I was referring to the design discussion. I use the design stated above because I have multiple /user/:id/[...] routes which I want to be usable by both users (id=me) and admins (any id).

lirantal:
can you make it more clear what is the current situation vs desired?

atomfried:
Currently users have:
* GET /users/me
* PUT /users/  (is this REST?)
* POST /users/picture
* etc.

Admins have
* GET /users
* GET/PUT/DELETE /users/:id

Now I want to *users* to have *things*. With the current design it would probably look like this:

* /things  (for users: list their things, for admins: list all things)
* /users/:id/things (only admins: list one user's things)

That's not bad I guess, but you have to implement the admin user distinction for each new type of thing. Intuitively I would prefer something like this:

* POST /users/ (create account, no auth required)
* GET /users (list users, only for admin)
* /users/me (just an alias for /users/:id, using the authenticated user id)
* /users/:id (allowed for admins or users with that id)
* /users/:id/things/ (allowed for admins or users with that id)
* /things (all things, only admins)

That way the user/admin distinction has to be implemented only once in the user middleware.

lirantal:
@atomfried why not have routes as `/things/:userId` for a user's own things?

atomfried:
Because the way I understand REST, `/things` is a route for things, not for users. So `/things/:id` should refer to the thing with the given id.

codydaig:
@atomfried Your correct there. What I would go is take in a userID query parameter to the get request to /things. It shouldn't be tied to the /users route.

atomfried:
That makes sense, thank you.

PR #720, opened 2015-07-28
Article middleware was calling getErrorMessage with a null argument

vaucouleur:
The article middleware was calling getErrorMessage with a null argument, causing a crash when this method tried to access 'code' on an null parameter.

The bug was not exposed by the original test, since it was mixing two (related) aspects:

* An invalid Id (a badly formed mongodb identifier)
* An non-existent Id (an identifier with no corresponding document in the database)

Modifications:

- Fixed the message property in the article controller (the error message follows the wording of the error message in "users.password.server.controller.js", in case of username not found)
- Added a new test to check modifications and avoid regressions


ilanbiala:
@lirantal LGTM.

lirantal:
Thanks guys.

Issue #722, opened 2015-07-28
angularFileUpload resolving to 1.1.8, when installing bower dependencies


mleanos:
I wiped out public/lib, and re-installed the Bower dependencies. angular-file-upload has resolved to 1.1.8; even though bower.json has `"angular-file-upload": "~1.1.5"`

In angularFileUpload `1.1.8`, the asset files have moved from `public/lib/angular-file-upload/` to `public/lib/angular-file-upload/dist`; and there's only a `.min.js` file. The project assets config is looking for angularFileUpload in the former location. Thus, I'm receiving an injection error. 

I'm able to resolve it by changing the assets config of default (& production) to the new location; `public/lib/angular-file-upload/dist/angular-file-upload.min.js`

Can anyone else confirm this is an issue?

trainerbill:
@mleanos it should match to version 1.1.8 as there is a ~ in the front with will match any patch version.  I think the issue is with angular file uploads semver.  They pushed a major/minor release to a patch version....

Answer to your question is that yes I am seeing the same issue and this needs to be fixed.  I did a bower update and got 1.1.8 which breaks MEAN.  Looks to me like there is a 1.2 release.  If we are going to work on anything maybe we should go to that?

https://github.com/nervgh/angular-file-upload/releases

mleanos:
@trainerbill Thank you for looking into this. 

I'm not sure that we need to decide on which version to go with at this exact moment. As each version after 1.1.5 has the `dist` folder. So the same fix can be applied. 

I'll submit a PR, for the above fixes. Afterward, I'll look into which version might be best, and go through some tests.

mleanos:
Created PR #723

mleanos:
PR #724 is the accurate fix. Leaving this issue open, for discussion about which version to land on.

simison:
I'd just bluntly test if the latest 5.x works with mean and upgrade bower.json accordingly.

mleanos:
Submitted a new PR for this. #729

PR #723, opened 2015-07-28
Updated location of asset file for angularFileUpload

mleanos:
This addresses #722. I've updated the default & production assets config to use angualr-file-upload's `public/lib/angular-file-upload/dist/angular-file-upload.min.js`

Currently, after installing the Bower dependencies, the patch version of angular-file-upload is breaking the MEAN app.


mleanos:
This PR is a duplicate of #724, which seems to be more accurate; as @simison is using the `src/module.js` asset for the default config.

mleanos:
Closing since #724 is the right fix.

PR #724, opened 2015-07-28
Fix invalid ng-file-upload paths

simison:
Did a fresh clone and tested 0.4.0 branch; found out there was an invalid path for ng-file-upload module.

~~It's actually very out-dated (1.1.5 and newest is 5.0.9) but I don't have time now to test if that would work.~~

### DIFF
--- a/config/assets/default.js
+++ b/config/assets/default.js
@@ -14,7 +14,7 @@ module.exports = {
 				'public/lib/angular-ui-router/release/angular-ui-router.js',
 				'public/lib/angular-ui-utils/ui-utils.js',
 				'public/lib/angular-bootstrap/ui-bootstrap-tpls.js',
-				'public/lib/angular-file-upload/angular-file-upload.js'
+				'public/lib/angular-file-upload/src/module.js'
 			],
 			tests: ['public/lib/angular-mocks/angular-mocks.js']
 		},
--- a/config/assets/production.js
+++ b/config/assets/production.js
@@ -14,7 +14,7 @@ module.exports = {
 				'public/lib/angular-ui-router/release/angular-ui-router.min.js',
 				'public/lib/angular-ui-utils/ui-utils.min.js',
 				'public/lib/angular-bootstrap/ui-bootstrap-tpls.min.js',
-				'public/lib/angular-file-upload/angular-file-upload.min.js'
+				'public/lib/angular-file-upload/dist/angular-file-upload.min.js'
 			]
 		},
 		css: 'public/dist/application.min.css',


mleanos:
@simison I submitted a PR for this fix #722. Wasn't aware that `src/module.js` was the correct asset for the default config. I, probably, mistakenly used the `.min.js` file in `dist` for default.

simison:
Ah, cool. I'm traveling so following mean partially only.

I'd ditch the whole .min.js set from prod conf but that's another discussion.

mleanos:
@simison Thank you. So using the `src/module.js` is definitely the correct choice?

simison:
For now it seems to be. There's some other bug disturbing profile editing now so I can't really test if the file upload itself works now with it — it throws some mongo error.

mleanos:
Good point. We're working on getting the PR merged for the profile edit bug #713. I can test the actual file upload, because I have the proposed fixes in a local branch.

mleanos:
LGTM. I've confirmed, the profile picture upload works with this fix.

simison:
:+1:

simison:
BTW if you have working setup and time, I'd also look into updating ng-file-upload plugin a few versions up. Would be pretty straightforward to test.

mleanos:
@trainerbill thought we should look into 1.2. I'll look into the more recent versions you mentioned.

mleanos:
@simison I think you were getting the file upload plugins. We're using angular-file-upload https://github.com/nervgh/angular-file-upload/releases

simison:
Oh. Confusing. :-) Thanks.

mleanos:
This fix is still valid.. still LGTM :)

ilanbiala:
@mleanos @trainerbill is 1.2.0 also using the same structure?

trainerbill:
I'm not sure... I just thought since we had to look into it we may as well make it work with the most recent version...  I'll let @mleanos decided whats best.

ilanbiala:
@mleanos let me know if it makes sense to move to 1.2.0.

mleanos:
@ilanbiala Actually, it doesn't. Instead of `src/module.js` in 1.1.8, the file looks to be `src/index.js`.. I'm about to test 1.2 out. I'll report back shortly.

mleanos:
Changing `config/assets/default` to use the `src/index.js` from 1.2 doesn't work; the index.js file doesn't seem to be valid, and there isn't any other file (that I see) that can be used for the default asset. I'm getting the injection error. Looks like they have some issues to work out with that asset file. 

Without really digging deep into why the 1.2 `src/index.js` file doesn't work, I'd recommend that we stick with 1.1.8 for now.

mleanos:
@ilanbiala I have confirmed that 1.2 does work when using the `.min` file for the default asset config; and the upload functionality works as well. I know that @simison didn't seem to think it was a good idea to use the .min file.

trainerbill:
@mleanos Why don't we just wait till 1.2 is functioning properly and add=1.1.5 to bower.json and be done with this in 0.4.0?

mleanos:
@trainerbill Fine by me. I don't see a problem with that. @ilanbiala @simison ?

mleanos:
@trainerbill I actually just set the my bower.json to use `"angular-file-upload": "1.1.5"`. I'm deploying to Heroku, and obviously it wants to install 1.1.8. I think this is fine to do, and wait 1.2  works; just as you suggested.

ilanbiala:
@mleanos I'd actually prefer to keep deps steady right now until we release, so can we just change this PR to change the bower.json to 1.1.5 static?

mleanos:
@ilanbiala That sounds good to me. @simison mentioned he was traveling atm. Should I submit a new PR?

mleanos:
Just added PR #729 to fix this.

PR #725, opened 2015-07-28
Pagination Directive POC

trainerbill:
Easily create pagination + search capabilities with an array, a view, and a directive.  Cuts out all the function calls in the admin users controller.  Can be used with articles module as well if we want pagination and search capability.

### DIFF
--- a/modules/core/client/directives/pagination.client.directive.html
+++ b/modules/core/client/directives/pagination.client.directive.html
@@ -1,0 +1,7 @@
+<div class="list-group">
+    <div data-ng-repeat="item in pagedItems">
+        <ng-include src="meanPaginationUrl"></ng-include>
+    </div>
+</div>
+<input data-ng-show="meanPaginationSearch === 'true'" class="form-control col-md-4" type="text" data-ng-model="search" placeholder="Search" data-ng-change="figureOutItemsToDisplay()" />
+<pagination boundary-links="true" max-size="8" items-per-page="itemsPerPage" total-items="filterLength" data-ng-model="currentPage" data-ng-change="pageChanged()"></pagination>
--- a/modules/core/client/directives/pagination.client.directive.js
+++ b/modules/core/client/directives/pagination.client.directive.js
@@ -1,0 +1,44 @@
+'use strict';
+
+//Pagination Directive
+angular.module('core').directive('meanPagination', function () {
+    return {
+        restrict: 'AEC',
+        scope: {
+            meanPaginationData: '=meanPaginationData',
+            meanPaginationItemCount: '=meanPaginationItemCount',
+            meanPaginationUrl: '@meanPaginationUrl',
+            meanPaginationSearch: '@meanPaginationSearch'
+        },
+        controller: ['$scope', '$filter', function($scope, $filter) {
+
+            $scope.buildPager = function () {
+                $scope.pagedItems = [];
+                $scope.itemsPerPage = $scope.meanPaginationItemCount;
+                $scope.currentPage = 1;
+                $scope.figureOutItemsToDisplay();
+            };
+
+            $scope.figureOutItemsToDisplay = function () {
+                $scope.filteredItems = $filter('filter')($scope.meanPaginationData, {
+                    $: $scope.search
+                });
+                $scope.filterLength = $scope.filteredItems.length;
+                var begin = (($scope.currentPage - 1) * $scope.itemsPerPage);
+                var end = begin + $scope.itemsPerPage;
+                $scope.pagedItems = $scope.filteredItems.slice(begin, end);
+            };
+
+            $scope.pageChanged = function () {
+                $scope.figureOutItemsToDisplay();
+            };
+
+            $scope.$watch('meanPaginationData', function (newValue, oldValue) {
+                if (newValue !== oldValue) {
+                    $scope.buildPager();
+                }
+            });
+        }],
+        templateUrl: '/modules/core/directives/pagination.client.directive.html'
+    };
+});
--- a/modules/users/client/views/admin/user-list.client.view.html
+++ b/modules/users/client/views/admin/user-list.client.view.html
@@ -4,17 +4,8 @@
 			<div class="col-md-4">
 				<h1>Users</h1>
 			</div>
-			<div class="col-md-4" style="margin-top: 2em">
-				<input class="form-control col-md-4" type="text" data-ng-model="search" placeholder="Search" ng-change="figureOutItemsToDisplay()" />
-			</div>
 		</div>
 	</div>
-	<div class="list-group">
-		<a data-ng-repeat="user in pagedItems" data-ui-sref="admin.user({userId: user._id})" class="list-group-item">
-			<h4 class="list-group-item-heading" data-ng-bind="user.username"></h4>
-			<p class="list-group-item-text" data-ng-bind="article.content"></p>
-		</a>
-	</div>
 
-	<pagination boundary-links="true" max-size="8" items-per-page="itemsPerPage" total-items="filterLength" ng-model="currentPage" ng-change="pageChanged()"></pagination>
+	<mean-pagination mean-pagination-data="users" mean-pagination-item-count="15" mean-pagination-url="/modules/users/views/admin/user.client.pagination.html" mean-pagination-search="true"></mean-pagination>
 </section>
--- a/modules/users/client/controllers/admin/user-list.client.controller.js
+++ b/modules/users/client/controllers/admin/user-list.client.controller.js
@@ -2,30 +2,11 @@
 
 angular.module('users.admin').controller('UserListController', ['$scope', '$filter', 'Admin',
 	function ($scope, $filter, Admin) {
+
+		$scope.users = [];
+
 		Admin.query(function (data) {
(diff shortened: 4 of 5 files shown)


modules/core/client/directives/pagination.client.directive.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+//Pagination Directive
+angular.module('core').directive('meanPagination', function () {
+    return {
+        restrict: 'AEC',
+        scope: {
+            meanPaginationData: '=meanPaginationData',
```
ilanbiala: These can just be '=' and '@'.

modules/core/client/directives/pagination.client.directive.js
```diff
@@ -0,0 +1,44 @@
...
+            meanPaginationItemCount: '=meanPaginationItemCount',
+            meanPaginationUrl: '@meanPaginationUrl',
+            meanPaginationSearch: '@meanPaginationSearch'
+        },
+        controller: ['$scope', '$filter', function($scope, $filter) {
+
+            $scope.buildPager = function () {
+                $scope.pagedItems = [];
+                $scope.itemsPerPage = $scope.meanPaginationItemCount;
```
ilanbiala: Why assign this again if we can just rename the variable we pass in?

modules/core/client/directives/pagination.client.directive.js
```diff
@@ -0,0 +1,44 @@
...
+                $scope.currentPage = 1;
```
ilanbiala: Doesn't 0-index make the code easier? We would just do `{{currentPage + 1}}`, which I think more people are used to.

modules/core/client/directives/pagination.client.directive.js
```diff
@@ -0,0 +1,44 @@
...
+                $scope.figureOutItemsToDisplay();
+            };
+
+            $scope.figureOutItemsToDisplay = function () {
+                $scope.filteredItems = $filter('filter')($scope.meanPaginationData, {
+                    $: $scope.search
+                });
+                $scope.filterLength = $scope.filteredItems.length;
```
ilanbiala: Where is this used?

modules/core/client/directives/pagination.client.directive.js
```diff
@@ -0,0 +1,44 @@
...
+                var end = begin + $scope.itemsPerPage;
+                $scope.pagedItems = $scope.filteredItems.slice(begin, end);
+            };
+
+            $scope.pageChanged = function () {
+                $scope.figureOutItemsToDisplay();
+            };
+
+            $scope.$watch('meanPaginationData', function (newValue, oldValue) {
+                if (newValue !== oldValue) {
```
ilanbiala: Is this for when you have an object/array that is the same content but from a different source?

modules/users/client/views/admin/user-list.client.view.html
```diff
@@ -4,17 +4,8 @@
...
 	</div>
-	<div class="list-group">
-		<a data-ng-repeat="user in pagedItems" data-ui-sref="admin.user({userId: user._id})" class="list-group-item">
-			<h4 class="list-group-item-heading" data-ng-bind="user.username"></h4>
-			<p class="list-group-item-text" data-ng-bind="article.content"></p>
-		</a>
-	</div>
 
-	<pagination boundary-links="true" max-size="8" items-per-page="itemsPerPage" total-items="filterLength" ng-model="currentPage" ng-change="pageChanged()"></pagination>
+	<mean-pagination mean-pagination-data="users" mean-pagination-item-count="15" mean-pagination-url="/modules/users/views/admin/user.client.pagination.html" mean-pagination-search="true"></mean-pagination>
```
ilanbiala: I don't think the mean prefix is necessary since we aren't a library, we're actually planning on people using this, so maybe something simpler like `<paginated-list data="users" items-per-page="15"></paginated-list>`. I also think the template URL and search should be those values above by default and we can always use the values a user passes in to override, but the simpler the better.

modules/core/client/directives/pagination.client.directive.html
```diff
@@ -0,0 +1,7 @@
+<div class="list-group">
+    <div data-ng-repeat="item in pagedItems">
```
ilanbiala: Is there a simple way we can do a virtual repeat to improve performance?

trainerbill: Not sure what you mean?  ngRepeat is all I have ever used...

modules/core/client/directives/pagination.client.directive.js
trainerbill: meanPaginationUrl has to be an @ as it needs to be interpreted as a string.  Same with Search since that is what I am testing ng-show with. I could possibly change the Search to an =.

modules/core/client/directives/pagination.client.directive.js
trainerbill: I wrote this for another application a long time ago and just ported it here :)  I could probably just reference meanPaginationItemCount in the view

modules/core/client/directives/pagination.client.directive.js
trainerbill: For some reason I think that angular boostrap ui needs it to be current page 1 to start.  Otherwise I am guessing the first page may be called 0

modules/core/client/directives/pagination.client.directive.js
trainerbill: modules/core/client/directives/pagination.client.directive.html

it is required for the angular bootstrap ui.  If you enter a search criteria it has to recalculate how many total items there are based on the filter.

modules/core/client/directives/pagination.client.directive.js
trainerbill: I always do this on my watches for performance.  If the same api call is made that returned the same data set then this would not fire off again.

modules/users/client/views/admin/user-list.client.view.html
trainerbill: It is kind of a MEAN thing tho.  basically just reworked the angular boostrap ui pagination to include a search and view.  I'm not going to fight about the name tho :)  Whatever makes sense to you guys

modules/core/client/directives/pagination.client.directive.html
ilanbiala: http://www.williambrownstreet.net/blog/2013/07/angularjs-my-solution-to-the-ng-repeat-performance-problem/
http://www.williambrownstreet.net/blog/2013/07/angularjs-my-solution-to-the-ng-repeat-performance-problem/

modules/core/client/directives/pagination.client.directive.js
ilanbiala: No I'm saying you don't need the variable after the sign.

modules/core/client/directives/pagination.client.directive.js
ilanbiala: let's do that since it will save on scope watchers.

modules/core/client/directives/pagination.client.directive.js
ilanbiala: Where are we using UI Bootstrap here? I thought this was a custom directive?

modules/core/client/directives/pagination.client.directive.js
ilanbiala: Isn't this a custom directive? What code is calling UI Bootstrap?

modules/core/client/directives/pagination.client.directive.html
mleanos: +1 on using `track by` 

@ilanbiala I think this is a good idea... using the `limitTo`, in conjunction with the filters and `track by` will enhance the performance. User lists tend to get unwieldy. 

@trainerbill You can use the pagination buttons to mimic a 'scroll' by updating the filter range & offsets based on how the user is interacting with the pagination.

modules/core/client/directives/pagination.client.directive.html
```diff
@@ -0,0 +1,7 @@
...
+        <ng-include src="meanPaginationUrl"></ng-include>
+    </div>
+</div>
+<input data-ng-show="meanPaginationSearch === 'true'" class="form-control col-md-4" type="text" data-ng-model="search" placeholder="Search" data-ng-change="figureOutItemsToDisplay()" />
+<pagination boundary-links="true" max-size="8" items-per-page="itemsPerPage" total-items="filterLength" data-ng-model="currentPage" data-ng-change="pageChanged()"></pagination>
```
trainerbill: @ilanbiala This is where UI Bootstrap is called.  They already have a pagination directive.  This directive just adds the search and filter capabilities to it in an "extended" directive.

modules/core/client/directives/pagination.client.directive.html
trainerbill: I don't think I can have a track by with this directive unless we make an option/attribute to pass in what to track by.  Since this can be used to paginate/search any array we really can't guarantee that a object._id is going to be available.  What would you suggest tracking by? $index?

trainerbill:
closing in favor of https://github.com/meanjs/mean/pull/795

PR #727, opened 2015-07-28
Unauthorized route

trainerbill:
Attempting to fix the unauthorized client side issues. Added HTTP interceptors that broadcast events to the child scopes. Added a modal to display on unauthorized api calls. Added authorized route and view. To view changes:

Login with a non admin and try and get to the admin routes /admin/users. Should redirect to unauthorized

Add an article with one user, then login with another non admin user, view the article, add /edit at the end of the url and try and edit the article. Should produce a modal that says forbidden.

PR also comes with an angular bootstrap modal in core that can be reused

### DIFF
--- a/modules/core/client/app/init.js
+++ b/modules/core/client/app/init.js
@@ -4,13 +4,64 @@
 angular.module(ApplicationConfiguration.applicationModuleName, ApplicationConfiguration.applicationModuleVendorDependencies);
 
 // Setting HTML5 Location Mode
-angular.module(ApplicationConfiguration.applicationModuleName).config(['$locationProvider',
-	function($locationProvider) {
-		$locationProvider.html5Mode(true).hashPrefix('!');
-	}
-]);
+angular.module(ApplicationConfiguration.applicationModuleName).config(['$locationProvider', function($locationProvider) {
+        $locationProvider.html5Mode(true).hashPrefix('!');
+}]).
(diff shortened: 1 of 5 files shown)


ilanbiala:
@trainerbill I'm not sure I really like the idea of the modals...but some sort of notification would be good. I think the better question is how unauthorized API calls are even being made. I'd rather prevent that if possible.

trainerbill:
@ilanbiala Aww man I love modals!

There has been a lot of chatter on this subject in the chat so I just made a PR to get this out there.  I like this setup.  If the module is built correctly you would never see a modal as you wouldn't be making unauthorized calls.  However during development it may happen and I like to know whats up without looking at the call.  I can remove the $scope.$on that fires the modal but I think its a pretty cool feature.  You can even do a $scope.$broadcast from modules now to display the unauthorized modal if you want to.

The issue with unauthorized api calls is that right now the articles module needs to split the controllers out by state.  Which will happen in 0.5.0.  If we had a separate controller for editArticle then you could compare logged in user to article user and go to the unauthorized page.  The way article is currently setup does not allow that and so you can just add /edit to the end of view article to get to the edit page.  The server api call is locked down tho so they cant do anything.

Regardless we need to agree on something for this PR so that at least the unauthorized route/view and the interceptors go in 0.4.0

trainerbill:
closing.  Reintroduced to master branch https://github.com/meanjs/mean/pull/796

PR #584, opened 2015-06-02
Added uploads folder under module/users/client/img/profile so that us…

ycmjason:
…ers can upload photos properly.
Couldn't find a way to just add the folder so.. uhh...

PR #731, opened 2015-07-29
Add Grunt task to make sure uploads directory exists

simison:
Using Grunt's internal [mkdir()](http://gruntjs.com/api/grunt.file#grunt.file.mkdir).

It creates the folder only if it doesn't exist already.

Alternative solution for #584

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -6,7 +6,8 @@
 var _ = require('lodash'),
 	defaultAssets = require('./config/assets/default'),
 	testAssets = require('./config/assets/test'),
-	fs = require('fs');
+	fs = require('fs'),
+	path = require('path');
 
 module.exports = function (grunt) {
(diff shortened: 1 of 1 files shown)


ilanbiala:
@lirantal LGTY?

lirantal:
We just need one more thing before we merge - please add that uploads directory `/modules/users/client/img/profile/uploads` to .gitignore so that users don't accidentally push binary data to repositories/PRs.

Other than that looks good, thanks @simison

simison:
@lirantal it's there already: https://github.com/meanjs/mean/blob/0.4.0/.gitignore#L23

I was a bit confused about ./uploads/ directory appearing there as well, but that comes from some multipart middleware.

lirantal:
Great, so we're good here.

PR #735, opened 2015-07-29
Add generator notice

codydaig:
@lirantal @ilanbiala 

This is what I was thinking about doing for the notice about the generator. So it's still on the website, and still listed in the 0.3.x version of the docs, it's just removed from the 0.4.0 docs and and 0.4.0 README until we get it ready to use for 0.4.x.


lirantal:
that looks ok

PR #736, opened 2015-07-30
Refactoring server application service

lirantal:
Refactoring server application service to enable modularity and callbacks, as well as provide app, db, config variables outside of server.js

* Note: this isn't just a general refactoring for the fun of it, it's required to be able to integrate better the tests and the code coverage, but I didn't yet finish working on those so this at least lays the foundation so I can have more time to prepare them in a later PR.

### DIFF
--- a/config/lib/app.js
+++ b/config/lib/app.js
@@ -1,0 +1,52 @@
+'use strict';
+
+/**
+ * Module dependencies.
+ */
+var config = require('../config'),
+	mongoose = require('./mongoose'),
+	express = require('./express'),
+	chalk = require('chalk');
+
+// Initialize Models
+mongoose.loadModels();
+
+module.exports.loadModels = function loadModels() {
+	mongoose.loadModels();
+};
+
+module.exports.init = function init(callback) {
+
+	mongoose.connect(function (db) {
+		// Initialize express
+		var app = express.init(db);
+		if (callback) callback(app, db, config);
+
+	});
+};
+
+module.exports.start = function start(callback) {
+	var _this = this;
+
+	_this.init(function(app, db, config) {
+
+		// Start the app by listening on <port>
+		app.listen(config.port, function() {
+
+			// Logging initialization
+			console.log('--');
+			console.log(chalk.green(config.app.title));
+			console.log(chalk.green('Environment:\t\t\t' + process.env.NODE_ENV));
+			console.log(chalk.green('Port:\t\t\t\t' + config.port));
+			console.log(chalk.green('Database:\t\t\t\t' + config.db.uri));
+			if (process.env.NODE_ENV === 'secure') {
+				console.log(chalk.green('HTTPs:\t\t\t\ton'));
+			}
+			console.log('--');
+
+			if (callback) callback(app, db, config);
+		});
+
+	});
+
+};
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -234,6 +234,20 @@ module.exports = function (grunt) {
 		});
 	});
 
+	grunt.task.registerTask('server', 'Starting the server', function() {
+		// Get the callback
+		var done = this.async();
+
+		var path = require('path');
+		var app = require(path.resolve('./config/lib/app'));
+		var server = app.start(function() {
+			done();
+		});
+
+	});
+
+
+
 	// Lint CSS and JavaScript files.
 	grunt.registerTask('lint', ['sass', 'less', 'jshint', 'csslint']);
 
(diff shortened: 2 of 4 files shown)


codydaig:
@lirantal I love the new refactor. 

Only question, which may be a part of the "I didn't yet finish working on those", how would one execute the code coverage script?

gruntfile.js
```diff
@@ -234,6 +234,20 @@ module.exports = function (grunt) {
 		});
 	});
 
+	grunt.task.registerTask('server', 'Starting the server', function() {
+		// Get the callback
+		var done = this.async();
+
+		var path = require('path');
+		var app = require(path.resolve('./config/lib/app'));
+		var server = app.start(function() {
```
mleanos: beautiful!

config/lib/app.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Module dependencies.
+ */
+var config = require('../config'),
+	mongoose = require('./mongoose'),
+	express = require('./express'),
+	chalk = require('chalk');
+
+// Initialize Models
+mongoose.loadModels();
```
mleanos: What's the logic behind loading the mongoose models here? and then exporting the function below?

mleanos:
Love the new organization.. very clean, and very clear as to what's going on. Taking SOC to new levels. I am still curious about the line comment, I made above.

config/lib/app.js
lirantal: It's required to load the models so that they are already available for the rest of the code/tests when we do `User = mongoose.model('user')`.
I also created an exported function for it in-case some scenario will require just loading the models and not requiring to init() or start() the app.

gruntfile.js
lirantal: thanks :)

lirantal:
@codydaig I'll add the code coverage stuff later and they should be ran through grunt I guess, or we can also create an npm command, doesn't really matter

@mleanos see my reply on your comment. Basically we 'might' need that for the code coverage, I don't know yet because I'm not yet finished getting that running.

@ilanbiala WDYT?

config/lib/app.js
mleanos: Thank you. That makes sense. 

Could one of those scenario to use the exported function, be loading the models for the tests? Forgive me, I haven't spent much time with the test coverage, but I am trying to learn :)

mleanos:
@lirantal I've spent quite a bit of time testing this, with many different use-cases. I haven't found any issues that would stop this from going into 0.4.0. LGTM.

config/lib/app.js
lirantal: Probably. The point is that you can't really "test" or get anything to function without loading the models because they are part of the app. But there could be some cases where you want to load the app but not get it connected to the db or not load express (which is what init does) and cases where you also not want to start the app and make it listen (where is what start does).

Basically it could be argued that the loadModels is redundant because it is already called when the app.js is required. Thing is, I don't know yet in advance but there could be cases where you want to load them again.

lirantal:
Thanks Michael

config/lib/app.js
mleanos: Makes sense. I don't think it's terribly redundant. You definitely want to provide an ability to load them again, for whatever reason. I'm interested in how it could be implemented to *not* load the models when requiring the app. But that's not the scope of what you're doing here; you've made it very flexible. Carry on :)

codydaig:
LGTM

server.js
```diff
@@ -3,32 +3,6 @@
...
-	console.log(chalk.green('Port:\t\t\t\t' + config.port));
-	console.log(chalk.green('Database:\t\t\t\t' + config.db.uri));
-	if (process.env.NODE_ENV === 'secure') {
-		console.log(chalk.green('HTTPs:\t\t\t\ton'));
-	}
-	console.log('--');
-});
+var path = require('path');
+var app = require(path.resolve('./config/lib/app'));
+var server = app.start();
```
ilanbiala: need a newline here

ilanbiala:
Other than the newline, looks good.

ilanbiala:
@lirantal did you have anything else to add?

lirantal:
Not for now at least.
I also updated `server.js` to remove the path and simply load config/lib/app.js since it's from a relative location so this simplifies server.js even more.

server.js
```diff
@@ -3,32 +3,5 @@
...
-	console.log(chalk.green('Environment:\t\t\t' + process.env.NODE_ENV));
-	console.log(chalk.green('Port:\t\t\t\t' + config.port));
-	console.log(chalk.green('Database:\t\t\t\t' + config.db.uri));
-	if (process.env.NODE_ENV === 'secure') {
-		console.log(chalk.green('HTTPs:\t\t\t\ton'));
-	}
-	console.log('--');
-});
+var app = require('./config/lib/app');
+var server = app.start();
```
lirantal: @ilanbiala better?

ilanbiala: Yes

lirantal:
Great, merging.

PR #639, opened 2015-07-10
[WIP] Design/UX Enhancements

rhutchison:
This is a Work In Progress.  This PR is open for discussion with the community.  Please pull this branch and provide feedback.

These are intended to be simple updates to the UX to make it more fresh.

### TODO
 - [x] Header
 - [ ] Homepage
  - [x] Initial page design
  - [ ] Update "MEAN.JS Documentation" well.
 - [x] Sign In
 - [x] Sign Up
 - [ ] Manage Social Accounts in Settings?
 - [ ] @lirantal Merge https://github.com/meanjs/mean/pull/637 & https://github.com/meanjs/mean/pull/638
 - [x] @rhutchison Fix active state hover styles on social buttons.

PR #746, opened 2015-08-02
scroll to top automatically on each page transition

TakayukiSakai:
For most mean applications, scrolling to top on each page transition should look natural.
This change would save time of many developers (like me.)

How do you guys think?


rhutchison:
I don't think this solution would be good enough.  Something like this is a guarantee.  https://github.com/meanjs/mean/pull/639/files#diff-dea85680b1eef2a8f98fdf282b8158e5R9

codydaig:
@rhutchison So is this issue addressed in #639?

PR #747, opened 2015-08-02
DeBug: Remove sensitive data before login In password reset - 0.4.0

almegdad:
Remove sensitive data before login
file: users.password.server.controller.js
function: reset

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ Master Branch:
 [![Dependencies Status](https://david-dm.org/meanjs/mean.svg)](https://david-dm.org/meanjs/mean)
 
 Dev Branch:
-[![Build Status](https://travis-ci.org/meanjs/mean.svg?branch=0.4.0)](https://travis-ci.org/meanjs/mean)
-[![Dependencies Status](https://david-dm.org/meanjs/mean/0.4.0.svg)](https://david-dm.org/meanjs/mean/0.4.0)
+[![Build Status](https://travis-ci.org/meanjs/mean.svg?branch=0.4.1)](https://travis-ci.org/meanjs/mean)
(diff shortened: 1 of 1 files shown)


modules/users/server/controllers/users/users.password.server.controller.js
```diff
@@ -138,6 +138,10 @@ exports.reset = function(req, res, next) {
 									message: errorHandler.getErrorMessage(err)
 								});
 							} else {
+							    // Remove sensitive data before login
+							    user.password = undefined;
```
lirantal: I understand removing salt, but isn't user.password is expected to be there so that passport can verify it? also where exactly is it exposed? it's going straight from code to the passport method, it's not broadcasted back to the user...

almegdad:
* res.json(user); // expose password & salt 
* check: signup & signin in file:users.authentication.server.controller.js
* don't mix between passport authenticate & login functions (authenticate:for authentication,login:for user session)

lirantal:
@almegdad I would move move the change then to line 149 just before the res.json()
Can you also please squash your commits and clear the conflict there

modules/users/server/controllers/users/users.password.server.controller.js
codydaig: @lirantal When a user signs into the website, it also strips the password and salt. 
https://github.com/meanjs/mean/blob/0.4.0/modules/users/server/controllers/users/users.authentication.server.controller.js#L64
I am not sure why it needs to be stripped, but if we want to keep it consistent, then removing both the salt and the password would be appropriate. (If I had to take a guess, I don't think it matters because req.login is a method added to req by passport. And when passport performs a login, it already strips the sensitive info.)

lirantal: Right, although @almegdad did point out that later on they are sent back to the user in res.json() (line 149).

PR #751, opened 2015-08-03
Remove sensitive data before return authenticated user in password reset

almegdad:
Remove sensitive data before return authenticated user in password reset
users.password.server.controller.js
Function: reset
note: ready to merge :+1:   reopened after solved conflict in the previous PR #747

### DIFF
--- a/modules/users/client/views/authentication/signin.client.view.html
+++ b/modules/users/client/views/authentication/signin.client.view.html
@@ -6,7 +6,7 @@
         <div class="form-group">
           <label for="username">Username</label>
           <input type="text" id="username" name="username" class="form-control"
-               data-ng-model="credentials.username" placeholder="Username">
+               data-ng-model="credentials.username" data-ng-change="credentials.username=credentials.username.toLowerCase();" placeholder="Username">
         </div>
         <div class="form-group">
           <label for="password">Password</label>
--- a/modules/users/client/views/authentication/signup.client.view.html
+++ b/modules/users/client/views/authentication/signup.client.view.html
@@ -15,13 +15,12 @@
         </div>
         <div class="form-group">
           <label for="email">Email</label>
-          <input type="email" id="email" name="email" class="form-control" data-ng-model="credentials.email"
-               placeholder="Email">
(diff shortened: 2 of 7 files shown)


mleanos:
I agree with @lirantal that this should be added right before the `res.json`. Other than that, LGTM.

lirantal:
@almegdad please make that change?

almegdad:
there is a tiny difference:
user (with password & salt) will be assigned to req.user by req.login function,
so butting it before req.login will strip out password & salt from req.user too (only in this request , cos other requests passport will strip it)

lirantal:
So you are saying that we want to keep the password&salt fields out for the rest of the req, meaning they won't be available in req.user but I'm not sure that this is desired. Can you justify why?

Otherwise, I'd just prefer at the moment to remove it from the response that is being sent out.

almegdad:
@lirantal why this PR is still pending , please check the last commit "username & email lowercase normalization"

lirantal:
@almegdad couple of stuff:
1. Why do we need to enforce lowercase emails and usernames?
2. Once we're ready you should actually submit another PR for `master` branch with just one commit

@codydaig @mleanos @rhutchison @ilanbiala WDYT?

almegdad:
@lirantal 
to solve the case sensitive issue for username & password:
someone@mail.com & SomeOne@mail.com treated as a different emails 
someone & SOMEONE also treated as a different usernames
commit message:
The email is a string that you will want to normalize to lower case, in
order to avoid one email having more than one account -- e.g.,
otherwise, avenue@q.com can be registered for 2 accounts via
avenue@q.com and AvEnUe@Q.CoM.
refrence:
http://mongoosejs.com/docs/pi.html#schematype_SchemaType-set
* the same thing applies to username

codydaig:
@almegdad @lirantal I agree with the lowercase usernames and emails. However, if we are going to do that, we need to add controller logic to the signin to force the username to lowercase before checking against the server or it won't be able to find the user if they add an uppercase letter anywhere. 

@almegdad This pull request does need to be made against the master branch before it can be merged in.

almegdad:
reopened #837

lirantal:
@almegdad you're totally right, let's take it up in #837 to further discuss the PR.

PR #755, opened 2015-08-04
Bug: reset-password enhancements and bug fix

rhutchison:
Resolved issues:
* models are not loaded.  Regression from #736 (https://github.com/meanjs/mean/pull/736/files#diff-d8679454c2d07935926663de0e9cc9a1L34)
* callback was not passed corresponding user data.
* process exits before mail queue is processed.

Enhancements:
* suppress error messages unless debug mailer option is true.
* from e-mail address comes from config.
* added marker messages to each callback line. (Example: [processed/total] Could not send email for x)
* added status message on-complete to tell how many successful e-mails were sent.

### DIFF
--- a/scripts/reset-password.js
+++ b/scripts/reset-password.js
@@ -1,50 +1,90 @@
 var nodemailer = require('nodemailer'),
-	mongoose = require('mongoose'),
-	config = require('../config/config'),
-	mg = require('../config/lib/mongoose');
+  mongoose = require('mongoose'),
+  chalk = require('chalk'),
+  config = require('../config/config'),
+  mg = require('../config/lib/mongoose');
 
 var transporter = nodemailer.createTransport(config.mailer.options);
 var link = 'reset link here'; // PUT reset link here
+var email = {
+  from: config.mailer.from,
+  subject: 'Security update'
+};
+var text = [
+  'Dear {{name}},',
+  '\n',
+  'We have updated our password storage systems to be more secure and more efficient, please click the link below to reset your password so you can login in the future.',
+  link,
+  '\n',
+  'Thanks,',
+  'The Team'
+].join('\n');
 
-mg.connect(function(db) {
-	var User = mongoose.model('User');
+mg.loadModels();
 
-	User.find().exec(function(err, users) {
-		if (err) {
-			throw err;
-		}
+mg.connect(function (db) {
+  var User = mongoose.model('User');
 
-		var email = {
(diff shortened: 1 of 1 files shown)


mleanos:
LGTM.. tested, and confirmed it's sending the emails. 

Thanks @rhutchison

lirantal:
Cool, thanks guys.

mleanos:
@lirantal @rhutchison Slight issue with this script. If there aren't any users in the database, this will hang with no exit. It may seem trivial, but if this script is automated or added to a batch it would reek havoc.

You can add this simple fix after the err check at the beginning of the User.find() callback
```
if (!users.length) {
      console.log();
      console.log(chalk.yellow('No users were found.'));
      process.exit(0);
    }
```
I have a more elegant solution to the *no users found* issue in the PR i submitted to your fork for the user-refactor email migration.

rhutchison:
@mleanos please re-test

mleanos:
@rhutchison I'm going to be away from my computer for most of tonight. I'll test it tomorrow morning.

scripts/reset-password.js
```diff
@@ -1,50 +1,90 @@
...
+          return reportAndExit(processedCount, errorCount);
+        }
+      };
+    };
+
+    // report the processing results and exit
+    function reportAndExit(processedCount, errorCount) {
+      var successCount = processedCount - errorCount;
+
+      console.log();
```
ilanbiala: Extra?

mleanos:
@rhutchison I've tested this and it works great. However, what's the purpose of the `alerts()`? No alert dialog's popped up for me; is that the intention? If so, I think merely a console log is sufficient.

scripts/reset-password.js
mleanos: @ilanbiala I like the extra line written here. It separates any *line item* messages from the overall processing results.

rhutchison: @ilanbiala provides line separation, as @mleanos mentioned.  It was intentional

rhutchison:
@mleanos alert is just a mapped function to chalk.green, chalk.red, or chalk.yellow depending on the result status.

mleanos:
@rhutchison thank you. I obviously, didn't pay enough attention. Jumped the gun, when I saw the alert :)

LGTM.

lirantal:
Great. Thanks @rhutchison for fixing this.

Issue #761, opened 2015-08-05
Grunt issue


ilanbiala:
Grunt's --force option seems to make a lot of our builds that shouldn't pass pass. For example, this build https://travis-ci.org/meanjs/mean/jobs/74211973 should have failed since 1 test failed, but --force made it continue. Is there a reason we are using --force @lirantal @rhutchison? Have either of you noticed this issue in any of your builds?

lirantal:
@ilanbiala glad you're back :)
I'm submitting a PR for that

ilanbiala:
@lirantal only for a little bit while I'm on vacation so the notifications don't pile up. I ran into it while I'm looking into adding Coveralls to visualize our coverage. @lirantal on that note, did you have coverage set up and did you already start looking into setting up a service such as Coveralls or Codecov?

lirantal:
Yes I already have a pending PR for the coverage, which I was working on for a while and I actually refactored the app.js / server.js especially to make that work :)

I didn't address the coveralls though and some other items so if you want I can submit my PR and you can add your coveralls and other changes to it.

ilanbiala:
@lirantal yeah if you submit your PR, I'll work on adding Coveralls to it, which I started doing in #760 just to make it available to anyone who may have had something.

lirantal:
Cool.

with regards to the job you were referring to (https://travis-ci.org/meanjs/mean/jobs/74211973) and about grunt error handling - so in that case jshint doesn't fail (although if it does it doesn't fail grunt at all), and I do see an error there too which doesn't fail the grunt either.

My PR is a change to grunt's 'force' option to default it to false so it fails on any other task that fails, this way we can be sure that we are catching any kind of issue, whether these are tests, or jshint conventions. Is that what you were referring to?

ilanbiala:
Yes.

codydaig:
@lirantal Fixed in #762?

lirantal:
Yes, thanks @codydaig

Issue #717, opened 2015-07-27
Travis CI Improvements


cdriscol:
I've had a few thoughts/suggestions on improving our current (0.4.0 branch) CI and wanted to get community input:

* **jshint should fail the build**
 * This can help catch some simple mistakes.
* **test errors should fail the build**
 * otherwise, what's the point of tests?
* **Can we run Travis CI on all commits?  Or force PRs?**
 * [This commit](https://github.com/meanjs/mean/commit/9fc88e6e5b5bd05c1e532246356725a3fa16a223) is what introduced some breaking tests, but CI didn't run on it because it wasn't a PR.

Mistakes happen, by everyone, and that's why we have a CI build that runs jshint, tests, etc.. now we just need it to tell us when things aren't okay.

FYI - [Here](https://travis-ci.org/meanjs/mean/jobs/72572487) is an example from Travis CI where **jshint and karma tests failed** but Travis marked this as **Passed**.

codydaig:
@cdriscol Agreed!

ilanbiala:
@cdriscol I'd love to see these changes. Any idea of what we can do?

lirantal:
@cdriscol I've already started working on those changes exactly (apart from your last item on travis-ci configuration), although I was quite busy the last week so if you have anything to submit a PR for go ahead, otherwise I'll just get to it within a couple of days.

cdriscol:
@lirantal - good enough for me!  thanks!

PR #762, opened 2015-08-05
Grunt - fail the build if a task fails

lirantal:
Updating grunt 'force' option to default/force it to false so that we can fail any task that is failing and thus fail the build in case of any issues in tests, jshint conventions, and so on.

This ensures that our CI process is catching any issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,6 @@ env:
   - NODE_ENV=travis
 services:
   - mongodb
+before_install:
+  - gem update --system
+  - gem install sass --version "=3.3.7"
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -207,9 +207,6 @@ module.exports = function (grunt) {
   // Load NPM tasks
   require('load-grunt-tasks')(grunt);
 
-  // Making grunt default to force in order not to break the project.
-  grunt.option('force', true);
-
   // Make sure upload directory exists
   grunt.task.registerTask('mkdir:upload', 'Task that makes sure upload directory exists.', function () {
     // Get the callback


gruntfile.js
```diff
@@ -208,7 +208,7 @@ module.exports = function (grunt) {
   require('load-grunt-tasks')(grunt);
 
   // Making grunt default to force in order not to break the project.
```
ilanbiala: Update the comment too, but can't we just delete it from the file?

lirantal:
addresses issue https://github.com/meanjs/mean/issues/761 and https://github.com/meanjs/mean/issues/717

gruntfile.js
lirantal: about the comment yeah, but we have this:
`grunt.option('force', false);`
if we leave it there in the gruntfile then even if you do `grunt test --force` the force option won't override it and it will force the build to fail in any case. If we do remove this option then grunt indeed defaults to failing tests but you can override it with the --force option from cmdline.

WDYT we should choose?

lirantal:
and yay, we finally have some build fails.

codydaig:
We need to update the travis config then to install SASS properly. That's what this PR is failing on.

lirantal:
Yep, noticed.
I'll update the PR later.

codydaig:
LGTM

gruntfile.js
ilanbiala: I think letting people override from cmd is better because the cmd is already open to run grunt anyway.

.travis.yml
```diff
@@ -6,3 +6,6 @@ env:
   - NODE_ENV=travis
 services:
   - mongodb
+before_install:
+  - gem update --system
+  - gem install sass --version "=3.3.7"
```
ilanbiala: newline?

lirantal: @ilanbiala  fixed

ilanbiala:
LGTM now.

lirantal:
@ilanbiala, Guys

We should really get this one pushed in quick because our CI is simply not reliable anymore.
I found one of my commits showing green on the merge PR window but looking into the travis-ci log (knowing that it's not reliable) I found this:

image

So let's try to push this in ASAP so we can add some trust to the travis-ci safety-net

ilanbiala:
@lirantal I don't see any issues with it, let's merge it in now no?

lirantal:
Cool

PR #766, opened 2015-08-06
Updating JSHINT settings

lirantal:
removing deprecated options in the latest versiosn of jshint which will be replaced by adding JSCS settings.

@rhutchison when you add your JSCS PR this PR will complement it to remove those code style options from JSHINT which are deprecated

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -5,20 +5,13 @@
   "browser": true, // Standard browser globals e.g. `window`, `document`.
   "esnext": true, // Allow ES.next specific features such as `const` and `let`.
   "bitwise": false, // Prohibit bitwise operators (&, |, ^, etc.).
-  "camelcase": false, // Permit only camelcase for `var` and `object indexes`.
   "curly": false, // Require {} for every new block or scope.
   "eqeqeq": true, // Require triple equals i.e. `===`.
-  "immed": true, // Require immediate invocations to be wrapped in parens e.g. `( function(){}() );`
   "latedef": true, // Prohibit variable use before definition.
-  "newcap": true, // Require capitalization of all constructor functions e.g. `new F()`.
   "noarg": true, // Prohibit use of `arguments.caller` and `arguments.callee`.
-  "quotmark": "single", // Define quotes to string values.
-  "regexp": true, // Prohibit `.` and `[^...]` in regular expressions.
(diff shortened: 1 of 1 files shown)


ilanbiala:
@lirantal when were these deprecated. @rhutchison is there a comparison table for JSCS and JSHint? I'm also hoping that whatever new syntax/formatting tools we add will be able to support ES6?

lirantal:
@ilanbiala from jshint.com website  http://jshint.com/docs/options/
image

go ahead and update your local jshint settings to indent 8, or 10, and run grunt and you'll see that it doesn't report anything, because it has no more impact, it's deprecated...

ilanbiala:
Ok, so then there's no need for the comparison table. Still, I want to try to support ES6 already if possible.

lirantal:
what does it mean/entitle to support ES6 from your perspective?

ilanbiala:
Linters/formatters will have an issue with ES6 code because the syntax is different. Ideally, our tools will be able to deal with ES6 and ES5 code, so not only users of MEAN can write ES6 if they want to, but we can start moving to it as it becomes more commonplace.

lirantal:
Is it too soon to talk about ES6 now, though? We have enough PRs on our plate to handle so I say let's get that PR backlog cleaned and can discuss further roadmap.

ilanbiala:
@lirantal I'm not saying we should be actively trying to use ES6, but so we don't have to think about it again in the future, if possible we should spend a couple extra minutes just verifying that there won't be issues if we do start using ES6.

lirantal:
What do you suggest we do? review which settings we need to add to JSHint / JSCS?

ilanbiala:
Just check if ES6 is supported and how well. No need to do so much setup just yet in my opinion.

lirantal:
From a quick research both JSHint and JSCS have support for ES6:
1. http://jscs.info/overview.html - through the esnext operator
2. https://github.com/jshint/jshint/blob/master/examples/.jshintrc - same esnext operator

lirantal:
@ilanbiala any other concerns you'd like to consider for this to be merged?

ilanbiala:
Maybe just add esnext and then you can merge.

lirantal:
Great, added, and squashed commits.

PR #768, opened 2015-08-06
Updating User model tests for synchronous test and fixing done() calls

lirantal:
refactoring the async nature in the user model tests to account for mocha 2 second timeouts causing travis-ci build fails

### DIFF
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -16,7 +16,7 @@ var user, user2, user3;
  * Unit tests
  */
 describe('User Model Unit Tests:', function () {
-  before(function (done) {
+  before(function () {
     user = {
       firstName: 'Full',
       lastName: 'Name',
@@ -45,7 +45,6 @@ describe('User Model Unit Tests:', function () {
       provider: 'local'
     };
 
-    done();
   });
 
   describe('Method Save', function () {
@@ -71,7 +70,7 @@ describe('User Model Unit Tests:', function () {
     it('should fail to save an existing user again', function (done) {
       var _user = new User(user);
       var _user2 = new User(user2);
-
+      
       _user.save(function () {
         _user2.save(function (err) {
           should.exist(err);
@@ -130,7 +129,7 @@ describe('User Model Unit Tests:', function () {
       });
     });
 
-    it('should not be able to save different user with the same email address', function (done) {
+    it('should not be able to save different user with the same email address', function () {
       var _user = new User(user);
       var _user3 = new User(user3);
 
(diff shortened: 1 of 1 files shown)


modules/users/tests/server/user.server.model.tests.js
```diff
@@ -153,6 +150,8 @@ describe('User Model Unit Tests:', function () {
   });
 
   after(function (done) {
-    User.remove().exec(done);
+    User.remove().exec(function() {
```
ilanbiala: I believe you can just do `User.remove().exec(done);`.

modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -102,6 +102,8 @@ describe('User CRUD tests', function () {
   });
 
   afterEach(function (done) {
-    User.remove().exec(done);
+    User.remove().exec(function() {
```
ilanbiala: Same here if passing done still works.

modules/users/tests/server/user.server.model.tests.js
```diff
@@ -71,7 +70,6 @@ describe('User Model Unit Tests:', function () {
     it('should fail to save an existing user again', function (done) {
       var _user = new User(user);
       var _user2 = new User(user2);
-
```
ilanbiala: Don't we want the space here to separate variable declarations for the test from the actual execution?

lirantal: yeah, good catch.

modules/users/tests/server/user.server.model.tests.js
lirantal: you're right but I noticed some odd inconsistency with the after() call failing sometimes and wasn't sure if it is attributed to that or not.
WDYT?

ilanbiala: When doing `User.remove().exec(done);`, the `after()` would fail? That seems weird..I've never run into that issue and it's not like something crazy is happening. Does it work consistently in other callbacks? What about in the `afterEach()` below?

lirantal: not consistently unfortunately :(
I don't mind returning to the original `.exec(done)` and let's keep an eye really close on tests, there is a lot more work required there to get the testing automation safety-net to be really bullet proof.

Ok with you?

ilanbiala: Yeah sounds good.

lirantal: Yeah, if we change the convention to just `.exec(done)` then I'll change it back as it was to be consistent.

lirantal:
@ilanbiala I updated and squashed, this should solve still the async/sync issue with that one test.

codydaig:
LGTM

lirantal:
Thanks, I'll merge.

Issue #770, opened 2015-08-07
User Model tests


rmlopes:
Hi,

I am trying out the 0.4.0 branch. The user model unit test for the email is failing (should not be able to save different user with the same email address).

lirantal:
I noticed issues with that too, actually the test for that specific is a false positive, I've got a PR to send later this evening.

In the meanwhile, can you show me the output of the errors you got?

rmlopes:
Running "server" task

MEAN.JS - Test Environment
Environment:                    test
Port:                           3001
Database:                               mongodb://localhost/mean-test

Running "mochaTest:src" (mochaTest) task


  Article Model Unit Tests:
    Method Save
      √ should be able to save without problems
      √ should be able to show an error when try to save without title

  Article CRUD tests
    √ should be able to save an article if logged in (127ms)
    √ should not be able to save an article if not logged in
    √ should not be able to save an article if no title is provided (55ms)
    √ should be able to update an article if signed in (81ms)
    √ should be able to get a list of articles if not signed in
    √ should be able to get a single article if not signed in
    √ should return proper error for single article with an invalid Id, if not s
igned in
    √ should return proper error for single article which doesnt exist, if not s
igned in
    √ should be able to delete an article if signed in (70ms)
    √ should not be able to delete an article if not signed in

  User Model Unit Tests:
    Method Save
      √ should begin with no users
      √ should be able to save without problems (43ms)
      √ should fail to save an existing user again (45ms)
      √ should be able to show an error when try to save without first name
      √ should confirm that saving user model doesnt change the password
      √ should be able to save 2 different users (51ms)
      1) should not be able to save different user with the same email address

  User CRUD tests
    √ should not be able to retrieve a list of users if not admin (57ms)
    √ should be able to retrieve a list of users if admin (59ms)


  20 passing (3s)
  1 failing

  1) User Model Unit Tests: Method Save should not be able to save different user with the same email address:
     Error: timeout of 2000ms exceeded at null.<anonymous> (C:\Users\rmlopes\meanprojects\meancegi\node_modules\mocha\lib\runnable.js:139:19)
      at Timer.listOnTimeout (timers.js:119:15)


The remaining PhantomJS tests are all marked with SUCCESS

lirantal:
@rmlopes is this the latest version of master branch? can you pull please?

rmlopes:
I just pulled the master in OS X and all the tests pass.

lirantal:
Great, thanks.

[...]

Issue #486, opened 2015-03-21
server side route tests not working in 0.4.0 branch


deepakbhattarai:
the `app`  is not exported in the server.js file. Message
``
timeout of 2000ms exceeded
``
I think the problem is because the variable `server` is not exported in the `server.js` file. So there is no server running in the test.

ilanbiala:
@deepakbhattarai more details please.

deepakbhattarai:
The error is in this code. This hook gives a timeout
``
	before(function(done) {
        var app = express.init(mongoose);
        agent = request.agent(app);
        done();
	});
``

The error
``

  1) Article CRUD tests "before all" hook:
     Error: timeout of 2000ms exceeded
      at done (D:\webdocuments\china-volunteers\node_modules\mocha\lib\runnable.
js:205:67)
      at D:\webdocuments\china-volunteers\node_modules\mocha\lib\runnable.js:226
:9
      at Server.<anonymous> (D:\webdocuments\china-volunteers\modules\articles\test
s\server\article.server.routes.tests.js:29:13)
      at Server.g (events.js:180:16)
      at Server.EventEmitter.emit (events.js:92:17)
      at net.js:1055:10
      at process._tickDomainCallback (node.js:463:13)
``

ilanbiala:
Put in console.log throughout the functions to see where it stops. Then see what's wrong with the isolated statements.

deepakbhattarai:
`before` is never completed. It gives a timeout. It stops in 'done()' line.

ilanbiala:
Where is the last console.log that you see? You have something hanging when you call express.init most likely.

TsarS:
Same issue 
  timeout of 2000ms exceeded   in test

TsarS:
Same issue on fresh install 
 timeout of 2000ms exceeded

netusco:
Same issue on fresh install 
timeout of 2000ms exceeded

lirantal:
can you make sure you've got the latest master branch checked out please?
there is some issue there though, I'll submit a PR to fix it later tonight

netusco:
ok I was a few commits behind and with the latest master branch checked out all tests pass smothly ;)

lirantal:
great.
there is anyway an issue there with that test that I'll commit a fix in a couple of hours

deepakbhattarai:
Great. It is working in the master branch.

PR #771, opened 2015-08-07
Fixing user model test's last test which fail or report a false positive

lirantal:
fixing up a user model test which was not setup correctly without the async done() callback, which led to false postivies. Adding timeouts to the test ensures that the test completes in time, otherwise mocha's 2s timeout will fail the test

### DIFF
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -129,19 +129,24 @@ describe('User Model Unit Tests:', function () {
       });
     });
 
-    it('should not be able to save different user with the same email address', function () {
+    it('should not be able to save another user with the same email address', function (done) {
+      // Test may take some time to complete due to db operations
+      this.timeout(10000);
+
       var _user = new User(user);
       var _user3 = new User(user3);
 
       _user.remove(function (err) {
         should.not.exist(err);
         _user.save(function (err) {
-          var user3_email = _user3.email;
+          should.not.exist(err);
           _user3.email = _user.email;
           _user3.save(function (err) {
             should.exist(err);
-            // Restoring the original email for test3 so it can be used in later tests
-            _user3.email = user3_email;
+            _user.remove(function(err) {
+              should.not.exist(err);
+              done();
+            });
           });
         });
       });


lirantal:
addresses issues with https://github.com/meanjs/mean/issues/770 and https://github.com/meanjs/mean/issues/486

lirantal:
@ilanbiala  @mleanos review please?

modules/users/tests/server/user.server.model.tests.js
```diff
@@ -129,19 +129,24 @@ describe('User Model Unit Tests:', function () {
       });
     });
 
-    it('should not be able to save different user with the same email address', function () {
+    it('should not be able to save another user with the same email address', function (done) {
+      // Test may take some time to complete due to db operations
+      this.timeout(10000);
```
ilanbiala: Quite long no? How long does the test take on average?

modules/users/tests/server/user.server.model.tests.js
```diff
@@ -129,19 +129,24 @@ describe('User Model Unit Tests:', function () {
...
         should.not.exist(err);
         _user.save(function (err) {
-          var user3_email = _user3.email;
+          should.not.exist(err);
           _user3.email = _user.email;
           _user3.save(function (err) {
             should.exist(err);
-            // Restoring the original email for test3 so it can be used in later tests
-            _user3.email = user3_email;
+            _user.remove(function(err) {
```
ilanbiala: Can we not just do `.remove(done);`?

modules/users/tests/server/user.server.model.tests.js
lirantal: They take more than 2 seconds which is mocha default timeout and they usually fail like that on travis-ci. I don't mind having 10 seconds there as a timeout (it's not a delay) just in case because we're not testing mongodb or travis's servers here.

modules/users/tests/server/user.server.model.tests.js
lirantal: that's possible because done does accept an error argument so it would work but it's less verbose and the rest of our tests are more descriptive rather than based on magic behind the scene.

lirantal:
@ilanbiala we're good?

lirantal:
I'd like to push this fix in as well as other tests I have worked on in another PR ASAP so we can get that tests safety next as wide as possible.

Anything else @codydaig @ilanbiala @mleanos ?

ilanbiala:
Yes LGTM.

lirantal:
Great, thanks.

PR #773, opened 2015-08-08
Remove Dev Branch Badges

codydaig:
Since we agreed to make the master branch the current live branch, and the stable releases will be tagged under releases, I removed the development badges from the README.


lirantal:
LGTM

PR #774, opened 2015-08-08
Remove 'data' prefix from attributes

rhutchison:
Remove the 'data' prefix from custom attributes.

### DIFF
--- a/modules/articles/client/views/create-article.client.view.html
+++ b/modules/articles/client/views/create-article.client.view.html
@@ -1,4 +1,4 @@
-<section data-ng-controller="ArticlesController">
+<section ng-controller="ArticlesController">
   <div class="page-header">
     <h1>New Article</h1>
   </div>
@@ -19,8 +19,8 @@
         <div class="form-group">
           <input type="submit" class="btn btn-default">
         </div>
-        <div data-ng-show="error" class="text-danger">
-          <strong data-ng-bind="error"></strong>
+        <div ng-show="error" class="text-danger">
+          <strong ng-bind="error"></strong>
         </div>
       </fieldset>
     </form>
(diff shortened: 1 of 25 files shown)


lirantal:
@ilanbiala per your request...

trainerbill:
Doesn't this go against HTML validation standards?  data- is supposed to be included anytime you have a custom attribute not defined by W3C

codydaig:
@trainerbill The angular style guide that I personally refer back too (and I think it's been mentioned before in a previous issue) never even references using data-. I'm not familiar with how it plans into validation standards though. 
https://github.com/johnpapa/angular-styleguide

lirantal:
`ng-submit` kind of custom attribute is probably not valid by W3C, but the `data-ng-submit` *is* valid. I'm not sure what reasons did @ilanbiala had about removing it

pgrodrigues:
Personally, Im going to keep `data-` prefixes in my apps due to W3C validation tests and also to avoid future browser implementations that might create conflicts with `ng-` attrs which I'm pretty sure there won't be any but, from my point of view it's safer and since it is defined in the [HTML5 spec] (http://www.w3.org/TR/2011/WD-html5-20110525/elements.html#embedding-custom-non-visible-data-with-the-data-attributes) why not keep it? :D

lirantal:
yeah I'm not sure why we removed it, need to wait for Ilan's response here

ilanbiala:
Data- was just to comply with XHTML IIRC. HTML5 allows all kinds of custom attributes if I'm not mistaken and syntax highlighters will work with either one. data- also just adds extra characters and makes everything longer. Anyway Some Angular HTML sometimes doesn't pass through W3C (and we know that isn't what necessarily makes a page better), so I don't see why we should try to conform more towards it.

lirantal:
HTML5 is also strict about the attributes type and it would prefer a data-* over plain custom.
Since AngularJS normalizes those attribute names I don't think it's worth any hassle to remove the data-* version that we have.

Unless there's a strong argument in removing data-* I think we can close this PR.

ilanbiala:
@lirantal I don't really care, but we shouldn't be inconsistent and I see no reason to do `data-` when almost no one does.

lirantal:
I haven't seen a strong argument towards removing `data-*` attributes so I'm not 100% convinced about removing it.

codydaig:
LGTM :-D
:+1:

Issue #777, opened 2015-08-08
Menus service bug in removeSubMenuItem and removeMenuItem


cdriscol:
The following methods in the Menus service are currently buggy..

[_removeMenuItem_](https://github.com/meanjs/mean/blob/master/modules/core/client/services/menus.client.service.js#L146)
[_removeSubMenuItem_](https://github.com/meanjs/mean/blob/master/modules/core/client/services/menus.client.service.js#L163)

__Can reproduce with:__
```javascript
   it('should remove menu item', function() {
      Menus.addMenu('menuId');
      Menus.addMenuItem('menuId', {state:'menu.state'});
      var menu = Menus.removeMenuItem('menuId', 'menu.state');
      expect(menu.items.length).toBe(0);
    });
```

__Cause:__
Both methods attempt to remove a menu item that has a _link property_ (that we no longer use after 0.4.0) that matches the URL passed into the method.

PR #778, opened 2015-08-08
Fixing Menus service removeMenuItem and removeSubMenuItem

cdriscol:
Fixing Menus service removeMenuItem and removeSubMenuItem by.. 

* modified tests to reproduce issues
* replaced references to old link property with new state property

Fixes #777

PR #779, opened 2015-08-09
[bug fix] Fixed Menus remove methods

mleanos:
The Menus client service was attempting to remove Menu items & Sub Menu
items by comparing the `link` field. This field no longer exists; we're
using state. I double checked that there were no existing uses of the Menus `link` field. 

This was pointed out by @cdriscol with #777 

One question/concern.. this could potentially be a breaking change, if users merge this fix, and are using the `link` field somewhere in their custom Menu's. Is this a valid concern? I'm still a little shaky on how we're handling the bug fixes with the current release/branching strategy.

@rhutchison @codydaig @ilanbiala @trainerbill @lirantal

### DIFF
--- a/modules/core/client/services/menus.client.service.js
+++ b/modules/core/client/services/menus.client.service.js
@@ -137,13 +137,13 @@ angular.module('core').service('Menus', [
     };
 
     // Remove existing menu object by menu id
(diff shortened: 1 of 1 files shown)


mleanos:
Closing due to duplicate #778. I also missed a couple referenced to link in the add menu items.

PR #785, opened 2015-08-10
Added database seeding for users in development

trainerbill:
use seedDB in the configs to add users to the database.  It removes the username user and admin and adds them back so the passwords should always be test123


config/env/development.js
```diff
@@ -65,5 +65,6 @@ module.exports = {
       }
     }
   },
-  livereload: true
+  livereload: true,
+  seedDB: true
```
lirantal: let's have it set to false by default, even on dev.

lirantal:
Looks like an Ok solution to seed the admin user, let's make sure the users you created there don't overlap with the users created in the testing environment

config/env/development.js
trainerbill: Are you sure?  Almost every other framework I have worked with auto seeds on development...

trainerbill:
@lirantal I thought about checking to see if the user exists first, but I actually like removing and readding in development.  That way every time you start the app you know you have 2 users with the right passwords to login with.  I guess if it is production we prolly do not want to overwrite...  Let me think on this a bit

trainerbill:
@lirantal I added a check on seed.js so if it is production then admin user will only seed if it doesn't already exist.  Development seeds them everytime by removing and readding...  I like this approach so I always know my passwords in dev should I be testing password setting etc.  I really think we should leave seedDB as true in dev.  Up for discussion @rhutchison @ilanbiala @codydaig @mleanos

config/lib/app.js
```diff
@@ -11,6 +11,9 @@ var config = require('../config'),
 // Initialize Models
 mongoose.loadModels();
 
+// Populate DB with sample data
+if(config.seedDB) { require('./seed'); }
```
ilanbiala: Formatting

config/lib/seed.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+var mongoose = require('mongoose'),
+    User = mongoose.model('User');
```
ilanbiala: Indentation?

config/env/development.js
ilanbiala: It would be nice to allow this to be overriden from the console through an environment variable for quick switching

config/lib/seed.js
```diff
@@ -0,0 +1,82 @@
...
+                } else {
+                    console.log('Local admin added');
+                }
+            });
+        }
+        else {
+            console.log('Admin user exists');
+        }
+    });
+}
```
ilanbiala: Formatting

config/lib/app.js
trainerbill: Ternary or line breaks?

config/lib/seed.js
trainerbill: You are supposed to indent when declaring multiple variables

config/lib/seed.js
trainerbill: Can you be more specific?  passes lint testing.

config/env/development.js
trainerbill: process.env.MONGO_SEED?

config/lib/seed.js
lirantal: looks like it's 4 spaces, instead of 2.

config/lib/seed.js
lirantal: @trainerbill not that. we're referring to 2 spaces indentation.

lirantal:
@trainerbill a few concerns I'd like to raise, see if you can relate to them:
1. in some cases, our 'dev' environment aren't just a matter of a parameter that sets whether the app is running on 'dev' mode or 'production' mode. At times, it's more than that. For example, imagine you are already running your app in production for a while, and you have a big DB, and part of your automation is to export that DB from prod to your staging or dev environment to be able to reproduce bug based on existing data, and such. In that kind of scenarios you might already have users with user name "user" or "admin", which will conflict with your seed code. More than that, the seed is completely removing all users so if you would like to reproduce some bug related to a user that has issues updating his profile fields/pictures/login etc you essentially take this option away by deleting all the users.

2. security concern - having a privileged user named 'admin' is bad practice. even if you 'think' you're on 'dev' environment that's not always the case because: a. users might have their 'dev' environment also exposed on the public Internet; b. users might use the seed action to deploy it on their 'production' environment first and never care to change back the admin username. Actually, more scary is users accidentally running the seed action on production by mistake, or through their CI/CD processes.

Both of these concerns are from my experience working on enterprise large-scale applications.

mleanos:
@trainerbill I share the same concerns that @lirantal mentioned. Whenever you automate something like this, things can go very wrong; with devastating results.

I have a few scripts that I added to `scripts/` used for these exact purposes. Using a script that *I must* run, intentionally, will minimize these accidental environmental mistakes/errors. By doing it this way, I'm in complete control of how/when/what/where any Users that have special privileges get added to my databases. 

To reiterate what @lirantal said: From personal experience, there are some deployments that end up being classified as a *development* environment by an application, but are exposed publicly.

trainerbill:
@lirantal The script does not remove all users, just the admin and user, and readds them only in development.  If production then it searches for an admin user only and adds it only if the admin user does not already exist.

I really don't see this as any sort of security issue.  If you are setting up a production/enterprise application, you better know enough to change the seeded admin password.  Its like root.  Everyone knows its there so you best have a strong password.

So is the end case to your arguments that you want the seedDB set to false in every environment?  Not have a seeding script?

trainerbill:
All.  I updated all the configs to be false by default.  Also added an Environment variable to enable the seeding quickly.  I am not sure on the indentations... I was using webstorm before and it is supposed to automatically pick up .editorconfig.  I switched to atom and downloaded the .editorconfig plugin.  Everything seems to be 2 space indents from what I can tell.

config/lib/app.js
```diff
@@ -11,6 +11,11 @@ var config = require('../config'),
 // Initialize Models
 mongoose.loadModels();
 
+// Populate DB with sample data
+if (config.seedDB) {
+	require('./seed');
```
lirantal: that's still not 2 spaces indentation

lirantal:
@trainerbill you're right, I wasn't paying attention there and it's good we're not adding/removing users sparsely.

I'm still quite paranoid about the default admin user/pass/role there so here's a suggestion: how about we generate the admin user/pass details on the fly, and then print it out to the console/logfile (there's logging enabled by morgan that I added a while back) which makes it much more secure?

trainerbill:
@lirantal Would randomly generating the password for the admin account in production work for you?  I really don't want to randomly generate the username as well.  Could just randomly generate a 20 character password for seeding in production?  But then if they seed once and don't save the password it will not seed again...  They would need to remove the admin user from mongo manually and re seed.

config/lib/app.js
trainerbill: This is super annoying... It is 2 spaces on my editor but committing as 4 spaces.  Not sure what to do:

image

rmlopes: Have u seen [this](http://stackoverflow.com/questions/2316677/can-git-automatically-switch-between-spaces-and-tabs)?

trainerbill:
Alright I have the indentation fixed.  Had to checkout rhutchensens recent commit and just redo my require seed part.  @lirantal Still need the go ahead on what to do with password generation

lirantal:
@trainerbill I can go for just generating the admin password but I would do it for all devs, not just production. Of course whatever we generate on the fly is not going to be there afterwards, that's the point... anyone else is welcome to comment and suggest another method.


BTW I have quickly searched for seeding functional in 3 major nodejs frameworks and couldn't find any mention of that (sailsjs, meanio, and loopback). If you found it elsewhere maybe we can take a look at what they are doing there.

lirantal:
Also now that we have a *proper* jshint working (yay!) we can catch those stuff in the CI process.
Check out the build failures, you're missing a semicolon on https://github.com/meanjs/mean/pull/785/files#diff-e4a0ef5ca61cb235bb9ed42cac9853ffR16

trainerbill:
@lirantal added random password and fixed semicolon.  Let me know if this looks good.  I will rebase once I get the OK that it will be merged ;)

config/lib/seed.js
```diff
@@ -0,0 +1,85 @@
...
+
+var mongoose = require('mongoose'),
+  chalk = require('chalk'),
+  crypto = require('crypto'),
+  User = mongoose.model('User');
+
+console.log(chalk.bold.red('Warning:  Database seeding is turned on'));
+
+//If production only seed admin if it does not exist
+if (process.env.NODE_ENV === 'production') {
```
mleanos: @trainerbill What's the purpose of this block? If the environment is *production* you're creating the Admin account, if it doesn't exist. But below at https://github.com/meanjs/mean/pull/785/files#diff-7dfe67f86c4c0773b98c0422a0cb606aR64, you're removing and re-creating the Admin account anyway. I'm just trying to understand what's going on there; not that it's wrong. Perhaps, add a bit more comments? 

Also, how about breaking this file up into exported funcitons? With an `init` routine, and a couple maintenance/utility functions that will allow you to interact with the seeded users from other area's of the application? For instance, you can have a function that removes/locks one of the accounts.

mleanos:
@trainerbill I'm not against this getting merged in. I would like to understand the benefits a bit more though. Personally, I don't really see myself using this. Could you explain the benefits, over handling it more manually? 

I feel comfortable with the scripts I currently have, since I know exactly when & what databases they are executed against.

mleanos:
@trainerbill Also, how about adding the user info (email, username, displayName, etc) to a `seedDb` config setting? Might be nice to have different settings for the seeded users, based on the environment.

config/lib/seed.js
trainerbill: Basically it will seed the admin account in production only if it does not already exist....  In DEV and TEST it always seeds the admin and user accounts so you always have something to login with.  I am just trying to get something in so first time users don't have to use mongo to insert an admin user.  Once its seeded they can use the admin console to make any updates that they want to make.

trainerbill:
@mleanos The benefit is that so first time users don't have to use mongo to add an admin user to be able to use the admin console.  Also in DEV and TEST the users are always overwritten so you can always login... so if you accidentally changed a password for admin in dev you wouldn't have to worry about it.  Just do MONGO_SEED=true grunt and grab the password from the console.  It is only turned on by changing the config or adding the environment variable so I think we just document that for first time users that they can seed the DB once by doing:

MONGO_SEED=true grunt prod

Then save the admin password that is generated.

mleanos:
@trainerbill Aren't you removing and adding the *Admin* account regardless of the environment with the last code block here https://github.com/meanjs/mean/pull/785/files#diff-7dfe67f86c4c0773b98c0422a0cb606aR64 ? Thus negating the add in the above *IF* block?

How do you feel about adding the user's info as config settings? So we could have different username's & emails used for different environments?

mleanos:
@trainerbill I apologize. I think the indentation was throwing me off at first. Thought that block was outside the IF statement.

lirantal:
I'm generally ok with this unless anyone else has any other comments? I think we've been raising all points already and everything was addressed to a satisfying extent by @trainerbill

codydaig:
LGTM. (Haven't tested it, but I like the solution overall.) As long as it has been tested, squash the commits and it looks ready to me. :-D

package.json
```diff
@@ -23,18 +23,20 @@
     "async": "^1.3.0",
     "body-parser": "^1.13.1",
     "bower": "^1.4.1",
+    "cfenv": "~1.0.0",
     "chalk": "^1.1.0",
```
ilanbiala: Where is this used?

mleanos:
LGTM as well. Any thoughts on adding the user info to the configs?

package.json
```diff
@@ -23,18 +23,20 @@
...
     "compression": "^1.5.0",
     "connect-flash": "~0.1.1",
     "connect-mongo": "~0.8.1",
     "consolidate": "~0.13.1",
     "cookie-parser": "^1.3.2",
+    "crypto": "0.0.3",
     "express": "^4.13.1",
```
ilanbiala: Is node's built-in crypto not good enough/have what we need? For something like this where the password could just be pick 5 letters from the alphabet, I don't really wanna add a library for this.

package.json
ilanbiala: It was just rearranged from the bottom of looks like, no?

mleanos: @ilanbiala I think this was from the Cloud Foundry implementation; probably got pulled in when @trainerbill checkout out the latest merge from @rhutchison . This shouldn't be in the commit once @trainerbill squashed & has rebased from upstream.

mleanos: @ilanbiala I think you're right. Looks like @trainerbill re-arranged to alphabetize the list.

lirantal:
@mleanos what do you mean adding the user info to the configs?

package.json
lirantal: @ilanbiala actually I think we can have other good uses for the crypto library in the user model tests so I'm not against.

mleanos:
@lirantal I mentioned it a couple times in my previous comments...

I was thinking it could be nice to add these seeded users' email, username, etc as environmental configs. This would allow us to create different *seeded* users for each of the environments. For instance, I may want to have a specific email & username used with my development environment. For *production* I could use a different email/username. 

This is just something that crossed my mind, and I wanted to see what others thought of it. Might be overkill though.

package.json
ilanbiala: I'm not against it either, but I don't think Node's crypto module is insufficient. Am I mistaken?

lirantal:
@mleanos ahh ok, gotcha. I think it might be nice but not that important as the seeding process should be very manual anyway. I'd say let's add it as it is and we can enhancement as we go.

codydaig:
@lirantal +1 to adding it in as is and having enhancements follow if desired.

trainerbill:
Wow... how on earth did I get all these other updates on here... ugh...  I am just going to do a new branch and submit a new PR.

trainerbill:
I totally messed up this PR.  Closing in favor of:

https://github.com/meanjs/mean/pull/793

PR #794, opened 2015-08-13
used __dirname instead of process.cwd(), this leads to server errors …

bastianwegge:
…if the executable is not run in the same directory.

Simple thing, tested the new meanjs on our farm and instantly ran into this.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -98,7 +98,7 @@ var initGlobalConfigFolders = function (config, assets) {
   };
 
   // Setting globbed client paths
-  config.folders.client = getGlobbedPaths(path.join(process.cwd(), 'modules/*/client/'), process.cwd().replace(new RegExp(/\\/g), '/'));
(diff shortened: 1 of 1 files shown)


lirantal:
Thanks @bastianwegge 
1. How can we reproduce the errors you faced?
2. We need you to submit a PR for `master` branch

bastianwegge:
Hey lirantal,

the usual way to reproduce this is to start a node instance not by beeing in the folder itself. Everybody can reproduce this behaviour on his own machine, but in my purpose this is to run it on the server.

1. clone mean
2. run `npm install`
3. cd into any other directory (server-specific would be like 1 directories up like `/root/app/`)
4. `node app/server.js` or as I did: `forever start /root/app/server.js`

bastianwegge:
Created the pull for master!
https://github.com/meanjs/mean/pull/797

lirantal:
@bastianwegge why aren't you running it with grunt?

lirantal:
let's move the discussion to #797

PR #796, opened 2015-08-13
Unauthorized client side routing

trainerbill:
Reintroducing to master branch.  closing https://github.com/meanjs/mean/pull/727

@ilanbiala @lirantal @rhutchison @codydaig

### DIFF
--- a/modules/core/client/app/init.js
+++ b/modules/core/client/app/init.js
@@ -4,13 +4,16 @@
 angular.module(ApplicationConfiguration.applicationModuleName, ApplicationConfiguration.applicationModuleVendorDependencies);
 
 // Setting HTML5 Location Mode
-angular.module(ApplicationConfiguration.applicationModuleName).config(['$locationProvider',
-  function ($locationProvider) {
+angular.module(ApplicationConfiguration.applicationModuleName).config(['$locationProvider', '$httpProvider',
+  function ($locationProvider, $httpProvider) {
     $locationProvider.html5Mode(true).hashPrefix('!');
+
+    $httpProvider.interceptors.push('authInterceptor');
   }
 ]);
 
 angular.module(ApplicationConfiguration.applicationModuleName).run(function ($rootScope, $state, Authentication) {
+
   // Check authentication before changing state
   $rootScope.$on('$stateChangeStart', function (event, toState, toParams, fromState, fromParams) {
     if (toState.data && toState.data.roles && toState.data.roles.length > 0) {
@@ -24,11 +27,11 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
 
       if (!allowed) {
         event.preventDefault();
-        $state.go('authentication.signin', {}, {
-          notify: false
-        }).then(function () {
-          $rootScope.$broadcast('$stateChangeSuccess', 'authentication.signin', {}, toState, toParams);
-        });
+        if (Authentication.user !== undefined && typeof Authentication.user === 'object') {
+          $state.go('forbidden');
+        } else {
+          $state.go('authentication.signin');
+        }
       }
     }
   });
--- a/modules/core/client/config/core.client.routes.js
+++ b/modules/core/client/config/core.client.routes.js
@@ -13,16 +13,30 @@ angular.module('core').config(['$stateProvider', '$urlRouterProvider',
 
     // Home state routing
     $stateProvider
-      .state('home', {
-        url: '/',
-        templateUrl: 'modules/core/client/views/home.client.view.html'
-      })
-      .state('not-found', {
-        url: '/not-found',
-        templateUrl: 'modules/core/client/views/404.client.view.html',
-        data: {
-          ignoreState: true
-        }
-      });
+    .state('home', {
+      url: '/',
+      templateUrl: 'modules/core/client/views/home.client.view.html'
+    })
+    .state('not-found', {
+      url: '/not-found',
+      templateUrl: 'modules/core/client/views/404.client.view.html',
+      data: {
+        ignoreState: true
+      }
+    })
+    .state('bad-request', {
+      url: '/bad-request',
+      templateUrl: 'modules/core/client/views/400.client.view.html',
+      data: {
+        ignoreState: true
+      }
+    })
+    .state('forbidden', {
+      url: '/forbidden',
+      templateUrl: 'modules/core/client/views/403.client.view.html',
+      data: {
+        ignoreState: true
+      }
+    });
   }
 ]);
--- a/modules/core/client/services/interceptors/auth.interceptor.client.service.js
+++ b/modules/core/client/services/interceptors/auth.interceptor.client.service.js
@@ -1,0 +1,22 @@
+'use strict';
+
+angular.module('core').factory('authInterceptor', ['$q', '$injector',
+  function ($q, $injector) {
+    return {
+      responseError: function(rejection) {
+        if (!rejection.config.ignoreAuthModule) {
+          switch (rejection.status) {
+            case 401:
+              $injector.get('$state').transitionTo('authentication.signin');
+              break;
+            case 403:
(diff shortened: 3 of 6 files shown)


modules/core/client/config/core.client.routes.js
```diff
@@ -16,6 +16,14 @@ angular.module('core').config(['$stateProvider', '$urlRouterProvider',
...
         url: '/not-found',
         templateUrl: 'modules/core/views/404.client.view.html'
-      });
+      })
+      .state('bad-request', {
+        url: '/bad-request',
+        templateUrl: 'modules/core/views/400.client.view.html'
+      })
+      .state('unauthorized', {
+				url: '/unauthorized',
```
ilanbiala: indentation

modules/core/client/views/modal.client.view.html
```diff
@@ -0,0 +1,18 @@
+<div class="modal-header">
+  <h3 class="modal-title">{{modal.title}}</h3>
+</div>
+<div class="modal-body">
+  {{modal.body}}
+  <div data-ng-if="modal.items">
```
ilanbiala: data-

trainerbill:
@ilanbiala Not that I am looking at this, I don't think we should do interceptors for bad request on the client side....   Server side is throwing them quite a bit, even for things like article is not found.  Guessing we don't want a redirect on these errors...  Or do we?

modules/core/client/views/modal.client.view.html
lirantal: We have another PR for data- handling and I think at the moment we can leave the `data-*` stuff in there as is until we get a better decision on that change.

lirantal:
@trainerbill can you elaborate more on the requirement/purpose of this PR? I'm not really following what's it about.

rhutchison:
> Server side is throwing them quite a bit, even for things like article is not found. Guessing we don't want a redirect on these errors... Or do we?

@trainerbill yes, this is the intention. 404 should go use $state.transitionTo on the client to go to the 404 page.  This will all make better sense when resolves are used properly.

I am not a fan of having the pub/sub implementation here, I don't see the value of doing it this way.  The only thing I can think of is you tried implementing state.go in the http interceptor and ran in to the circular reference issue.  You need to use `$injector` like so, `$injector.get('$state').transitionTo('bad-request');`

trainerbill:
@lirantal HTTP interceptors will do certain actions if the server responds with certain error codes.  So for like an unauthorized 401 response I created a route so that the client will redirect to the route so the user knows they are unauthorized.  This will come in handy when a user views an article then adds /edit to the end of the URL of an article that they don't own.  I suggest pulling the branch and checking it out.  Login with a user account and try to go to /admin/users.  View an article the user doesn't own and add a /edit to the URL and try and edit the article.  You will go to the unauthorized route.

@rhutchison The example you gave me was based on pub/sub.  I actually like the idea.  So then in the module controllers we can broadcast the unauth event and it does the same thing every time for certain events.  $state couldn't be injected into config so that is why I went with pub/sub but after doing it I kinda like it.  As a framework/middleware we should really have events like this to make it easier on the modules.  Just do a $rootScope.broadcast(events) and we handle everything for the modules.

As for 404, the only reason I don't like it is because the backend throws them a lot.  I don't think that redirecting to a bad-request route is a better option then just displaying the error on the client since the bad request error message is variable.  Then the user has to hit the back button to get back to their previous state.

trainerbill:
@rhutchison Removed Modal, Removed Pub/Sub, changed unauthorized to forbidded to align with the http status code.  401 goes to authentication 403 goes to forbidden.

rhutchison:
LGTM - thank you.

trainerbill:
@lirantal.  This is ready to merge

lirantal:
@ilanbiala ?

mleanos:
@trainerbill I pulled this branch and tested. A few things I noticed... Forgive me if you're aware of these issues.. 

When not signed in, I go to an invalid route. I'm redirected to the Bad Request state. Now if sign in successfully, I'm redirected back to the Bad Request state. 

If a 400 status is returned from the server, it redirects to the Bad Request, even if the 400 was caused by an invalid form submission; for instance, incorrect credentials at login.

I saw mention of the issue where the user can add `/edit` to an article they are viewing & get to that state. However, they are correctly sent to the Forbidden state if they attempt to submit the form. What is the plan with addressing that? Wouldn't we want to just avoid allowing that state transition in the first place? Perhaps, that's outside the scope of this PR.

@trainerbill Overall this looks really good. Thanks.

trainerbill:
@mleanos Thanks for taking it for a test drive.  I was actually worried on the bad request 400 interceptor as the backend seems to throw them quite a bit.  I am going to remove the 400 interceptor just to get the PR in.  The forbidden route and interceptor is what I really want in.  @rhutchison any issue with this?

modules/core/client/services/interceptors/auth.interceptor.client.service.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+//Articles service used for communicating with the articles REST endpoints
```
rhutchison: remove unrelated comment

modules/core/client/app/init.js
```diff
@@ -4,13 +4,16 @@
 angular.module(ApplicationConfiguration.applicationModuleName, ApplicationConfiguration.applicationModuleVendorDependencies);
 
 // Setting HTML5 Location Mode
-angular.module(ApplicationConfiguration.applicationModuleName).config(['$locationProvider',
-  function ($locationProvider) {
-    $locationProvider.html5Mode(true).hashPrefix('!');
-  }
+angular.module(ApplicationConfiguration.applicationModuleName).config(['$locationProvider', '$httpProvider',
+function ($locationProvider, $httpProvider) {
```
rhutchison: fix indentation

modules/core/client/app/init.js
```diff
@@ -24,11 +27,12 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
...
         event.preventDefault();
-        $state.go('authentication.signin', {}, {
-          notify: false
-        }).then(function () {
-          $rootScope.$broadcast('$stateChangeSuccess', 'authentication.signin', {}, toState, toParams);
-        });
+        if (Authentication.user !== undefined && typeof Authentication.user === 'object') {
+          $state.go('forbidden');
+        }
+        else {
```
rhutchison: should be on previous line

modules/core/client/config/core.client.routes.js
```diff
@@ -2,20 +2,28 @@
 
 // Setting up route
 angular.module('core').config(['$stateProvider', '$urlRouterProvider',
-  function ($stateProvider, $urlRouterProvider) {
+function ($stateProvider, $urlRouterProvider) {
```
rhutchison: fix indentation

rhutchison:
LGTM thanks @trainerbill

mleanos:
@trainerbill I see the 400 is removed, so the issue with the invalid form submission is resolved. Thanks. 

I'm still seeing that when I get sent to the not-found state, and then login I'm getting redirected back to the not-found state. Were you waiting to handle that in a separate PR?

Other than that everything looks good to me.

@rhutchison I relinquish my stance on the order of which to merge these PR's. Sounds like we need to discuss, or get feedback, on #807 anyways. 

This LGTM, and appears to be ready for merging.

codydaig:
LGTM

rhutchison:
@mleanos @trainerbill @codydaig  Please review #843 
> I'm still seeing that when I get sent to the not-found state, and then login I'm getting redirected back to the not-found state. Were you waiting to handle that in a separate PR?

trainerbill:
Don't merge yet.  Writing tests

trainerbill:
@lirantal @rhutchison Ready to merge

rhutchison:
LGTM

PR #797, opened 2015-08-13
used __dirname instead of process.cwd(), this leads to server errors ...

bastianwegge:
…if the executable is not run in the same directory.

Simple thing, tested the new meanjs on our farm and instantly ran into this.

The usual way to reproduce this is to start a node instance not by beeing in the folder itself. Everybody can reproduce this behaviour on his own machine, but in my purpose this is to run it on the server.

1. clone mean
2. run ´npm install´
3. cd into any other directory (server-specific would be like 1 directories up like `/root/app/`)
4. `node app/server.js` or as I did: forever start `/root/app/server.js`
--> process.cwd() is going for the directory in which the process was started.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -98,7 +98,7 @@ var initGlobalConfigFolders = function (config, assets) {
   };
 
   // Setting globbed client paths
-  config.folders.client = getGlobbedPaths(path.join(__dirname, 'modules/*/client/'), __dirname.replace(new RegExp(/\\/g), '/'));
+  config.folders.client = getGlobbedPaths(path.join(__dirname, '../modules/*/client/'), path.resolve(__dirname + '/..'));
 };
 
 /**
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -63,7 +63,7 @@ module.exports.initMiddleware = function (app) {
   }));
 
   // Initialize favicon middleware
-  app.use(favicon('./modules/core/client/img/brand/favicon.ico'));
+  app.use(favicon(__dirname + '../../../modules/core/client/img/brand/favicon.ico'));
 
   // Environment dependent middleware
   if (process.env.NODE_ENV === 'development') {


bastianwegge:
`So why not grunt?`

I've been running forever with upstart ever since, what's the benefit?

lirantal:
Was just wondering.

Anyway:
1. meanjs app is on `/dir1/meanjs_app/server.js`
2. I `cd` to a directory outside meanjs_app so I'm now in `/dir1`
3. I'm still getting errors:
image

codydaig:
Why not use path.resolve and just base the paths from the root of the project?

bastianwegge:
The Problem with this is (as you can see) `modules` ist defined twice. I'll adjust and test that as soon as I'm back home (probably about 3-4 hours).

bastianwegge:
So I've corrected the use of favicon. My general idea was to make mean more independent in the case of server or local usages, since I'm not the only one starting it not directly from the directory.

It's your decision to take it or leave it. Would be a pleasure for me to contribute to you. Any hints are highly appreciated.

lirantal:
@bastianwegge we're all about community feedback, thanks, but it doesn't mean we merge code blindly so bear with us :)

What do you think about @codydaig suggestion to just use path.resolve?

bastianwegge:
I see no example whatsoever. Is the suggestion that we should use a base path? @codydaig 
path.resolve is a nice method as is, but how "base" them, that's the whole point of the PR

config/config.js
```diff
@@ -144,24 +144,24 @@ var initGlobalConfig = function () {
   validateEnvironmentVariable();
 
   // Get the default assets
-  var defaultAssets = require(path.join(process.cwd(), 'config/assets/default'));
+  var defaultAssets = require(path.join(__dirname, 'assets/default'));
```
codydaig: try:
`path.resolve(__dirname, 'assets/default')`

codydaig:
@bastianwegge I made a comment on a specific line with an example of what I meant by path.resolve. This will resolve to an absolute path, from the main server.js file.

bastianwegge:
@codydaig I don't really get the point, since both methods exactly return the same value.

bildschirmfoto 2015-08-17 um 07 25 51

codydaig:
@bastianwegge If they resolve to the same thing then it's prefference. I prefer absolute over relative, but that aside there's still issues to be resolved. When I pulled your branch down and tested it runs fine in the console, but when I get a 404 error when trying to view the app in the browser so there's still some missing components to this PR to make it do what your attempting to do.

bastianwegge:
Hey @codydaig, since both use absolute paths your point is not really a point.
Where do you get the 404? Since I'm running the same version without getting errors.

codydaig:
image

codydaig:
screen shot 2015-08-16 at 11 27 18 pm

bastianwegge:
Oh yes you're right, will have a look at this in the evening.

codydaig:
@bastianwegge Any luck?

bastianwegge:
Sorry I'm busy on a customer project. Will take a look at it soon.

bastianwegge:
I've checked the flaw and it's making its way through. I thought about making the rootdirectory globally but I wanted to hear some oppinion on the global namespace from you guys.

@codydaig  @lirantal what do you think about that? Since this would increase compatibility of mean it might be worth a thought... or someone has a better idea. 

Wouldn't mind dropping the PR if you don't think it's worth the way.

lirantal:
What does it mean? is there an update you want to commit to make sure everything runs ok?

bastianwegge:
I have begun making it work again. The actual flaw was that the glob-function is looking in the wrong directory and a lot of files already depend on the process to be run by grunt. That's why I asked if you think it's worth taking the way or if we should drop the Pullrequest at all.

lirantal:
I guess you're referring to the glob function in config/config? to refactor it?

bastianwegge:
https://github.com/meanjs/mean/blob/master/config/config.js#L31
could be fixed using the following:

```
var path = require('path'),
  rootDirectory = path.resolve(__dirname + '/..');

var files = glob.sync(rootDirectory + '/' + globPatterns);
```

Actually not much of a deal but the change keeps hitting through. Because most of the paths used in mean depend on the process being executed by grunt. Just some small examples:

https://github.com/meanjs/mean/blob/master/modules/articles/server/controllers/articles.server.controller.js#L9
https://github.com/meanjs/mean/blob/master/modules/users/server/config/users.server.config.js#L9

The list goes on and on.

lirantal:
Got it. We'll drop this change for now then as I'm unable to measure the exact impact and risk and this may be quite a daunting task to overtake.

I'm wondering though why must you run the server from outside the directory? why can't you `cd` into it  and then run node?

bastianwegge:
The point was that the usual way upstart-script takes an application-directory path. So if I use forever with upstart I'd point to a directory with a binary that's defined like so:

```
exec forever \
      --pidFile $PIDFILE \
      -a \
      -l $LOG \
      --minUptime $MIN_UPTIME \
      --spinSleepTime $SPIN_SLEEP_TIME \
      start $APPLICATION_PATH
```

But I guess I have to figure out another way then.

codydaig:
@bastianwegge Personally, I use pm2 and don't have any issues with this.

bastianwegge:
thank you @codydaig I'll definetly give it a try

lirantal:
we also use pm2 for an enterprise app, that's why I was asking about this use case.

PR #798, opened 2015-08-13
Express sessionKey configuration option

lirantal:
Adding support for sessionKey configuration parameter to allow to easily change the session key that is used to hold the session value. The default is a generic sessionId key to introduce security through obscurity.


trainerbill:
Pretty small change.  I grabbed your branch and ran around on the application and it seems to work properly.  Not losing session or anything.  LGTM

lirantal:
thanks @trainerbill

codydaig:
LGTM

PR #799, opened 2015-08-13
Adding suite of tests for the e-mail validation field in the users model

lirantal:
This further strengthens our test suites from 21 tests to 35 already!

image

### DIFF
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -155,6 +155,261 @@ describe('User Model Unit Tests:', function () {
 
   });
 
+  describe("User E-mail Validation Tests", function() {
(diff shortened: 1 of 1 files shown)


trainerbill:
@lirantal Not that having more tests is a bad thing, but we are using validator so wouldn't this be overkill?  Guessing that module has the proper tests already.  Unless there is a way to add email outside of validator, which would probably be bad.   The more tests we have the longer CI takes.

codydaig:
LGTM.

lirantal:
@trainerbill that's *exactly* the point - imagine a scenario that someone in the future will replace the `validator` package with something else? or that we upgrade to a newer validator package? these tests ensure that the expected behavior is always maintained.

Also I wouldn't worry that much about tests time and rather would aim for a high code coverage.

PR #793, opened 2015-08-13
Database seeding

trainerbill:
New PR for:

https://github.com/meanjs/mean/pull/785

PR #801, opened 2015-08-13
Add seed user instructions to README

codydaig:
In reference to #793.

Not entirely sure how much I like the wording, so any suggestions would be awesome!

@trainerbill

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,21 @@ $ grunt prod
 
 * explore `config/env/production.js` for production environment configuration options
 
+### Running with User Seed
+To have default account(s) seeded at runtime:
+
+In Development:
+```bash
+MONGO_SEED=true grunt
+```
+It will try to seed the users 'user' and 'admin'. If one of the user already exists, it will display an error message on the console. Just grab the passwords from the console.
+
+In Production:
+```bash
+MONGO_SEED=true grunt prod
+```
+This will seed the admin user one time if the user does not already exist. You have to copy the password from the console and save it.
+
 ### Running with TLS (SSL)
 Application will start by default with secure configuration (SSL mode) turned on and listen on port 8443.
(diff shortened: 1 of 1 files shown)


lirantal:
I think the documentation needs an update to correct it - for either dev or prod both user and admin accounts are seeded if they aren't available.

README.md
```diff
@@ -93,6 +93,21 @@ $ grunt prod
...
 * explore `config/env/production.js` for production environment configuration options
 
+### Running with User Seed
+To have an 'admin' and 'user' user account seeded at runtime:
+
+In Development:
+```bash
+MONGO_SEED=true grunt
+```
+This will seed the users everytime, just grab the password from the console.
```
vaucouleur: "It will try to seed the users everytime. If one of the user already exists, it will display an error message on the console. Just grab the passwords from the console"

https://github.com/meanjs/mean/blob/ea97f2e7d83281a5bf34cb294ce9f1483135dbf2/config/lib/seed.js#L55

README.md
```diff
@@ -93,6 +93,21 @@ $ grunt prod
 
 * explore `config/env/production.js` for production environment configuration options
 
+### Running with User Seed
+To have an 'admin' and 'user' user account seeded at runtime:
```
vaucouleur: May be: "To have default account(s) seeded at runtime". Since admin is not created in development mode

lirantal: I agree with the correction there for the admin account, though let's make sure we also specify the actual username used for seeding so it's not ambiguous.

lirantal:
@codydaig whenever you can address the comments, other than that looks good.

codydaig:
@lirantal Totally forgot about this PR. I updated and squashed.

lirantal:
Great, merging.

PR #807, opened 2015-08-15
Articles admin feature

mleanos:
This is a work in progress, but is intended to gather feedback. There has been talk of separating the Articles views between authenticated User's and non-authenticated *Guests*. This is an attempt to explore going down this road. However, it turned into more of a *Public* states exploration :) 

I know there is competing changes to the Menu' service; #780 . I feel like the `shouldRender` method has been batted around lately. Once we determine how to proceed with implementing these desired features, I will adjust this PR to reflect the solution to the Menu's service refactoring. I think the difficulty in working with the `shouldRender` with these types of modifications, again shows the need for refactoring the Menu service.

@rhutchison @codydaig @trainerbill What do y'all think of this approach? Adding the Public route to the Core config felt a little funny. But it could be a decent solution. This would allow us to create *Public* routes for other modules.

### DIFF
--- a/modules/articles/client/config/articles-admin.client.config.js
+++ b/modules/articles/client/config/articles-admin.client.config.js
@@ -1,0 +1,17 @@
+﻿(function () {
+  'use strict';
+
+  // Configuring the Articles Admin module
+  angular
+    .module('articles.admin')
+    .run(menuConfig);
+
+  menuConfig.$inject = ['menuService'];
+
+  function menuConfig(Menus) {
+    Menus.addSubMenuItem('topbar', 'admin', {
+      title: 'Manage Articles',
+      state: 'admin.articles.list'
+    });
+  }
+}());
--- a/modules/articles/client/config/articles-admin.client.routes.js
+++ b/modules/articles/client/config/articles-admin.client.routes.js
@@ -1,0 +1,65 @@
+﻿(function () {
+  'use strict';
+
+  angular
+    .module('articles.admin.routes')
+    .config(routeConfig);
+
+  routeConfig.$inject = ['$stateProvider'];
+
+  function routeConfig($stateProvider) {
+    $stateProvider
+      .state('admin.articles', {
+        abstract: true,
+        url: '/articles',
+        template: '<ui-view/>'
+      })
+      .state('admin.articles.list', {
+        url: '',
+        templateUrl: 'modules/articles/client/views/admin/list-articles.client.view.html',
+        controller: 'ArticlesListController',
+        controllerAs: 'vm',
+        data: {
+          roles: ['admin']
+        }
+      })
+      .state('admin.articles.create', {
+        url: '/create',
+        templateUrl: 'modules/articles/client/views/admin/form-article.client.view.html',
+        controller: 'ArticlesController',
+        controllerAs: 'vm',
+        data: {
+          roles: ['admin']
+        },
+        resolve: {
+          articleResolve: newArticle
+        }
+      })
+      .state('admin.articles.edit', {
+        url: '/:articleId/edit',
+        templateUrl: 'modules/articles/client/views/admin/form-article.client.view.html',
+        controller: 'ArticlesController',
+        controllerAs: 'vm',
+        data: {
+          roles: ['admin']
+        },
+        resolve: {
+          articleResolve: getArticle
+        }
+      });
+  }
+
+  getArticle.$inject = ['$stateParams', 'ArticlesService'];
+
+  function getArticle($stateParams, ArticlesService) {
+    return ArticlesService.get({
+      articleId: $stateParams.articleId
+    }).$promise;
+  }
+
+  newArticle.$inject = ['ArticlesService'];
+
+  function newArticle(ArticlesService) {
+    return new ArticlesService();
+  }
+}());
--- a/modules/articles/client/config/articles.client.routes.js
+++ b/modules/articles/client/config/articles.client.routes.js
@@ -23,32 +23,6 @@
           pageTitle: 'Articles List'
         }
       })
-      .state('articles.create', {
-        url: '/create',
-        templateUrl: 'modules/articles/client/views/form-article.client.view.html',
-        controller: 'ArticlesController',
-        controllerAs: 'vm',
-        resolve: {
-          articleResolve: newArticle
-        },
-        data: {
-          roles: ['user', 'admin'],
-          pageTitle: 'Articles Create'
-        }
-      })
-      .state('articles.edit', {
-        url: '/:articleId/edit',
-        templateUrl: 'modules/articles/client/views/form-article.client.view.html',
-        controller: 'ArticlesController',
-        controllerAs: 'vm',
-        resolve: {
-          articleResolve: getArticle
-        },
-        data: {
-          roles: ['user', 'admin'],
-          pageTitle: 'Edit Article {{ articleResolve.title }}'
-        }
-      })
       .state('articles.view', {
         url: '/:articleId',
         templateUrl: 'modules/articles/client/views/view-article.client.view.html',
@@ -69,11 +43,5 @@
     return ArticlesService.get({
       articleId: $stateParams.articleId
     }).$promise;
-  }
-
-  newArticle.$inject = ['ArticlesService'];
-
-  function newArticle(ArticlesService) {
-    return new ArticlesService();
   }
 }());
--- a/modules/articles/tests/client/admin.articles.client.controller.tests.js
+++ b/modules/articles/tests/client/admin.articles.client.controller.tests.js
@@ -1,0 +1,168 @@
+﻿(function () {
+  'use strict';
+
+  describe('Articles Controller Tests', function () {
+    // Initialize global variables
+    var ArticlesController,
+      $scope,
+      $httpBackend,
+      $state,
+      Authentication,
+      ArticlesService,
+      mockArticle;
+
+    // The $resource service augments the response object with methods for updating and deleting the resource.
+    // If we were to use the standard toEqual matcher, our tests would fail because the test values would not match
+    // the responses exactly. To solve the problem, we define a new toEqualData Jasmine matcher.
+    // When the toEqualData matcher compares two objects, it takes only object properties into
+    // account and ignores methods.
+    beforeEach(function () {
+      jasmine.addMatchers({
+        toEqualData: function (util, customEqualityTesters) {
+          return {
+            compare: function (actual, expected) {
+              return {
+                pass: angular.equals(actual, expected)
+              };
+            }
+          };
+        }
+      });
+    });
+
+    // Then we can start by loading the main application module
+    beforeEach(module(ApplicationConfiguration.applicationModuleName));
+
+    // The injector ignores leading and trailing underscores here (i.e. _$httpBackend_).
+    // This allows us to inject a service but then attach it to a variable
+    // with the same name as the service.
+    beforeEach(inject(function ($controller, $rootScope, _$state_, _$httpBackend_, _Authentication_, _ArticlesService_) {
+      // Set a new global scope
+      $scope = $rootScope.$new();
+
+      // Point global variables to injected services
+      $httpBackend = _$httpBackend_;
+      $state = _$state_;
+      Authentication = _Authentication_;
+      ArticlesService = _ArticlesService_;
+
+      // create mock article
+      mockArticle = new ArticlesService({
+        _id: '525a8422f6d0f87f0e407a33',
+        title: 'An Article about MEAN',
+        content: 'MEAN rocks!'
+      });
+
+      // Mock logged in user
+      Authentication.user = {
+        roles: ['user']
+      };
+
+      // Initialize the Articles controller.
+      ArticlesController = $controller('ArticlesController as vm', {
+        $scope: $scope,
+        articleResolve: {}
+      });
+
+      // Spy on state go
+      spyOn($state, 'go');
+    }));
+
+    describe('vm.save() as create', function () {
+      var sampleArticlePostData;
+
+      beforeEach(function () {
+        // Create a sample article object
+        sampleArticlePostData = new ArticlesService({
+          title: 'An Article about MEAN',
+          content: 'MEAN rocks!'
+        });
+
+        $scope.vm.article = sampleArticlePostData;
+      });
+
+      it('should send a POST request with the form input values and then locate to new object URL', inject(function (ArticlesService) {
+        // Set POST response
+        $httpBackend.expectPOST('api/articles', sampleArticlePostData).respond(mockArticle);
+
+        // Run controller functionality
+        $scope.vm.save(true);
+        $httpBackend.flush();
+
+        // Test URL redirection after the article was created
+        expect($state.go).toHaveBeenCalledWith('admin.articles.list');
+      }));
+
+      it('should set $scope.vm.error if error', function () {
+        var errorMessage = 'this is an error message';
+        $httpBackend.expectPOST('api/articles', sampleArticlePostData).respond(400, {
+          message: errorMessage
+        });
+
+        $scope.vm.save(true);
+        $httpBackend.flush();
+
+        expect($scope.vm.error).toBe(errorMessage);
+      });
+    });
+
+    describe('vm.save() as update', function () {
+      beforeEach(function () {
+        // Mock article in $scope
+        $scope.vm.article = mockArticle;
+      });
+
+      it('should update a valid article', inject(function (ArticlesService) {
+        // Set PUT response
+        $httpBackend.expectPUT(/api\/articles\/([0-9a-fA-F]{24})$/).respond();
+
+        // Run controller functionality
+        $scope.vm.save(true);
+        $httpBackend.flush();
+
+        // Test URL location to new object
+        expect($state.go).toHaveBeenCalledWith('admin.articles.list');
+      }));
+
+      it('should set $scope.vm.error if error', inject(function (ArticlesService) {
+        var errorMessage = 'error';
+        $httpBackend.expectPUT(/api\/articles\/([0-9a-fA-F]{24})$/).respond(400, {
+          message: errorMessage
+        });
+
+        $scope.vm.save(true);
+        $httpBackend.flush();
+
+        expect($scope.vm.error).toBe(errorMessage);
+      }));
+    });
+
+    describe('vm.remove()', function () {
+      beforeEach(function () {
+        // Setup articles
+        $scope.vm.article = mockArticle;
+      });
+
+      it('should delete the article and redirect to articles', function () {
+        // Return true on confirm message
+        spyOn(window, 'confirm').and.returnValue(true);
+
+        $httpBackend.expectDELETE(/api\/articles\/([0-9a-fA-F]{24})$/).respond(204);
+
+        $scope.vm.remove();
+        $httpBackend.flush();
+
+        expect($state.go).toHaveBeenCalledWith('admin.articles.list');
+      });
+
+      it('should should not delete the article and not redirect', function () {
+        // Return false on confirm message
+        spyOn(window, 'confirm').and.returnValue(false);
+
+        $scope.vm.remove();
+
+        expect($state.go).not.toHaveBeenCalled();
+      });
+    });
+  });
+}());
--- a/modules/articles/tests/server/admin.article.server.routes.tests.js
+++ b/modules/articles/tests/server/admin.article.server.routes.tests.js
@@ -1,0 +1,282 @@
+﻿'use strict';
+
+var should = require('should'),
+  request = require('supertest'),
+  path = require('path'),
+  mongoose = require('mongoose'),
+  User = mongoose.model('User'),
+  Article = mongoose.model('Article'),
+  express = require(path.resolve('./config/lib/express'));
+
+/**
+ * Globals
+ */
+var app,
+  agent,
+  credentials,
+  user,
+  article;
+
+/**
+ * Article routes tests
+ */
+describe('Article Admin CRUD tests', function () {
+  before(function (done) {
+    // Get application
+    app = express.init(mongoose);
+    agent = request.agent(app);
+
+    done();
+  });
+
+  beforeEach(function (done) {
+    // Create user credentials
+    credentials = {
+      username: 'username',
+      password: 'M3@n.jsI$Aw3$0m3'
+    };
+
+    // Create a new user
+    user = new User({
+      firstName: 'Full',
+      lastName: 'Name',
+      displayName: 'Full Name',
+      email: 'test@test.com',
+      roles: ['user', 'admin'],
+      username: credentials.username,
+      password: credentials.password,
+      provider: 'local'
+    });
+
+    // Save a user to the test db and create new article
+    user.save(function () {
+      article = {
+        title: 'Article Title',
+        content: 'Article Content'
+      };
+
+      done();
+    });
+  });
+
+  it('should be able to save an article if logged in', function (done) {
+    agent.post('/api/auth/signin')
+      .send(credentials)
+      .expect(200)
+      .end(function (signinErr, signinRes) {
+        // Handle signin error
+        if (signinErr) {
+          return done(signinErr);
+        }
+
+        // Get the userId
+        var userId = user.id;
+
+        // Save a new article
+        agent.post('/api/articles')
+          .send(article)
+          .expect(200)
+          .end(function (articleSaveErr, articleSaveRes) {
+            // Handle article save error
+            if (articleSaveErr) {
+              return done(articleSaveErr);
+            }
+
+            // Get a list of articles
+            agent.get('/api/articles')
+              .end(function (articlesGetErr, articlesGetRes) {
+                // Handle article save error
+                if (articlesGetErr) {
+                  return done(articlesGetErr);
+                }
+
+                // Get articles list
+                var articles = articlesGetRes.body;
+
+                // Set assertions
+                (articles[0].user._id).should.equal(userId);
+                (articles[0].title).should.match('Article Title');
+
+                // Call the assertion callback
+                done();
+              });
+          });
+      });
+  });
+
+  it('should be able to update an article if signed in', function (done) {
+    agent.post('/api/auth/signin')
+      .send(credentials)
+      .expect(200)
+      .end(function (signinErr, signinRes) {
+        // Handle signin error
+        if (signinErr) {
+          return done(signinErr);
+        }
+
+        // Get the userId
+        var userId = user.id;
+
+        // Save a new article
+        agent.post('/api/articles')
+          .send(article)
+          .expect(200)
+          .end(function (articleSaveErr, articleSaveRes) {
+            // Handle article save error
+            if (articleSaveErr) {
+              return done(articleSaveErr);
+            }
+
+            // Update article title
+            article.title = 'WHY YOU GOTTA BE SO MEAN?';
+
+            // Update an existing article
+            agent.put('/api/articles/' + articleSaveRes.body._id)
+              .send(article)
+              .expect(200)
+              .end(function (articleUpdateErr, articleUpdateRes) {
+                // Handle article update error
+                if (articleUpdateErr) {
+                  return done(articleUpdateErr);
+                }
+
+                // Set assertions
+                (articleUpdateRes.body._id).should.equal(articleSaveRes.body._id);
+                (articleUpdateRes.body.title).should.match('WHY YOU GOTTA BE SO MEAN?');
+
+                // Call the assertion callback
+                done();
+              });
+          });
+      });
+  });
+
+  it('should not be able to save an article if no title is provided', function (done) {
+    // Invalidate title field
+    article.title = '';
+
+    agent.post('/api/auth/signin')
+      .send(credentials)
+      .expect(200)
+      .end(function (signinErr, signinRes) {
+        // Handle signin error
+        if (signinErr) {
+          return done(signinErr);
+        }
+
+        // Get the userId
+        var userId = user.id;
+
+        // Save a new article
+        agent.post('/api/articles')
+          .send(article)
+          .expect(400)
+          .end(function (articleSaveErr, articleSaveRes) {
+            // Set message assertion
+            (articleSaveRes.body.message).should.match('Title cannot be blank');
+
+            // Handle article save error
+            done(articleSaveErr);
+          });
+      });
+  });
+
+  it('should be able to delete an article if signed in', function (done) {
+    agent.post('/api/auth/signin')
+      .send(credentials)
+      .expect(200)
+      .end(function (signinErr, signinRes) {
+        // Handle signin error
+        if (signinErr) {
+          return done(signinErr);
+        }
+
+        // Get the userId
+        var userId = user.id;
+
+        // Save a new article
+        agent.post('/api/articles')
+          .send(article)
+          .expect(200)
+          .end(function (articleSaveErr, articleSaveRes) {
+            // Handle article save error
+            if (articleSaveErr) {
+              return done(articleSaveErr);
+            }
+
+            // Delete an existing article
+            agent.delete('/api/articles/' + articleSaveRes.body._id)
+              .send(article)
+              .expect(200)
+              .end(function (articleDeleteErr, articleDeleteRes) {
+                // Handle article error error
+                if (articleDeleteErr) {
+                  return done(articleDeleteErr);
+                }
+
+                // Set assertions
+                (articleDeleteRes.body._id).should.equal(articleSaveRes.body._id);
+
+                // Call the assertion callback
+                done();
+              });
+          });
+      });
+  });
+
+  it('should be able to get a single article if signed in and verify the custom "isCurrentUserOwner" field is set to "true"', function (done) {
+    // Create new article model instance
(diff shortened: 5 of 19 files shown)


mleanos:
Okay, so now this PR is more of what it should be. I've scrapped the whole `/public` routes per the advice from @rhutchison and @codydaig 

I've separated the Articles views into two distinct views; one for Admins & one for Users (including non-Authenticated users). With these changes, only User's with the `admin` role can create & edit Articles. This may be a big change, and need more input from the community. But it's something worth exploring. 

I know the test's are failing. I will work on updating the tests to account for these changes, once the direction of this feature is defined more clearly & we're ready to merge.

modules/articles/client/config/articles.client.config.js
```diff
@@ -8,20 +8,14 @@ angular.module('articles').run(['Menus',
       title: 'Articles',
       state: 'articles',
       type: 'dropdown',
-      roles: ['*']
+      roles: ['*', 'user', 'admin']
```
rhutchison: this could just be `roles: ['*']`

modules/articles/client/config/articles.client.config.js
```diff
@@ -8,20 +8,14 @@ angular.module('articles').run(['Menus',
...
-      state: 'articles.list'
-    });
-
-    // Add the dropdown create item
-    Menus.addSubMenuItem('topbar', 'articles', {
-      title: 'Create Articles',
-      state: 'articles.create',
-      roles: ['user']
+      state: 'articles.list',
+      roles: ['*', 'user', 'admin']
```
rhutchison: this could just be roles: ['*']

mleanos:
Added additional tests for Admin functionality when interacting with Articles. Also, removed unnecessary tests for non-Admin users.

Addressed @rhutchison line notes.

mleanos:
Removed a few files that only had spacing changes.

vaucouleur:
@mleanos In the framework that I worked this, people chose instead to have a single view for both the user and the admin, and instead played with showing/hiding buttons (delete/edit/..) depending on the access rights, and of course, duplicating the checks at the server to avoid "by-pass". The main reason being: to avoid duplicate code. The default Article module is tiny, hence the code duplication issue might not be obvious using the default meanjs code (and by code here, I mean both the html code and the js code), but in a "real" SPA site, with complex screens, code duplication is a maintenance nightmare and should be avoided at all cost. Custom directives (or similar "concepts") can help mitigate the code duplication by factoring out a bit a of the code, but typically the problem remain: you want your screens to be defined in a single location.

trainerbill:
I have not had time to look much into this PR but I would have to agree with @vaucouleur.  The only reason we had to completely write views for user admin is because there was nothing to start with.  Why not just direct them to the articles views and do some small modifications to the views based on the role?

mleanos:
@vaucouleur @trainerbill I think the main change introduced in this PR is what we should be focusing on. With these changes, *ONLY* an admin can create/edit Articles. This stems from a conversation @rhutchison and I had. We interpret the intention of the Articles module to be one that provides Article type functionality, rather than a user Post.

So naturally, if only a user with the `admin` role can create/edit Articles, there is no need for a User to ever have access to the same view/route that *allows* edits. Even with this reasoning, we could keep the current implementation by putting checks in the view to show/hide elements. However, once the *editing* aspects of the view gets more complex, it probably should be moved out on it's own anyways. IMO, we should hide as much implementation from the User as possible. Separating the views accomplishes this.

rhutchison:
Since I gave direction to @mleanos, I'm a little bias, but I agree with @mleanos :dancers: 

Maybe we just never understood what articles is trying to be, but by it's name, I see it being similar to how it would be implemented in any CMS.  We have been open to discussion about what articles in this project really means, and I was hoping that we would all be able to come to a consensus and build it out further.

I don't think it's a great idea to build out a shared view and then have ng-show/hide or ng-if statements everywhere.  Eventually it will become unmanageable and not make much sense to have in the same view.

Things we have discussed are active/expiration dates, rich text editing, additional roles like publisher/editor, etc.

If someone wants to create an issue for improvement, we can have a more directed discussion.  Additionally, we can discuss in the [gitter room](https://gitter.im/meanjs/mean)

rhutchison:
@mleanos create/edit can be 1 page - article-form.html (or some naming convention that won't get my hand slapped)

vaucouleur:
It's a very interesting and noble discussion. To me, this is one of those key questions at the heart of software design. Let's consider two extreme cases:

- Case 1: What we have is the classical, MVC example. You have some data that you want to display using different views. One the is a pie chart, the other is a table display.

- Case 2: What we have is a complex SPA screen based on a "business object" (article, user, bank account, etc.). We have lots of things going-on in the screen, and what we want is to disable or enable some features -- typically, show/hide a delete button.

Naturally, cases are not always so extreme, we have a continuum. Now, if you consider that you are closer to Case 1, it makes sense to have different views. If you feel that you are closer to Case 2, it makes sense to avoid code duplication and have a single view.

Whatever your choice, what would be nice (and I am not sure whether or not you this in mind for the long-run), would be to avoid hard-coding as much as possible the roles+permissions on the client-side. Ideally, to me at least, the permission logic should be centralized in the policy file on the server side, and come back-up to the client-side, and be checked (again), when the calls comes back to server side (on save/update/etc.), using the same centralized policy file.

A concrete example would the following: 

- Says, our articles have a theme: sport, politics, science etc. 
- These give rise to "editors" who are admins just for the articles that are in their domain of expertise. 
- Hence, we have a role "admin_science", a role "admin_politics", etc. 
- When a user GET an article, a function in the policy file for article get invoked that returns the permissions for this user. 
- Since our user has an "admin_science" role and the article is about science, he gets the permissions  ['delete', 'edit', ...], and we are either routed toward the right screen (Case 1 above), or the features are enabled/disabled in the screen (Case 2 above).

rhutchison:
@vaucouleur without going in to much detail, regarding case 2, I can see this being managed through ui-view and still support multiple controller/views.

vaucouleur:
@rhutchison That would be nice. For example a part of the screen that is only available for users that have write access ? 

What about permissioning ? Are you planning to hard-code on the client side that there is a role called 'admin' that can do exactly this-and-that ? Roles and permissions are typically modified dynamically  with permissions stored in the database using for example acl.mongodbBackend

vaucouleur:
I'll try to come up with a PR for the permission idea, and I will try to make it so that it integrates gracefully with your latest work on client side routing/roles/"article admin", etc.

mleanos:
@vaucouleur I think role based permissions are sufficient in this case. What need do you see the extra permissions satisfying?

vaucouleur:
@mleanos I have a generalization in mind. In the acl module, they are roles, resources, and permissions. Hence role1 can have delete permission over object1, but so do role2 and role3. The idea would be that this information ("who can do what on what") remains centralized on the server side, in the policy files. Multiple roles should be able edit/delete the objects. Feature enabling/disabling would permission based and not role based, and not hard-coded in the client code.

mleanos:
@vaucouleur Ok I'd probably have to see the client-side implementation before I had too much of an opinion :) However, the ACL's on the server are already setup to handle the permissions; which should be sufficient.

I think what you're driving at is a client-based ACL type implementation? That would probably be the route to go. However, I'm not sure if it would solve the problem of "hard-coding" logic (feature enabling/disabling). I still think there is a need to separate views based on object *permissions* or roles. I'm very interested in seeing what you come up with though.

vaucouleur:
@mleanos Not sure what you mean by a "client-side ACL implementation". Again, in my mind, the ACL logic should be on the server side. And, "as much as possible" this logic should be *not* be duplicated on the client side (and often "cannot be duplicated" since the role/resource/permission triples are typically stored in the database, this is quite standard -- except for the most simple sites). I agree that "Views should be separated"  (or alternatively that features should be disabled by ng-show/hide -- "pick your poison"), but not based on the roles of the user. IMHO, it should be based on what the *backend* policy said: "the current user that made this GET request on that object has exactly that set of permissions on it". 

Hopefully the code should be less ambiguous. I hope that I can come up with something interesting.

mleanos:
@vaucouleur I meant that what you're proposing sounds like a client-side ACL policy type implementation. Currently, the server-side ACL policies handle the permissions *per role* for the API routes. The server-side is all squared away. IMO, the back-end and front-end permissions should not care about each other; meaning they can be implemented in different ways that are not related to the other. The back-end should just care if a user has permission to perform a specific action (request *method*) on a particular route.

What I'm confused about, I guess, is what you're referring to when you say 
> Feature enabling/disabling would permission based and not role based, and not hard-coded in the client code.

What do you have in mind to implement such *permissions* based access/functionality on the client? To me that's what sounds similar to the ACL policies on the back-end. This is the part that I'm confused about. 

I think we're pretty much in agreement here. But I'd like to understand exactly what you mean.

vaucouleur:
@mleanos We could be in agreement, and what you're trying to achieve could be orthogonal to what I described, although it's not fully clear at this point. Let me take a minimalist example to answer you question, and to check whether we agree:

Let's take completely the "client-side perspective". Let's assume a user is signed-in. He already GET, well 'GOT' actually :), an article. Now the client-side must decide whether the delete button is shown to him with respect to that article (how this is implemented: views, or anything else, is not specified). The only information that we have is the user _id, the article _id, and the user.roles array that shows: ['user', 'role1', 'role2']. No other useful piece of information is available. Should the delete button be visible ?
- If you answer is "From a client-side perspective alone, we have no way of knowing", then we agree
- If your answer is "The role 'admin' is not in the roles array of the user therefore it should not be visible", then we disagree.

mleanos:
@vaucouleur I actually would answer with both of those points.. and then land on the 2nd as the solution. We must figure out *WHO* can use the delete button. Unless we assign a set of permissions that would be *user* specific, on each of the articles. Which could look like this...

```
permissions: [
  user: { _id: fd3334fas33x3fdds },
  rights: ['read', 'delete']
]
```
Now this becomes more of a *sharing* feature for articles. Which I don't think is in the scope of what articles are. I think it should be role based. 

We must *decide* on what the business rule/logic would be. Just as we decide that only logged in user's can access the Chat feature.

And if you're answer is the first point, then how do you plan on solving that problem?

vaucouleur:
@mleanos I'll answer your question (right now I'm fighting against the favicon.ico..) But in a few words, and IMHO, we cannot send to the client side the full list of users that have access to a specific article that would be leaking information..  very dangerous. Hence your permissions object above should be restricted to the currently signed in user.

mleanos:
@vaucouleur I agree on the security concerns. That example was just what the data would look like.

ilanbiala:
@mleanos is this ready to be reviewed?

modules/articles/client/config/articles-admin.client.routes.js
```diff
@@ -0,0 +1,37 @@
+﻿'use strict';
+
+// Setting up route
+angular.module('articles.admin.routes').config(['$stateProvider',
+  function ($stateProvider) {
+    $stateProvider
+    .state('admin.articles', {
```
ilanbiala: indent

modules/articles/client/config/articles-admin.client.routes.js
```diff
@@ -0,0 +1,37 @@
...
+      abstract: true,
+      url: '/articles',
+      template: '<ui-view/>'
```
ilanbiala: does this not need to be '<ui-view></ui-view>'?

modules/articles/client/config/articles-admin.client.routes.js
```diff
@@ -0,0 +1,37 @@
...
+    })
+    .state('admin.articles.list', {
+      url: '',
```
ilanbiala: Why is there no url?

modules/articles/client/config/articles.client.routes.js
```diff
@@ -5,32 +5,18 @@ angular.module('articles').config(['$stateProvider',
...
-        url: '/:articleId/edit',
-        templateUrl: 'modules/articles/client/views/edit-article.client.view.html',
-        data: {
-          roles: ['user', 'admin']
-        }
-      });
+    .state('articles', {
+      abstract: true,
+      url: '/articles',
+      template: '<ui-view/>'
```
ilanbiala: same thing here?

modules/articles/client/config/articles.client.routes.js
```diff
@@ -5,32 +5,18 @@ angular.module('articles').config(['$stateProvider',
...
+    })
+    .state('articles.list', {
+      url: '',
```
ilanbiala: here?

modules/articles/client/controllers/admin.articles.client.controller.js
```diff
@@ -0,0 +1,70 @@
...
+        $scope.title = '';
+        $scope.content = '';
+      }, function (errorResponse) {
+        $scope.error = errorResponse.data.message;
+      });
+    };
+
+    // Remove existing Article
+    $scope.remove = function (article) {
+      if (confirm('Are you sure you want to delete this article?')) {
```
ilanbiala: Isn't this a browser popup?

modules/articles/client/controllers/admin.articles.client.controller.js
```diff
@@ -0,0 +1,70 @@
...
+        if (article) {
+          article.$remove();
+
+          for (var i in $scope.articles) {
```
ilanbiala: Why is this loop not a var i =, i <, i++?

modules/articles/tests/client/admin.articles.client.controller.tests.js
```diff
@@ -0,0 +1,220 @@
+﻿'use strict';
+
+(function () {
```
ilanbiala: IIFE?

modules/articles/tests/client/admin.articles.client.controller.tests.js
```diff
@@ -0,0 +1,220 @@
...
+  // Articles Admin Controller Spec
+  describe('Articles Admin Controller Tests', function () {
+    // Initialize global variables
+    var ArticlesAdminController,
+        scope,
```
ilanbiala: indentation?

modules/articles/tests/server/admin.article.server.routes.tests.js
```diff
@@ -0,0 +1,198 @@
+﻿'use strict';
+
+var should = require('should'),
+    request = require('supertest'),
```
ilanbiala: indentation?

modules/articles/client/config/articles-admin.client.routes.js
mleanos: this route is inherited from it's parent; `admin.articles` so it doesn't need a url. Essentially, it makes the *list* route the default: `/admin/articles`

mleanos:
@ilanbiala I will fix the indentation. As for your other comments, I left the implementation's the same as it currently is in the project. As for the IIFE's.. all the client side tests are implemented this way. I didn't change those. Some of these files are new, so that's why they appear to be *diffs*; rather the code was just copied over from the existing scripts.

For instance
https://github.com/meanjs/mean/blob/master/modules/articles/tests/client/articles.client.controller.tests.js
https://github.com/meanjs/mean/blob/master/modules/users/tests/client/authentication.client.controller.tests.js

https://github.com/meanjs/mean/blob/master/modules/articles/client/controllers/articles.client.controller.js
https://github.com/meanjs/mean/blob/master/modules/articles/client/config/articles.client.routes.js

I'm fine with updating this PR to implement coding best practices, where it's necessary. However, I didn't want to veer away from the current implementations for the sake of this PR.

mleanos:
@ilanbiala I'd like to get this PR back on track. I've been waiting on #874 to get finalized.

This PR still needs to be discussed, as it introduces a big change in how the Articles are implemented. I'd love your input on the logic of what I'm trying to accomplish with this.

mleanos:
@ilanbiala Any feedback here? I addressed your comments.

ilanbiala:
@mleanos this allows for admins to edit any article, right? I'm not opposed to it, we just need to clean it up a bit. @lirantal objections to it?

mleanos:
@ilanbiala Not only that, it also restricts the ability to create/edit Articles to **only** Admins; although, I think we could intriduce a new role for this. So essentially *normal* users can only view Articles. This is a big difference in hiw its done now.

lirantal:
@mleanos why are we adding a restriction though for users? what's behind that?

mleanos:
@lirantal This stemmed from a conversation that @rhutchison & I had. We were discussing the ambiguity of what Articles really were; are they user *Posts* or actual Articles like in most CMS?

We determined that it makes more sense that they should be treated more like how CMS implements this type of feature. Where you have a specific role that manages the Articles, and users can view them.

From @rhutchison toward the top of this discussion: 
> Maybe we just never understood what articles is trying to be, but by it's name, I see it being similar to how it would be implemented in any CMS. We have been open to discussion about what articles in this project really means, and I was hoping that we would all be able to come to a consensus and build it out further.

> I don't think it's a great idea to build out a shared view and then have ng-show/hide or ng-if statements everywhere. Eventually it will become unmanageable and not make much sense to have in the same view.

> Things we have discussed are active/expiration dates, rich text editing, additional roles like publisher/editor, etc.

lirantal:
@mleanos I agree. In this case though we need to have with this PR an 'article' related role already assigned to the user so if we merge this PR then it won't break the functionality of users being unable to create/edit articles.

mleanos:
@lirantal I'm not quite following what you mean. Can you elaborate? 

Are you suggesting that we accompany this PR with a script that adds this new *article* related role to every user? Which a user of this project can run, if they wish.

lirantal:
With this PR the status is this:
* Only 'admin' role can create/edit/delete articles
* All users can view articles
Right?

Before this PR, all users and admins were able to create/edit/delete articles content.
So I'm saying, let's maintain the functionality of allowing users to create articles by creating an 'article' role and assigning it to them by default when they get created in the user model schema. Having a script to also grant this to existing users in other people's deployments sounds ok to me and can be added in the scripts/ folder

ilanbiala:
@lirantal I'm actually a fan of having an admin role that is the only role allowed to create/edit/delete articles. While it is a breaking change, we can introduce it in 0.5.0.

lirantal:
If you're all agree then fine, but sounds to me like we are breaking functionality that was working before with regards to permissions. With that said I'm ok to continue because the articles content is just provided as an example for modules and not part of core project for other users to adopt probably so not the end of the world.

ilanbiala:
@mleanos if we get this rebased, can this be merged in for 0.5.0 or is there something that needs to be merged in before this.

codydaig:
@ilanbiala The initial plan was to get #874 merged in before this.

mleanos:
@ilanbiala I think it makes sense to have this merged in with 0.5.0. IMO, Introducing a breaking change is usually better to be done with a major release.

ilanbiala:
Yes that's the plan.

ilanbiala:
@mleanos now that we are on track for 0.5.0, can you look into rebasing this and preparing it for 0.5.0?

mleanos:
@ilanbiala Sure thing. Was on my todo list for the week :) 

Once I do that, can you give it a once over again. I can't remember if we made any concrete decisions on the changes with the Roles/Permissions.

ilanbiala:
Yeah once we get it rebased I'll take another look to just re-review it.

mleanos:
@ilanbiala @codydaig I'm actually gonna wait until https://github.com/meanjs/mean/pull/874 gets merged. There will be some conflicts I'll need to resolve.

mleanos:
@trainerbill What was decided on how to handle a 400 (Bad Request) with this new UI route implementation?

For instance, when an invalid `:articleId` is provided, this will receive an error from the api call & not get handled on the client. Same for when the article isn't found.

mleanos:
@ilanbiala I'd like to get moving on this. If #874 doesn't have any movement soon, then I'll just start back up on this. One of these PR's is going to have major conflicts; I was hoping the Article Refactor would go first since there was a lot of work put into it.

ilanbiala:
SGTM.

codydaig:
@mleanos Now that the article refactor is merged in, what's the status / plan for this PR?

mleanos:
I'll get this ready next week. #1149 should go first, so the bug is fixed. 

The concept here is pretty well defined. However, we'll still need to discuss the details. For instance, how to redefine the roles. 

1) What new roles should we introduce?
2) As @lirantal pointed out, we'll be introducing a breaking change. What's our strategy for making it an easy transition?
3) If we allow roles other than 'admin' to manage, should we keep the create/edit/remove functionality in the main Article module, or do we allow these new roles to access the Articles admin module?

I'll be out of town over the next few days, but as soon as I get back I'll get right on this.

trainerbill:
Just as a heads up the JWT PR will not be compatible with this.  Someone need to decide which one will go first.

mleanos:
@trainerbill, can you explain why?

trainerbill:
@mleanos I guess I shouldn't have said not compatible.  If this PR goes first it will make the JWT PR larger, so someone needs to prioritize what goes in first.  The admin routes change a little bit in that PR.  In my mind JWT Auth would be given a higher priority and then this PR switched to conform to it, but thats my biased opinion :)

ilanbiala:
I'd actually prefer to get older stuff in first so the rebases don't get harder and so we don't lose track of it. Is there any way we can get this merged in quickly and easily for whatever JWT implementation we use?

trainerbill:
@ilanbiala I am fine with that, it will just make the JWT PR larger.  Can we merge then so I can JWTize this?

codydaig:
@trainerbill 
@mleanos Still needs to rebase this after the refactor. Once he does that and it goes over a final review, we can merge it. 
@ilanbiala :+1:

mleanos:
I'm planning on getting this done in the next few days; assuming nothing that was introduced since I last worked on it, doesn't throw any wrenches in the mix.

@trainerbill I really don't understand how this affects the JWT PR. Why would our Authentication method (whatever it may be) have any consequences on this implementation? We'll still have roles. And actually, this PR would make the JWT simpler for you since we would no longer have any need to reference the User's `_id` field on the client-side.

If you still feel this PR affects yours, can you provide some examples or explanations of what you mean?

trainerbill:
@mleanos.  It just makes the JWT PR larger as I will have to redo the routes file.  Everything else would likely stay the same.

mleanos:
@trainerbill Why do you have to mess with the routes file, for Authentication changes?

trainerbill:
@mleanos Yeah you have to call passport.verify or something like that.  Also the admin api routes change a little.  /api/admin/user /api/admin/article etc.

mleanos:
@trainerbill I still don't know what you mean. What do you need to call passport.verify with? And why do the api routes change?

mleanos:
@trainerbill I'm on Gitter. Let's chat there. I'm working on the rebase & merge conflicts for this PR right now.

mleanos:
@codydaig @ilanbiala @lirantal @rhutchison @trainerbill @vaucouleur I have rebased & resolved **all** the conflicts. 'Twas a doozey this one :) 

I'll recap what this PR is accomplishing here in a bit. My brain is frazzled from those merge conflict resolutions, and retro-fitting the tests.

codydaig:
@mleanos : I'll review this in the morning but are there any tests that can be added to prevent coverage from dropping?

mleanos:
I'm planning on adding some additional tests to this PR. Can you review in the meantime? There's still some additional considerations (mentioned above) that need addressing. The main thing is that we should add a new Role, that has permission to create/edit/remove Articles & then add this to all User's by default; @lirantal suggested this here.. https://github.com/meanjs/mean/pull/807#issuecomment-141372350 & makes a very good point.

trainerbill:
I am beginning to wonder what the benefit of having an articles admin really is.  Seems to me like it is just more stuff we have to maintain.  No one really uses the module anyway, it probably should just be renamed CRUD and be an example module for CRUD operations.  In that case I could see adding a very basic admin implementation to show users how to extend the admin routes and such, but a fullscale article admin that we have to maintain, write tests for, and make sure future PR's don't break?  I just don't see the need.  Do we really want to maintain this going forward?  We can barely maintain what we have and introduce new core features.  Everything like this that goes in increases the size of other PR's, which in turn increases amount of time it takes them to be merged.  I bet if you asked the community 98% would say they don't use the module and just reference it when building their own, because this is a skeleton module an that is all it should be.

No disrespect @mleanos, I know you have put a lot of work into this PR.  My 2 cents before this merges and I am probably in the minority here.

mleanos:
@trainerbill No worries. It's good constructive criticism. However, I don't share your idea that most users of this framework just scrap the Articles module. I think most expand on it. I think it adds value to the framework, so that we can show the implementation of an Admin feature for a module other than *users*. I don't see it being a big deal to maintain either. Once the tests are written & the implementation is there, then it's just a matter of fixing any bugs, or adding enhancements that would add more value.

This is a good sum up of the intention of this feature.. https://github.com/meanjs/mean/pull/807#issuecomment-134503192

IMO, it's a low impact feature that can add tremendous value to a high percentage of use cases that we're seeing on the web these days. Very CMS/Social like.

I'm all ears though, if others would like to join in on this discussion. If the community doesn't feel like this adds value, then I'll support that. The time I spent on it wasn't a waste at all; the experience & education was way worth it :)

trainerbill:
**I think it adds value to the framework, so that we can show the implementation of an Admin feature for a module other than users**  I agree with this, but it belongs in a separate module repo

I don't mind articles having an admin section, if articles was a module repo outside of the core framework.  Take my Nested Views PR for example.  Now I have to redo 4 more routes, at least 4 more views all for something that I think is rarely used.

codydaig:
I personally am in favor of this PR. The admin module is an example of a full CRUd module. 

The discussion of making it a separate repo or not is not relevant to this PR. 

Having an example of the admin in a full CRUD module is beneficial for a lot of people just getting started with the framework.

ilanbiala:
Let's definitely get this in. @mleanos can you add some tests and tidy it up a bit for a final review?

mleanos:
@ilanbiala @codydaig I'll add some tests. See this comment by @lirantal https://github.com/meanjs/mean/pull/807#issuecomment-141372350

Is this a concern that we need to address?

codydaig:
For easy reference: @lirantal said:
> Before this PR, all users and admins were able to create/edit/delete articles content.
So I'm saying, let's maintain the functionality of allowing users to create articles by creating an 'article' role and assigning it to them by default when they get created in the user model schema. Having a script to also grant this to existing users in other people's deployments sounds ok to me and can be added in the scripts/ folder

I'm on board of the the idea of an 'article' role. However, I don't know how necessary the script would be for backwards compatibility. 0.4 -> 0.5 is another big leap with a lot of changes, and I don't know how much we need to worry about that specific of backwards compatibility.

ilanbiala:
I'm not really sure it's necessary. 0.4 -> 0.5 can include breaking changes, and I think we're doing something smart by limiting permissions to admins.

ilanbiala:
@mleanos we need to rebase this and finish up the testing before we can merge, but I think we should be able to do that by the weekend.

mleanos:
@ilanbiala I've been pretty busy the last couple of days. I'll do my best to get this done by Friday morning.

@codydaig @ilanbiala So cool with just allowing the `admin` Role to create/edit/delete, rather than adding a new Role?

mleanos:
@codydaig @ilanbiala I've added tests, and moved a couple to the Admin Articles tests. The coverage didn't increase, but stayed the same; that feels like a win with this one :) Let me know how everything looks.

modules/articles/client/config/articles-admin.client.config.js
```diff
@@ -0,0 +1,11 @@
+﻿'use strict';
```
codydaig: Why no IFFE in this file?

modules/articles/tests/server/admin.article.server.routes.tests.js
```diff
@@ -0,0 +1,278 @@
+﻿'use strict';
```
codydaig: Why no IIFE?

codydaig:
Initial skim through this looks good. This does need to get rebased. I would like to clone this down later and test and give a more detailed review.

mleanos:
@codydaig I'll go through and make sure I've updated all the new files to use the Styleguide. I missed that after my rebase, after the Styleguide was implemented.

mleanos:
@codydaig I've rebased, and added the IIFE to that file. The other file you mentioned was a server-side file. AFAICS we're not using IIFE's on the server.

We'll see how this build goes, but in the meantime this should be ready for you to pull down & review.

modules/articles/tests/server/admin.article.server.routes.tests.js
ilanbiala: We don't need one for the server-side

modules/articles/tests/server/admin.article.server.routes.tests.js
```diff
@@ -0,0 +1,278 @@
...
+              .expect(200)
+              .end(function (articleUpdateErr, articleUpdateRes) {
+                // Handle article update error
+                if (articleUpdateErr) {
+                  return done(articleUpdateErr);
+                }
+
+                // Set assertions
+                (articleUpdateRes.body._id).should.equal(articleSaveRes.body._id);
+                (articleUpdateRes.body.title).should.match('WHY YOU GOTTA BE SO MEAN?');
```
ilanbiala: why match?

modules/articles/tests/server/admin.article.server.routes.tests.js
```diff
@@ -0,0 +1,278 @@
...
+                (articleInfoRes.body.isCurrentUserOwner).should.equal(true);
+
+                // Call the assertion callback
+                done();
+              });
+          });
+      });
+  });
+
+  afterEach(function (done) {
```
ilanbiala: Is there no need for an after to kill the app?

ilanbiala:
@mleanos looks pretty good to me. Only organizational thing, can you try to separate the admin tests into a separate describe and have another one for standard user tests so it's more grouped together? @codydaig wanna take a final look?

modules/articles/tests/server/admin.article.server.routes.tests.js
mleanos: I merely ported some of these tests over from the Articles tests. So we were probably already using `match` here. I could always change it.

modules/articles/tests/server/admin.article.server.routes.tests.js
mleanos: I don't think so. And we're not doing that anywhere else.

mleanos:
@ilanbiala I've already separated the normal User tests from the Admin tests. Are there some that I missed, or still have mixed in with the others?

ilanbiala:
Maybe it was fine actually, I think I just mixed up a few and so I thought they weren't separated.

mleanos:
@ilanbiala What's the review status of this? Did you have a chance to take it for a test drive?

trendzetter:
Seems to work. Is this a proposal to remove the user submitted article in favor of admins generating content and users consuming? You could demonstrate both and the use of roles by creating 2 types of cruds (articles & tags?). Of course this would be a more dramatic change.

mleanos:
@trendzetter Yes. I think it makes sense to restrict the creation/editing/deleting of Articles to Admin User's.

This is a pretty big change, and I don't think introducing "Tags" would make sense at this point.

I'm still not sure how the community feels about this change. I wonder what the typical use case for the Articles module really is. Allowing User's to create/manage Articles seems like a very narrow use case for most applications. 

My hunch is that more often than not, our users scrap the Article's feature or drastically change it's behavior. Given this assumption, I think the changes introduced here would be more in line with a standard use case for most applications.

@codydaig @ilanbiala @lirantal Perhaps we should keep this as a lower priority until we're closer to releasing 0.5.0, and then re-evaluate it's relevance at that time. WDYT?

ilanbiala:
SGTM, but maybe just rebase to avoid further conflicts and possibly redoing the PR from scratch if it gets too far from master?

trendzetter:
@mleanos I think it looks good. The practical functionality can be either, it doesn't really matter that much since none of the functionality should be used as is but rather as a starting point and those examples can be used for either user generated content as well as an application with administrators/editors managing the content.

lirantal:
@mleanos do you want to kick this in please?

mleanos:
@lirantal I would like to get this merged. However, I was hoping for more feedback & a few LGTM's since this is a pretty big change on how the Articles module works.

If you're fine with this merging, I'll test once again & review for any changes that might conflict.

codydaig:
I would like to see this get added!

lirantal:
@mleanos sure you're welcome to fix, test, and merge it.

ilanbiala:
@mleanos reviewed before and LGTM, just rebase and it should be fine, but I'll check again after the rebase to make sure.

mleanos:
@meanjs/contributors Thanks! I'll rebase & resolve the conflicts. These [changes](https://github.com/meanjs/mean/commit/a7370a780853cb494d9c1ab17cfe4ecc2f5611cc) I made to the Articles client controller & service are most likely the culprit.

mleanos:
@meanjs/contributors Rebased & resolved merge conflicts.

@ilanbiala Can you review once more before I merge?

ilanbiala:
LGTM.

mleanos:
Big change here. We probably should update our documentation to clearly explain these changes. I'll look at that, but may need someone else to tackle the documentation PR.

PR #812, opened 2015-08-16
Fixed bug with Socket IO session

mleanos:
New enhancements to the Express session, added "sessionId" as the new
default session key. Previously, the express session was using the
default "connect.sid" key. This caused the Socket configuration to be
unable to find the session id, thus causing issues with the Socket. One
such issue, was that the same Socket would be used for all users
connected to the server.

This also adds a bit more error handling of the Socket server
configuration.

### DIFF
--- a/config/lib/socket.io.js
+++ b/config/lib/socket.io.js
@@ -71,10 +71,15 @@ module.exports = function (app, db) {
     // Use the 'cookie-parser' module to parse the request cookies
     cookieParser(config.sessionSecret)(socket.request, {}, function (err) {
       // Get the session id from the request cookies
-      var sessionId = socket.request.signedCookies['connect.sid'];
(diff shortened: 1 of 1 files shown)


lirantal:
LGTM

mleanos:
Re-worked the error handling. Using the `return next()` pattern for handling the errors. This makes the code cleaner, and easier to read.

mleanos:
Attempted to fix the spacing issues.. I removed automatic indentation on new lines. @rhutchison How does it look to you now?

codydaig:
LGTM

Tested, and verified that it works.

rhutchison:
LGTM - thanks

lirantal:
Thanks @mleanos and @rhutchison @codydaig for reviewing!

PR #813, opened 2015-08-16
Correct <br /> Tags and set Title

Meistercoach83:


lirantal:
LGTM

lirantal:
Indentation should be 2 spaces

rhutchison:
@Meistercoach83 Why do we need to correct <br> tags to be compatible with XHTML? <br> is sufficient.

lirantal:
@rhutchison you're right, it shouldn't really matter because it's only relevant for XHTML but this template is for e-mails and not for browsers to maybe it's just safer to use XHTML notations for e-mail clients?

ilanbiala:
What's the difference between this and #814?

lirantal:
@ilanbiala look closely on the filename in the view - they are different email templates to be sent out.

codydaig:
@Meistercoach83 Can you put the indentation back to 2 spaces?

ilanbiala:
@lirantal @Meistercoach83 why don't we just have 1 PR for both files. It's the same situation.

rhutchison:
Can we get #813 and #814 in one PR

lirantal:
@rhutchison you're right but we already have this here so whatever @Meistercoach83 decides, doesn't matter that much to me.

@Meistercoach83 please squash the commits to 1 and address the indentation issue to 2 spaces so we can merge this in.

PR #814, opened 2015-08-16
Correct <br /> Tags and set Title

Meistercoach83:


lirantal:
@Meistercoach83 how about keeping the indentation to 2 spaces?

rhutchison:
@Meistercoach83 Why do we need to correct <br> tags to be compatible with XHTML? <br> is sufficient.

modules/users/server/templates/reset-password-confirm-email.server.view.html
```diff
@@ -1,16 +1,17 @@
 <!DOCTYPE html>
 <html lang="en" xmlns="http://www.w3.org/1999/xhtml">
```
ilanbiala: Why is the xhtml here when we are using HTML5 doctype?

pgrodrigues:
@ilanbiala I think it's better to use XHTML for email templates because of least inconsistencies, I'm using this [one](https://github.com/seanpowell/Email-Boilerplate/blob/master/email.html) and it provides some info on that topic (in top comments). However if the user wants to use the new elements from HTML5 in his emails then there should be no problem in using HTML5 I believe.

pgrodrigues:
@ilanbiala I removed my last comment because I misunderstood the comment :P

lirantal:
@Meistercoach83 can you quickly google up if email clients really do require an XHTML document? I'm pretty sure they do but it would be good if we can have some kind of assurance.

@rhutchison @ilanbiala I still think it is better to make the email templates XHTML compatible rather than HTML5 as they are actually used in order to be sent to email clients, and email clients != browsers.

Meistercoach83:
[Which doctype should I use in HTML email?] (https://www.campaignmonitor.com/blog/post/3317/correct-doctype-to-use-in-html-email/)

"If you’re going to use CSS3, XHTML 1.0 Transitional is the doctype for you (until HTML5 becomes standard, anyway)."

"XHTML for validation"
The other thing is, if you put a premium on validating your code (as well as you can when coding for email, anyway), then we recommend the XHTML 1.0 Transitional doctype. Naturally, not having a doctype will throw errors and and the HTML5 doctype is still marked as experimental. Mike Kleiman has a light summary on standards vs. quirks mode and email clients, should you wish to get all semantic.

Well, this is by no means an exhaustive test and we’re more than happy to test out specific elements that haven’t been covered here. But the moral of the story is that if you are keen on retaining the default margins on elements (and like CSS3), use the XHTML 1.0 Transitional doctype. That said, there are loads of email clients that arbitrarily add margins and padding to elements, so be sure to check our forums before you pin a rendering issue on the doctype used in your email.

Our friends at Email on Acid have been doing some pretty comprehensive doctype testing themselves. Well worth a read!

** What doctype should I use?
After all this deliberation, our recommendation is to use XHTML 1.0 Transitional doctype:" **

lirantal:
I agree with keeping the XHTML standards and doctype for the e-mails being sent.
Let's go with that, and in the future we can update. These are anyway not that very rich e-mail templates so I don't see this becoming an issue anytime soon.

lirantal:
@Meistercoach83 can you update the indentation there back to 2 spaces?

PR #816, opened 2015-08-17
Configurable SSL mode options

lirantal:
adding support for a configurable ssl-mode and and ssl cert/key files in the environment configuration


codydaig:
LGTM

lirantal:
Thanks.

Issue #817, opened 2015-08-17
File naming conventions: Users (Admin) module versus Article module


vaucouleur:
There are two minor (harmless) inconsistencies in the naming of the view files : 

- either object-verb ("user-edit") versus verb-object ("edit-article")
- the view page being either object alone ("user") versus view-object ("view-article")

Users (admin) pages:

* user-edit.client.view.html
* user.client.view.html
* user-list.client.view.html

versus Article pages:

* edit-article.client.view.html
* view-article.client.view.html
* list-articles.client.view.html

lirantal:
@vaucouleur right.
you're welcome to suggest a convention and a PR, once we've made a decision to resolve it.

codydaig:
I personally like the structure of the article page names.

mleanos:
Same here. I like prefer the articles view naming convention.

vaucouleur:
Ok, thanks for the feedback.

There is also a simple, yet important, argument in favor of the Article module naming convention: there is much more generated "Article" (or "Article-like") modules out-there, than the relatively recent "user admin" pages. The changes should have relatively less impact in terms of merging.

@rhutchison Is renaming the files using the Article module convention OK with you ?

trainerbill:
I like user.list.client.html or article.list.client.html.  I can include the article pieces in my refactor once its concluded:

https://github.com/meanjs/mean/pull/800

lirantal:
I thought this whole question was about whether the convention is to use 'user' as a first word in the files or the actual action 'list', so basically it's 'user.list' vs 'list.user'

Whether we use dots or dashes doesn't really matter to me, but I tend towards dots as I see it also used in other frameworks.

So I think we're quite in agreement here?
@ilanbiala or @rhutchison ?

trainerbill:
I personally like list.user.client.html.  Person is going to know they are in the users module so its important to put the action first.  IMO

mleanos:
:+1: I agree that the *action* should be first in the naming convention... and I prefer dots as well.

vaucouleur:
I messed up the initial PR while trying to do some git surgery (conflict resolution). Sorry about that. I will re-submit a new one.

lirantal:
Sure

vaucouleur:
I hope I did not miss anything. 

I only renamed one controller, one route file, and 3 html view pages.

The other files already followed the verb-object.client convention.

PR #823, opened 2015-08-19
Renamed users (admin) html files to follow the conventions of the Article module

vaucouleur:
Renamed users (admin) files to follow the naming convention of the Article module:

- user-edit.client.view.html -> edit-user.client.view.html
- user-list.client.view.html -> list-users.client.view.html
- user.client.view.html -> view-user.client.view.html

Discussion: https://github.com/meanjs/mean/issues/817

### DIFF
--- a/modules/core/client/views/home.client.view.html
+++ b/modules/core/client/views/home.client.view.html
@@ -24,7 +24,7 @@
     <ul>
       <li>
         Click
-        <em>Signup</em> to get started.
+        <em>Sign Up</em> to get started.
       </li>
       <li>
         Configure your app to work with your social accounts, by editing the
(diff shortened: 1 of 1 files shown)


codydaig:
LGTM

mleanos:
@vaucouleur What about using the dot notation? for instance, `edit.user.client.view.html`

trainerbill:
@mleanos +1  That is what I would like to see and I think @lirantal agrees.

vaucouleur:
@mleanos @trainerbill  Dots or dashes, any choice is fine with me. I just pointed out that it seemed wise to be consistent across modules. The Article module uses dashes, and there is a lot of "generated modules" out there that use the Article convention. If we do move to dots, we should also rename the Article html templates. Is that what we want ? If so, I can amend the PR.

trainerbill:
@vaucouleur You have my vote on action.module.client.view.html

but I am not a contributor.

codydaig:
@vaucouleur Sounds like the consensus is action.module.client.view.html (edit.user.client.view.html). If you could update and squash your commits, it should be good to go. Any objections @lirantal?

lirantal:
No objections, I'm all up for the dot notation.

vaucouleur:
The commits are squashed, and both the user admin and the article modules now follows the convention:
verb.object.client.view.html

mleanos:
LGTM :+1: Tested and confirmed this is working.

Thanks @vaucouleur

vaucouleur:
@mleanos Thanks to you for double checking this !

rhutchison:
Not to throw a wrench in to this, but I preferred the naming convention of users admin to following the naming convention of the state.

mleanos:
@rhutchison I see your point, but I think for views it makes more sense to have the *verb* first in the naming convention. The reason it makes sense to have the states named with their module/feature first, is of the hierarchical nature of states. Views don't follow that type of pattern, IMO.

lirantal:
right, I'm still ok with the naming convention proposed here unless @rhutchison want to suggest some higher level more consistent naming convention across the project.

rhutchison:
I am fine with flipping the words, but my point about inconsistency remains.  This PR was updated to implement [dot notation](https://github.com/meanjs/mean/pull/823#issuecomment-132626564), where-as the rest of the project is using dash. If you are going to switch to dot, then it should be across the project.

vaucouleur:
@rhutchison This is a good point, consistency was my main initial concern. Ideally the file naming convention should be consistent across the project. I would be happy to amend the PR and apply the same naming convention to all html pages, if there is a consensus about this.

Something else: some pages have 2 words for the object name, for example "profile picture". Those words can be either separated by a dash (to denote "a close binding between those words"),  or to also use a dot.

- change.profile-picture.client.controller.js

... versus ...

- change.profile.picture.client.controller.js

My preference would be to keep a dash in this case between "profile" and "picture". But again this is a minor point, any of those choice is probably good as long as it is used consistently.

lirantal:
I don't have a strong argument to support either so I'm fairly native.
Though is there a common style for that which other frameworks/projects adopt? I'm sure we're not inventing the wheel here.

rhutchison:
https://github.com/johnpapa/angular-styleguide#style-y121

`avenger-profile.directive.js`

lirantal:
I'm ok with having this:
```view-user.client.view.html```
and this to complement it
``` change-profile-picture.client.controller.js```

So this stays consistent where the first part of the filename (usually the verb/action) is formatted with a dash, and then the following filename info is with dots.

Sounds good?

rhutchison:
sounds good, @vaucouleur what do you say?

vaucouleur:
@lirantal  @rhutchison That's just fine. I will fix the PR with this pattern, and I will extend the PR to all html pages. Thanks for your feedback.

lirantal:
Great.

mleanos:
This is perfect. Thanks for getting this sorted out guys.

PR #824, opened 2015-08-19
Fixed issue where if local.js exists then grunt test will run on that…

lirantal:
… environment config and possibly delete collections

We only extend the config object with the local.js custom/local environment if we are on production or development environment. If test environment is used we don't merge it with local.js to avoid running test suites on a prod/dev environment (which delete records and make modifications)

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -159,9 +159,14 @@ var initGlobalConfig = function () {
   var environmentConfig = require(path.join(process.cwd(), 'config/env/', process.env.NODE_ENV)) || {};
 
   // Merge config files
-  var envConf = _.merge(defaultConfig, environmentConfig);
+  var config = _.merge(defaultConfig, environmentConfig);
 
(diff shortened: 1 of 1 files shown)


codydaig:
@lirantal In this case do we still want/need the double variable declaration?

```
// Don't need the envConf var anymore. 
// var envConf = _.merge(defaultConfig, environmentConfig);
var config = _.merge(defaultConfig, environmentConfig);

  config = _.merge(conf, (fs.existsSync(path.join(process.cwd(), 'config/env/local.js')) && require(path.join(process.cwd(), 'config/env/local.js'))) || {});  
  }
```

lirantal:
Thanks @codydaig , I updated it.

codydaig:
@lirantal You also have to change envConf to config on line 168. Sorry, I didn't make that very clear.

lirantal:
Sure, my bad. Long day this is :)

Just proves how much more tests we really need on the framework itself to make sure we're not messing things up.

codydaig:
@lirantal LGTM! Thanks! :-D

lirantal:
Thank *you*

PR #826, opened 2015-08-20
[test] Allow an email with sub domains

codydaig:
@lirantal Great job on the email tests you added. I took that and added it against my group's project and they all passed, but when I added the following it failed. It currently passes in the MEAN.js project, but I think it's worth checking for in case we switch validators down the road.


mleanos:
LGTM

lirantal:
Thanks @codydaig, glad to see that the tests actually caught something for you!

P.S. I'll merge but I think I'm seeing still a problem there with the tests in general (mine included) which need to be addressed

PR #830, opened 2015-08-20
[fix] Was storing a 6 char password in plain text [fixes #829]

### LINKED ISSUE #829: Password is being stored in clear text if length is 6
Password is being stored in clear text if length is 6, just an equal sign is required here, or make sure min length for password is 7

    /**
     * Hook a pre save method to hash the password
     */
    UserSchema.pre('save', function (next) {
      if (this.password && this.isModified('password') && this.password.length >6) {
        this.salt = crypto.randomBytes(16).toString('base64');
        this.password = this.hashPassword(this.password);
      }
    
      next();
    });

codydaig:
Fixes #829.

If the password was exactly 6 characters long, the password would not get hashed and was being stored in the database as plain text. I added a couple tests to ensure the password wasn't being stored in plain text as well as fixed the logic that was skipping over the password hashing. 

Thanks @sparshy! :-D

### DIFF
--- a/modules/users/server/models/user.server.model.js
+++ b/modules/users/server/models/user.server.model.js
@@ -107,7 +107,7 @@ var UserSchema = new Schema({
  * Hook a pre save method to hash the password
  */
 UserSchema.pre('save', function (next) {
-  if (this.password && this.isModified('password') && this.password.length > 6) {
+  if (this.password && this.isModified('password') && this.password.length >= 6) {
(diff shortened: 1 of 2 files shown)


sparshy:
Thank u guys for maintaining this awesome project

mleanos:
LGTM. Good catch @sparshy

rhutchison:
LGTM

lirantal:
Thanks guys, good catch indeed @sparshy

PR #837, opened 2015-08-22
Force username & email to lowerCase PLUS Remove sensitive data before…

almegdad:
… return authenticated user in password reset

### DIFF
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -86,14 +86,14 @@ describe('User Model Unit Tests:', function () {
     });
 
     it('should be able to update an existing user with valid roles without problems', function (done) {
-      var _user = new User(user);
+      var _user1 = new User(user1);
 
-      _user.save(function (err) {
+      _user1.save(function (err) {
         should.not.exist(err);
-        _user.roles = ['user', 'admin'];
-        _user.save(function (err) {
+        _user1.roles = ['user', 'admin'];
+        _user1.save(function (err) {
           should.not.exist(err);
-          _user.remove(function (err) {
+          _user1.remove(function (err) {
             should.not.exist(err);
             done();
           });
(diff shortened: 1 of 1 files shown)


lirantal:
@rhutchison @codydaig @ilanbiala @mleanos any thoughts?

modules/users/client/views/authentication/signin.client.view.html
```diff
@@ -6,7 +6,7 @@ <h3 class="col-md-12 text-center">Or with your account</h3>
         <div class="form-group">
           <label for="username">Username</label>
           <input type="text" id="username" name="username" class="form-control"
-               data-ng-model="credentials.username" placeholder="Username">
+               data-ng-model="credentials.username" data-ng-change="credentials.username=credentials.username.toLowerCase();" placeholder="Username">
```
lirantal: We seem to have decided not to use the `data-` prefix for angular attributes so please change this through-out the PR to just `ng-change` and so on..

rhutchison:
can we use css `text-transform: lowercase;` instead of ng-change, and then on save (client or server) force to lower?

mleanos:
I agree with @rhutchison that this should be done with the save method, either on the client or server side. I'm not a fan of using `ng-change` for these types of scenarios. 

I don't see a need to transform with css. The real need here is to change the username to lowercase, *sometime*, before authentication; in order to decrease the chance of user error. 

However, if the view must be updated to reflect the lowercase change, then I'd say that's an argument to do the conversion, in the client controller `save` method; thus, updating the view.

mleanos:
After looking at these changes a bit further, I think the lowercase conversion should be done at the very front-end with the user input; most likely in the client controller method that initiates any API calls. If these changes are made in the middleware, they can get lost & the SOC gets weaker.

To ensure the username & email always get converted to lowercase, the pre-`save` hook can be used on the User model.

rhutchison:
>I don't see a need to transform with css. The real need here is to change the username to lowercase, sometime, before authentication; in order to decrease the chance of user error.

>However, if the view must be updated to reflect the lowercase change, then I'd say that's an argument to do the conversion, in the client controller save method; thus, updating the view.

Why wouldn't you just have text-transform?  the user doesn't need to know when it's transformed, just that it is being transformed.

mleanos:
@rhutchison I don't like using css as a means to transform data based on business logic/needs. The client/server controller would be sufficient in most cases. If someone wants to update the view to show the transformation, then adding the css `text-transform` would be an enhancement, IMO.

ilanbiala:
.toLowerCase is good, but I'd do it in the JS and not in the HTML.

simison:
Those ng-change bits should be removed; after that this LGTM and pretty much how I'm doing it already at my copy.

+1 @mleanos - CSS/angular frontend tweaking seems hacky.

almegdad:
reopened #866

PR #839, opened 2015-08-22
[fix] Fixed Facebook Auth [Closes #697]

### LINKED ISSUE #697: Facebook Login is not Working
The Facebook Login is not working and it does not throw any error in the console or web console.
The Facebook redirect works, but then in the application nothing is happening.
I have configured the App_id and secret. What's missing?

codydaig:
Per the Facebook Docs: https://developers.facebook.com/docs/graph-api/reference/v2.3/user

There is no `emails` field, it is only `email`

Closes #697

### DIFF
--- a/modules/users/server/config/strategies/facebook.js
+++ b/modules/users/server/config/strategies/facebook.js
@@ -13,7 +13,7 @@ module.exports = function (config) {
       clientID: config.facebook.clientID,
       clientSecret: config.facebook.clientSecret,
       callbackURL: config.facebook.callbackURL,
-      profileFields: ['id', 'name', 'displayName', 'emails', 'photos'],
+      profileFields: ['id', 'name', 'displayName', 'email', 'photos'],
       passReqToCallback: true
     },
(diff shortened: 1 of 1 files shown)


lirantal:
Thanks @codydaig 
Would be helpful if someone can actually test it to confirm though I think going by the docs is also ok for now.

codydaig:
@rhutchison @trainerbill @mleanos 

I made this change on mean.codydaig.me and set-up a facebook app id and secret. However, upon login, I'm getting a not-found error. Any suggestions on where to start looking?

codydaig:
@lirantal I'm testing this. Don't merge this in.

rhutchison:
This is not a bug/fix, passport-facebook takes care of an [translation](https://github.com/jaredhanson/passport-facebook/blob/master/lib/strategy.js#L151-L154) ([map](https://github.com/jaredhanson/passport-facebook/blob/master/lib/strategy.js#L220)) before making a request.

codydaig:
@rhutchison I'll close for now. Thanks!

PR #842, opened 2015-08-23
Facebook authentication

### LINKED ISSUE #202: passport facebook (facebook api v2 & email permission disallowed)
Facebook API v2 no longer returns username, hence on FacebookStrategy callback we'll get undefined for profile.username.
Currently I'm using the below in `config/strategies/facebook.js` as my workaround. 
line 32: `username: profile.username || profile.id `

Another thing is when user chose not to allow the access of email from Facebook there'll be a small error which causes the app to skip saving/updating the user.
My workaround in `config/strategies/facebook.js` 
line 31: `email: (profile.emails)? profile.emails[0].value: ''`

### LINKED ISSUE #697: Facebook Login is not Working

rhutchison:
Closes #202, and closes #697 - if user does not authorize email scope, email will not be mapped.  Username will be generated from first initial of first name and last name.

.jshint latedef set to nofunc.

### DIFF
--- a/modules/users/server/config/strategies/facebook.js
+++ b/modules/users/server/config/strategies/facebook.js
@@ -27,7 +27,8 @@ module.exports = function (config) {
         firstName: profile.name.givenName,
         lastName: profile.name.familyName,
         displayName: profile.displayName,
-        email: profile.emails[0].value,
+        email: profile.emails ? profile.emails[0].value : undefined,
+        username: profile.username || generateUsername(profile),
         profileImageURL: (profile.id) ? '//graph.facebook.com/' + profile.id + '/picture?type=large' : undefined,
         provider: 'facebook',
         providerIdentifierField: 'id',
@@ -36,6 +37,18 @@ module.exports = function (config) {
 
       // Save the user OAuth profile
       users.saveOAuthUserProfile(req, providerUserProfile, done);
+
+      function generateUsername(profile) {
+        var username = '';
+
+        if (profile.emails) {
+          username = profile.emails[0].value.split('@')[0];
+        } else if (profile.name) {
+          username = profile.name.givenName[0] + profile.name.familyName;
+        }
+
+        return username.toLowerCase() || undefined;
+      }
     }
   ));
 };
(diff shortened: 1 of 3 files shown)


rhutchison:
I have tested these changes but need additional eyes.  If anyone is using facebook strategy please weigh in.

@simison @mleanos @codydaig @lirantal @ilanbiala @jcaspain @benmartin

modules/users/server/config/strategies/facebook.js
```diff
@@ -36,6 +37,18 @@ module.exports = function (config) {
...
+      function generateUsername(profile) {
+        var username = '';
+
+        if (profile.emails) {
+          username = profile.emails[0].value.split('@')[0];
+        } else if (profile.name) {
+          username = profile.name.givenName[0] + profile.name.familyName;
+        }
+
+        return username.toLowerCase() || undefined;
```
lirantal: What happens if on 45 or 47 the username field:
1. does not exist? meaning it's empty for some reason...
2. does not fit to go through validation? for example, it's length is too short/long or uses special UTF8 chars
3. already matches an existing username and then we'll fail on duplicate usernames

If we end up returning undefined, how would the call later to saveOauthUserProfile behave?

lirantal: Also, take a look at https://github.com/meanjs/mean/blob/3064b2882048beebf5ff6fc9c3f9599a5650c764/modules/users/server/controllers/users/users.authentication.server.controller.js#L155 looks like it's already trying to figure out a username for the user.

rhutchison: I saw this, but it does not handle it as nicely.  If there is no username, the app will not create a user, takes you to the login page.  Email is not required and if you reject access to email, then the current implementation does not work.  The case above is much more forgiving.  profile.name will always be returned, because that is the absolute minimum when authenticating with facebook.

mleanos:
I've tested this, and it works for me; after updating my local env config to the correct fb callbackURL... `callbackURL: '/api/auth/facebook/callback'`

I still don't fully understand the implementation here though. I'll go through it thoroughly.

lirantal:
Great, merging.
Thanks @rhutchison!

PR #843, opened 2015-08-23
[Bug] Previous State (History)

rhutchison:
This addresses the issue with redirecting to a not-found/exception state after login.

##### Steps
1. http://localhost:3000/khdshskddsakh
2. Click Sign In & Authenticate

##### Expected
User should be redirected to the previous valid state or default state (home).

##### Actual
User is redirected to 'not-found' state.

### DIFF
--- a/modules/core/client/config/core.client.routes.js
+++ b/modules/core/client/config/core.client.routes.js
@@ -15,7 +15,10 @@ angular.module('core').config(['$stateProvider', '$urlRouterProvider',
       })
       .state('not-found', {
         url: '/not-found',
-        templateUrl: 'modules/core/client/views/404.client.view.html'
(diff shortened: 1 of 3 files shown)


modules/core/client/config/core.client.routes.js
```diff
@@ -15,7 +15,10 @@ angular.module('core').config(['$stateProvider', '$urlRouterProvider',
       })
       .state('not-found', {
         url: '/not-found',
-        templateUrl: 'modules/core/client/views/404.client.view.html'
+        templateUrl: 'modules/core/client/views/404.client.view.html',
+        data: {
+          ignoreState: true
```
rhutchison: Open to suggestions for a more sensical propery name.

codydaig:
LGTM

PR #844, opened 2015-08-23
mask exception route for 404 responses.

rhutchison:
This will mask the not-found route by showing the not-found page without redirecting to the not-found state.

##### Steps
1. http://localhost:3000/khdshskddsakh

##### Expected
* Route should not change.
* User should see page not found.


codydaig:
LGTM

PR #845, opened 2015-08-23
Session Cookie Configuration enhancements

lirantal:
Refactoring the Session Cookie configuration and adding more configurable parameters

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -9,9 +9,18 @@ module.exports = {
   },
   port: process.env.PORT || 3000,
   templateEngine: 'swig',
-  // Session details
-  // session expiration is set by default to 24 hours
-  sessionExpiration: 24 * (60 * 60 * 1000),
+  // Session Cookie settings
+  sessionCookie: {
+    // session expiration is set by default to 24 hours
+    maxAge: 24 * (60 * 60 * 1000),
+    // httpOnly flag makes sure the cookie is only accessed
+    // through the HTTP protocol and not JS/browser 
+    httpOnly: true,
+    // secure cookie should be turned to true to provide additional
+    // layer of security so that the cookie is set only when working
+    // in HTTPS mode.
+    secure: false
+  },
   // sessionSecret should be changed for security measures and concerns
   sessionSecret: 'MEAN',
   // sessionKey is set to the generic sessionId key used by PHP applications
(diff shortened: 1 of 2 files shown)


config/env/default.js
```diff
@@ -9,9 +9,18 @@ module.exports = {
...
   port: process.env.PORT || 3000,
   templateEngine: 'swig',
-  // Session details
-  // session expiration is set by default to 24 hours
-  sessionExpiration: 24 * (60 * 60 * 1000),
+  // Session Cookie settings
+  sessionCookie: {
+    // session expiration is set by default to 24 hours
+    maxAge: 24 * (60 * 60 * 1000),
+    // httpOnly flag makes usre the cookie is only accessed
```
pgrodrigues: sure*

lirantal:
thanks @pgrodrigues - fixed spelling issue.

pgrodrigues:
Also I may be overlooking a bit the point but according to the docs: "Please note that secure: true is a recommended option. However, it requires an https-enabled website, i.e., HTTPS is necessary for secure cookies. If secure is set, and you access your site over HTTP, the cookie will not be set."

I'm not sure what would be the behaviour of the cookie session if by accident `secure` is set to `true` and https is disabled. For example: working with NODE_ENV='dev' with secure: true by mistake. Would the cookie still be set without throwing error and not work as it should?

Since cookie secure flag depends on the mean app secure flag being true shoulnd't this line be something like:

    cookie: {
        maxAge: config.sessionCookie.maxAge,
        httpOnly: config.sessionCookie.httpOnly,
        secure: config.sessionCookie.secure && config.secure
    },

Just a thought, probably it doesn't make sense but I had this doubt.

jloveland:
I was thinking the same thing @pgrodrigues

lirantal:
@pgrodrigues overall you're right, and that's exactly the reason why I made the secure option default to false. Because I didn't want to force it on for folks not using SSL. Also the rationale is that whomever will set that secure flag to true *knows* what they are doing and *knows* that they need to be running in SSL mode.

I like the option to kind of dummy proof it with also checking that secure is on, so unless anyone else sees an issue with that I'll indeed add it.

codydaig:
@jloveland @lirantal @pgrodrigues 

I've started a "How to Secure your MEAN.js application" section to the documentation to start covering the different options that are being added.

lirantal:
Thanks @codydaig 

@pgrodrigues @jloveland I applied the suggestion

lirantal:
Merging. 
@codydiag can you mention me in the comments for that security documention thread so I can follow up?

PR #851, opened 2015-08-25
Security check as part of our CI

lirantal:
Adding the NSP tool to audit our packages version for known security vulnerabilities as part of the CI process

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,6 @@ services:
 before_install:
   - gem update --system
   - gem install sass --version "=3.3.7"
+  - npm i nsp -g
+after_script:
+  - nsp audit-package


lirantal:
I was thinking of adding the `nsp` package as part of the project and to include a grunt task for it so anyone can locally test this too and include in their CI process if required.
WDYT?

codydaig:
I'm always down for more testing.

mleanos:
:+1: Adding `nsp` to the project is a great idea!

lirantal:
I think we'll leave nsp package to simply report vulnerable packages but not actually fail the build because some errors that it reports aren't really dependent on something we can do on the project level, for example:
```
Name       Installed   Patched   Vulnerable Dependency
uglify-js    2.3.6    >= 2.4.24  meanjs@0.4.0 > bower@1.5.2 > handlebars@2.0.0 > uglify-js@2.3.6
semver       2.3.2     >=4.3.2   meanjs@0.4.0 > bower@1.5.2 > semver@2.3.2
```
Where the latest version of bower is already 1.5.2 so there's really nothing to upgrade to, or fix, until bower itself updates.

codydaig:
:+1:

PR #853, opened 2015-08-26
Gulp task enhancements - template cache

rhutchison:
Orig: #704 - These are updates to gulp build automation.

### Gulp Template Cache

### Auto Prefixer


codydaig:
LGTM

vaucouleur:
This is brillant :+1: thanks @rhutchison  

Disabling the template cache is just a matter of removing 'templatecache' ? (for large web sites with lots of content)

rhutchison:
@vaucouleur correct.  Glad you like it!  Let me know your experience.  I have not seen any downsides yet.

vaucouleur:
@rhutchison I tested it with a mobile phone client, just to make the test a little bit more interesting, and it works just fine. Instinctively I would have tried to leverage the browser cache, and have more of a "pay as you use/go" approach -- but it's not obvious whether this is actually (easily) doable. Your approach seems to be recommended elsewhere,  and it's easy to disable it, so it's all good. Thanks !

lirantal:
Great!

PR #863, opened 2015-08-30
[hotfix] Fixed User model tests

mleanos:
@lirantal PR #840 changed the global var `user` to `user1` in the User model server tests. This was merged and then #858 was merged, which was still referencing the global var as `user` in the new *roles* tests. This is causing jshint failures.

This change updates the new *roles* tests to use `user1`

PR #864, opened 2015-08-31
[bug] Remove non CF specific files [closes #861]

### LINKED ISSUE #861: $injector:modulerr when deploying to Pivotal
The following:

```
$ git clone https://github.com/meanjs/mean.git && cd mean
$ npm install
$ grunt build
$ cf login -a api.run.pivotal.io
$ cf create-service mongolab sandbox mean-mongo
$ cf push
```

Produces:
```
Uncaught Error: [$injector:modulerr] http://errors.angularjs.org/1.3.17/$injector/modulerr?p0=mean&p1=Error%3A%2…epressible-petrology.cfapps.io%2Flib%2Fangular%2Fangular.min.js%3A17%3A381)
```

When I navigate to the link provided by `cf push`

I'm still new to all of this, so am I missing / not changing something that I need to be?

codydaig:
Since the CF env assets doesn't contain anything specific, I stripped it down so it will default to the default settings. 

Closes #861 

@ryanjbaxter

### DIFF
--- a/config/assets/cloud-foundry.js
+++ b/config/assets/cloud-foundry.js
@@ -1,23 +1,3 @@
 'use strict';
 
-module.exports = {
-  client: {
-    lib: {
-      css: [
-        'public/lib/bootstrap/dist/css/bootstrap.min.css',
-        'public/lib/bootstrap/dist/css/bootstrap-theme.min.css',
-      ],
-      js: [
-        'public/lib/angular/angular.min.js',
-        'public/lib/angular-resource/angular-resource.min.js',
-        'public/lib/angular-animate/angular-animate.min.js',
-        'public/lib/angular-ui-router/release/angular-ui-router.min.js',
-        'public/lib/angular-ui-utils/ui-utils.min.js',
-        'public/lib/angular-bootstrap/ui-bootstrap-tpls.min.js',
-        'public/lib/angular-file-upload/angular-file-upload.min.js'
-      ]
-    },
-    css: 'public/dist/application.min.css',
-    js: 'public/dist/application.min.js'
-  }
-};
+module.exports = {};


codydaig:
Tests are failing waiting on #863. They have nothing to do with this specific PR.

codydaig:
@jd-carroll Can you try this to see if this resolves your issue?

lirantal:
Re-running the CI here to make sure the tests pass.

mleanos:
LGTM

ryanjbaxter:
@codydaig My original though here was that if you were deploying to CF that you would want to use the minimized version of these files, just like in production.  However now that I think about it you may just be doing development and/or testing on CF so maybe it is not necessary.  I suppose if you want to run production you can add the files in after the fact.

jd-carroll:
That seems to do the trick! I'm a little confused by the change, I would have expected the opposite almost. So if someone could explain a little more why this fixes the issue it would be appreciated. 

Also, every time I run cf push, I first have to cut/paste my node_modules folder. Then once done, cut/paste the node_modules folder back into the project. Anyone know why this might be the case?

ryanjbaxter:
@jd-carroll the problem was that the CF environment file was missing one of the required angular dependencies.  By removing all the dependencies from the CF environment file we will inherit the default environment dependencies therefore eliminating the need to maintain dependencies in many different environments.

As for you issue with the node_modules folder, you shouldn't have to do that.  In fact the node_modules folder isn't even uploaded to CF when you do the `cf push` so it should not effect anything.  Ping me on Gitter and we can talk more.

codydaig:
@jd-carroll The configuration of cloud-foundry.js and default.js are merged together and the environment file takes priority. It's smart enough to merge objects deeply, however, when it gets to an array the array in the environment config overwrites the array in the default config. Hope that clarifies things a bit.

@ryanjbaxter That's the difficult part about having cloud-foundry as a seperate enviorment, it doesn't inherit development or production. It is its own env. So yes, once the user decides developemnt or production, they can manually set-up their configuration files. 

@lirantal Sounds like this is good to go then.

lirantal:
Great!

Issue #847, opened 2015-08-24
Security issue: Don't read the host of the application from the HTTP header


Koblaid:
mean.js reads the host of the application from the HTTP header and stores it in req.locals (see https://github.com/meanjs/mean/blob/master/config/lib/express.js#L43). This value is then used for example to generate the password reset link: https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js#L62

The HTTP host header field can be spoofed by an attacker. This could be used for example to let the server generate a password reset email with an url which contains the token but points to a completely different domain. More details: http://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

We solved this problem by explicitly specifying the domain of the app in the config and use this value whenever we want to generate urls.

lirantal:
@Koblaid the code https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js#L62 is simply creating the password reset URL, at that point it's not sent to anyone and it is created per the user who views that link.

Can you explain how would an attacker exploit that then?

lirantal:
@Koblaid nevermind I think I understand what you were referring to.
In my opinion we can handle this by:
1. introducing a config variable for users to setup their domain/host which will be used across the site
2. If (1) doesn't exist then we fallback to the `req.headers.host` or maybe the request's originating IP is more secure?

WDYT?

Koblaid:
I would prefer 1 and refuse to start if the config option does not exist.

Falling back to req.headers.host would mean to accept the security risk if the user forgets the option or is too lazy and does not understand the security implications. I think enforcing security is a good thing, especially if its done with a simple config variable.

The requests originating IP would be the IP-address of the requesting host, right? This doesn't help anything, does it? Or do you mean the IP-address under which the server is running? I'm not sure if/where it could be accessed in a non-spoofable-manner. In addition, the IP of the actual server could differ from the IP under which the site is accessable from the outside (for example on Heroku). Also the IP would look suspicious in the email and using HTTPS would result in a browser warning.

So I think enforcing the configuration option is the only secure way to handle this.

lirantal:
Well "refuse" is a bit harsh.
Some may choose to use MEAN.JS as an internal application and will not give much weight as to security considerations like that.

Can you offer another option without restricting the domain config?

Koblaid:
The only way of a fallback is using the HTTP header like it is done at the moment.

[...]

PR #871, opened 2015-09-02
Add config options 'domain'

Koblaid:
The value will be used when the backend has to generate urls which
point to the app. Using the hostname sent in the HTTP header is
dangerous since an attacker could send a request with a spoofed
HTTP header (HTTP Host header attack).

If the config is not set a loud warning is printed on startup and
the hostname of the HTTP header is used nevertheless.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -67,6 +67,15 @@ var validateEnvironmentVariable = function () {
 };
 
 /**
+ * Validate config.domain is set
+ */
+var validateDomainIsSet = function (config) {
+  if(!config.app.domain){
+    console.log(chalk.red('+ Important warning: config.domain is empty. For security reasons it should be set to the domain of the app.'));
+  }
+};
+
+/**
  * Validate Secure=true parameter can actually be turned on
  * because it requires certs and key files to be available
  */
@@ -186,6 +195,9 @@ var initGlobalConfig = function () {
   // read package.json for MEAN.JS project information
   var pkg = require(path.resolve('./package.json'));
   config.meanjs = pkg;
+
+  // Print a warning if config.domain is not set
+  validateDomainIsSet(config);
 
   // We only extend the config object with the local.js custom/local environment if we are on
   // production or development environment. If test environment is used we don't merge it with local.js
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -5,7 +5,8 @@ module.exports = {
     title: 'MEAN.JS',
     description: 'Full-Stack JavaScript with MongoDB, Express, AngularJS, and Node.js',
     keywords: 'mongodb, express, angularjs, node.js, mongoose, passport',
-    googleAnalyticsTrackingID: process.env.GOOGLE_ANALYTICS_TRACKING_ID || 'GOOGLE_ANALYTICS_TRACKING_ID'
+    googleAnalyticsTrackingID: process.env.GOOGLE_ANALYTICS_TRACKING_ID || 'GOOGLE_ANALYTICS_TRACKING_ID',
+    domain: process.env.DOMAIN
   },
   port: process.env.PORT || 3000,
   templateEngine: 'swig',
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -32,7 +32,8 @@ module.exports = {
     }
   },
   app: {
-    title: defaultEnvConfig.app.title + ' - Development Environment'
+    title: defaultEnvConfig.app.title + ' - Development Environment',
+    domain: process.env.DOMAIN || 'localhost:' + defaultEnvConfig.port
   },
   facebook: {
     clientID: process.env.FACEBOOK_ID || 'APP_ID',
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -40,8 +40,9 @@ module.exports.initLocalVariables = function (app) {
 
   // Passing the request url to environment locals
   app.use(function (req, res, next) {
-    res.locals.host = req.protocol + '://' + req.hostname;
-    res.locals.url = req.protocol + '://' + req.headers.host + req.originalUrl;
+    var domain = config.app.domain || req.headers.host;
+    res.locals.host = req.protocol + '://' + domain;
+    res.locals.url = req.protocol + '://' + domain + req.originalUrl;
     next();
   });
 };
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -56,7 +56,12 @@
 
   {% if livereload %}
   <!--Livereload script rendered -->
-  <script type="text/javascript" src="{{host}}:35729/livereload.js"></script>
+  <script type="text/javascript">
+    var scriptTag = document.createElement('script');
+    scriptTag.setAttribute('type', 'text/javascript');
+    scriptTag.setAttribute('src', '//' + window.location.hostname + ':35729/livereload.js');
+    document.head.appendChild(scriptTag);
+  </script>
   {% endif %}
 </body>
 
--- a/config/env/test.js
+++ b/config/env/test.js
@@ -33,7 +33,8 @@ module.exports = {
   },
   port: process.env.PORT || 3001,
   app: {
-    title: defaultEnvConfig.app.title + ' - Test Environment'
+    title: defaultEnvConfig.app.title + ' - Test Environment',
+    domain: process.env.DOMAIN || 'localhost:3001'
   },
   facebook: {
     clientID: process.env.FACEBOOK_ID || 'APP_ID',


Koblaid:
PR for #847 

This has one problem though: It breaks livereload in its current implementation.

The problem is that the port of the server has to be appended to the new config option, so the value is for example `http://localhost:3000`. Then in the template the livereload port is appended: `http://localhost:3000:35729`.

Omitting the port would break the URLs generated by the backend such as the passwort reset link. I also thought about just using `config.domain + config.port` to build the url. But this would break on Heroku. On Heroku `config.port` is set to some arbitary port even though the app is reachable under port 80.

On possibility would be to use `config.domain.split(':')[0]` to extract the host for livereload.

simison:
This works for the Livereload:
```
      document.write('\x3Cscript type="text/javascript" src="//' + window.location.hostname + ':35729/livereload.js">\x3C/script>');
    </script>
```
...then it doesn't matter if the thing is served at localhost, docker, staging... JS always picks up the current domain from the address bar.

Koblaid:
Jup, that should work.

Hmm... is the content of `app.locals` considered as API? Then changing app.locals.host would be a breaking change.

config/config.js
```diff
@@ -67,6 +67,15 @@ var validateEnvironmentVariable = function () {
 };
 
 /**
+ * Validate config.domain is set
+ */
+var validateDomainIsSet = function (config) {
+  if(!config.app.domain){
+    console.error(chalk.red('+ Important warning: config.domain is empty. To prevent HTTP Host header attacks please set config.domain.'));
```
codydaig: To be consistent, I don't think this should be `console.error`, just `console.log`.

config/env/development.js
```diff
@@ -22,7 +22,8 @@ module.exports = {
     }
   },
   app: {
-    title: defaultEnvConfig.app.title + ' - Development Environment'
+    title: defaultEnvConfig.app.title + ' - Development Environment',
+    domain: 'localhost:3000'
```
codydaig: Instead of hardcoding port 3000, it is possible have it default to the port specified in the config file?

mleanos: Why is the port required here? Is it because you're using `req.headers.hostname`?

config/lib/express.js
```diff
@@ -40,8 +40,9 @@ module.exports.initLocalVariables = function (app) {
 
   // Passing the request url to environment locals
   app.use(function (req, res, next) {
-    res.locals.host = req.protocol + '://' + req.hostname;
-    res.locals.url = req.protocol + '://' + req.headers.host + req.originalUrl;
+    var domain = config.app.domain || req.headers.hostname;
```
mleanos: Shouldn't we use `req.hostname`, so we don't have to worry about the port? We can get the port from the config. Unless there's something that I'm unaware of, this would fix the issue you encountered with *livereload*

Koblaid: Using req.hostname is dangerous, since that enables HTTP Host Header attacks. It doesn't really matter for livereload, but the variable `res.locals.host` could have been used for more important things as well.

Using the port from the environment breaks when the app runs behind a proxy (ie on Heroku or AWS). In this case the port of the app is completely arbitary while the external port is still 80.

config/env/development.js
mleanos: Could this be changed to `domain: process.env.DOMAIN || 'localhost'`? and added to the test env config as well?

codydaig:
What about: If a domain isn't set in the config, just default to req.hostname?

config/lib/express.js
mleanos: @Koblaid I meant that we could use `req.hostname` when there is no domain set in the env config; as a default instead of `req.headers.hostname`. This way a user of this project could implement the domain, if they are concerned with this vulnerability; for instance, when the project is early dev or used internally/locally only. Otherwise, if the domain is omitted then it would default to it's current behavior.

config/config.js
lirantal: Right

config/env/development.js
lirantal: indeed, let's not hardcode the port there with the domain.

config/lib/express.js
Koblaid: @mleanos I just checked the express docs: req.hostname is the same as req.headers.hostname (except when `trust proxy` is on). So I agree, we should switch to `req.hostname`.

Koblaid: @mleanos Wait, no. Both `req.hostname` and `req.headers.hostname` won't work since they both exclude the port. The values in `res.locals` should include the port (if its not 80), right?

So we have to use `req.headers.host` which includes the port.

Koblaid:
Alright, I just pushed some changes. The domain is now retrieved via `req.headers.host` which includes the port and should also work behind a proxy.

AFAICS the remaining issue is the hard-coded port in development.js. Omitting the port in `config.domain` would break urls generated by the app since the port has to be included.

So we could:

1. We could remove the default value in development.js. That would trigger the 'domain is not set' warning
   1. We could ommit the warning if NODE_ENV=development
2. We could add a new config variable `config.defaultPort` and use it in default.js for `config.port` and in develepment.js for `config.domain`

modules/core/server/views/layout.server.view.html
```diff
@@ -56,7 +56,9 @@
 
   {% if livereload %}
   <!--Livereload script rendered -->
-  <script type="text/javascript" src="{{host}}:35729/livereload.js"></script>
+  <script type="text/javascript">
+    document.write('\x3Cscript type="text/javascript" src="//' + window.location.hostname + ':35729/livereload.js">\x3C/script>');
```
lirantal: what's the reason for doing it like this? is that in order to take the "real" hostname from the URL rather than the config variable? If so, isn't there a cleaner way of doing it instead of document.write?

config/env/default.js
```diff
@@ -5,7 +5,8 @@ module.exports = {
     title: 'MEAN.JS',
     description: 'Full-Stack JavaScript with MongoDB, Express, AngularJS, and Node.js',
     keywords: 'mongodb, express, angularjs, node.js, mongoose, passport',
-    googleAnalyticsTrackingID: process.env.GOOGLE_ANALYTICS_TRACKING_ID || 'GOOGLE_ANALYTICS_TRACKING_ID'
+    googleAnalyticsTrackingID: process.env.GOOGLE_ANALYTICS_TRACKING_ID || 'GOOGLE_ANALYTICS_TRACKING_ID',
+    domain: process.env.DOMAIN
```
lirantal: this can be `domain: process.env.DOMAIN || 'localhost'` as @mleanos suggested

lirantal:
@Koblaid about the port number issue, we already have a port number config option, can't we use it? for example:
`domain: 'localhost' + port`

or will that not solve the issue?

modules/core/server/views/layout.server.view.html
Koblaid: Hehe, I didn't really like the code either, but was too lazy to really check it. Would this be better?
```
var scriptTag = document.createElement('script');
scriptTag.setAttribute('type', 'text/javascript');
scriptTag.setAttribute('src','//' + window.location.hostname + ':35729/livereload.js');
document.head.appendChild(scriptTag);

File: config/env/default.js (hunk 5 above)
Comment (Koblaid) [inline config/env/default.js]: `domain: process.env.DOMAIN || 'localhost'` won't work since it has to include the port.

For example it could (and should) be used to assemble the URL to reset the password. This url has to include the port.

**Comment by Koblaid** [comment]
This works on localhost but it will break if the app is running for example on Heroku. Heroku sets the PORT environment variable to some arbitary port, while the app is reachable with port 80 from the outside.

File: modules/core/server/views/layout.server.view.html (hunk 4 above)
Comment (lirantal) [inline modules/core/server/views/layout.server.view.html]: Yes, that's much better if we actually need to change it like that.

File: config/env/default.js (hunk 5 above)
Comment (lirantal) [inline config/env/default.js]: Of course, so just doing something like this then:
```
domain: process.env.DOMAIN || 'localhost' + port
```
where `port` is already a configuration parameter as part of the config.

Does that make sense?

Comment (Koblaid) [inline config/env/default.js]: Ah, yes, that would work. But what purpose does it serve? If the domain is missing it is already set dynamically in `express.js`.

Also I would only do this in `development.js`, not in `default.js` because `'localhost' + port` is most certainly wrong for production. Setting it `development.js` would prevent the 'no-domain-set' warning in development mode, so that would be  useful.

**Comment by simison** [comment]
@Koblaid that's true also for any proxy configuration, say Nginx/Phusion Passenger.

File: modules/core/server/views/layout.server.view.html (hunk 4 above)
Comment (simison) [inline modules/core/server/views/layout.server.view.html]: >is that in order to take the "real" hostname from the URL rather than the config variable?

Yep. If testing/doing development behind proxy (e.g. Nginx) setup inside Vagrant/Docker, livereload would not work.

File: config/env/development.js (hunk 2 above)
Comment (lirantal) [inline config/env/development.js]: @Koblaid I still don't like the port hardcoded there like that...

**Comment by lirantal** [comment]
@Koblaid see my latest comment for this PR: https://github.com/meanjs/mean/pull/871#discussion_r41043772

**Comment by Koblaid** [comment]
@lirantal Oh, sorry, at first I didn't know that you were still waiting for an answer and than I was busy ;-)

Well, I don't know how it could work out of the box without hard-coding the domain in development mode. (By the way: the default port is also hard-coded in `default.js`, isn't that the same thing?)

We could drop the `domain` line in `development.js` but then the user would be greeted with the "domain config is empty" warning, until he sets the option.

Or we could separate the default port in `default.js` so we would have two port variables: 

1. one under which the node instance is reachable 
2. one under which the site is reachable from the outside

Than the first one could be used to start up the server and the second one could be used to dynamically generate the domain.

But I think that would only make things more complicated.

File: config/env/default.js (hunk 5 above)
Comment (lirantal) [inline config/env/default.js]: @Koblaid I think you're right and we can leave this setting as is, but make the discussed change here in the `development.js` file and `test.js` file. Look on my comments there.

File: config/env/development.js (hunk 2 above)
Comment (lirantal) [inline config/env/development.js]: @Koblaid We want to allow setting the domain from an env variable, or to be defined explicitly in the configuration file, which is what @mleanos was suggesting. So this change should be made:
```javascript
domain: process.env.DOMAIN || 'localhost'
```
And it should be added after the `port` configuration option so you can re-use it of course.

I'm ok with setting that only on `development.js` and `test.js` as you suggested.

modules/core/server/views/layout.server.view.html
lirantal: Ok.

lirantal:
@Koblaid see my latest comments on the code section to finalize this PR.
@ilanbiala do you consider this a breaking change due to the domain/livereload change?

modules/core/server/views/layout.server.view.html
ilanbiala: Can we get this line changed to be better?

lirantal: Yep.
@Koblaid we're still waiting for you to change that line to your recommendation at https://github.com/meanjs/mean/pull/871/files#r39917678

Koblaid: Just pushed

Koblaid:
Alright, I hope its ok now. The livereload-tag is now generated as proposed, and the domain-config in development.js and test.js is set to a default value but overwritable by the environment variable.

I wasn't sure about including the port into the default value for the domain-config in develepment.js and test.js. Should I remove it? I think it's necessary for URLs generated from the config to be correct. But I can still remove it if you like.

config/env/development.js
```diff
@@ -22,7 +22,8 @@ module.exports = {
     }
   },
   app: {
-    title: defaultEnvConfig.app.title + ' - Development Environment'
+    title: defaultEnvConfig.app.title + ' - Development Environment',
+    domain: process.env.DOMAIN || 'localhost:3000'
```
lirantal: you can do here:
```javascript
domain: process.env.DOMAIN || 'localhost:' + defaultEnvConfig.port;
```
I think it's better because this way you don't hard code the port number.

ilanbiala:
Still needs to be rebased before we can even consider merging it.

lirantal:
@Koblaid see my latest comments about the port and:
1. update the PR to solve the conflicts
2. squash

can you think of adding some test to check the domain configuration works or this is just plain config option that has no point in testing?

Koblaid:
@lirantal 
- Changed `development.js` according to your latest comment
- rebased on master
- squashed

ilanbiala:
@Koblaid can you add any meaningful tests?

Koblaid:
Hmm... I didn't see any existing tests for the configuration. Are there any?

If not, what kind of tests do you have in mind? Checking if the configuration is correctly created for the different modes? Or that the created configuration acutally works for the different setups (with proxy, without proxy, ...)?

Anyway, I think creating a test infrastructure for the configuration should be a separate issue/PR.

ilanbiala:
@Koblaid tests always come with the implementation, that way we don't forget about them. Add whatever tests you see as meaningful and we'll go from there.

mleanos:
@Koblaid We've defined configuration tests here. Perhaps a test could be written that is similar to some of these?
https://github.com/meanjs/mean/blob/master/modules/core/tests/server/core.server.config.tests.js

codydaig:
@Koblaid Any update on this?

mleanos:
I think it makes sense to move this to the 0.5.0 Milestone since there hasn't been any movement on this for over a week.

Also, it seems that the Token Auth implementation might revolve around the domain/host options. We might just want to address this once we have a better idea on what Token Auth changes on the server-side.

codydaig:
I second @mleanos that we should push this back.

lirantal:
Sure let's move to 0.5.0 so it doesn't block anything (I don't see how it's related to the token auth though)

mleanos:
@lirantal I thought the domain/host may be related because in the proposed PR for Token Auth ( #1040 ), there were some changes to the API base URL; perhaps this isn't really related, but wanted to point it out.

lirantal:
@mleanos @codydaig can we discuss this again or?

mleanos:
@lirantal Sure. I think the changes introduced in https://github.com/meanjs/mean/pull/1101 are related. Correct me if I'm wrong.

Is there anything in particular that you'd like to discuss?

lirantal:
@melanos right, and I think we need to somehow merge this PR with the changes made with #1101

lirantal:
@Koblaid would you like to make the changes required here to accommodate for the livereload update?

codydaig:
#1101 was merged in. What's left to do on this PR?

lirantal:
@codydaig let's close it for now.
@Koblaid can re-open/comment in the future if required.

Issue #848, opened 2015-08-24
Option to require stronger passwords


jloveland:
We should considering adding an option to allow developers to set the password strength and have the client and server test for this. A good option would be to follow OWASP password length & complexity. This could leverage libs such as:
https://www.npmjs.com/package/owasp-password-strength-test

lirantal:
I vote for not making this an option but rather part of the framework.
I like it that we're making sure the framework is following secure guidelines and so we should adopt it.

@jloveland care to PR it?

jloveland:
sure, I will work it this week.

pgrodrigues:
:+1:

codydaig:
:+1:

0xNacho:
:+1:

jloveland:
Just and update on this: I got the server side working and accompanying unit tests this weekend. I am now working in the client side.

lirantal:
Nice!

jloveland:
@lirantal assuming you can close this issue.

lirantal:
Great, thanks, I updated the milestone too.

codydaig:
@lirantal Any reason in particular we're keeping this open?

jloveland:
you can close this issue..resolved in #878

PR #878, opened 2015-09-03
Adding stronger password security

jloveland:
Adding stronger password requirements for improving security based on OWASP. Fixing #848

### DIFF
--- a/modules/users/client/directives/password-validator.client.directive.js
+++ b/modules/users/client/directives/password-validator.client.directive.js
@@ -1,0 +1,42 @@
+'use strict';
+
+angular.module('users')
+  .directive('passwordValidator', ['PasswordValidator', function(PasswordValidator) {
+    return {
+      require: 'ngModel',
+      link: function(scope, element, attrs, modelCtrl) {
+        modelCtrl.$parsers.unshift(function (password) {
+          var result = PasswordValidator.getResult(password);
+          var strengthIdx = 0;
+
+          // Strength Meter - visual indicator for users
+          var strengthMeter = [
+            { color: "danger", progress: "20" },
+            { color: "warning", progress: "40"},
+            { color: "info", progress: "60"},
+            { color: "primary", progress: "80"},
+            { color: "success", progress: "100"}
+          ];
+          var strengthMax = strengthMeter.length;
+
+          if (result.errors.length < strengthMeter.length) {
+            strengthIdx = strengthMeter.length - result.errors.length - 1;
+          }
+
+          scope.strengthColor = strengthMeter[strengthIdx].color;
+          scope.strengthProgress = strengthMeter[strengthIdx].progress;
+
+          if (result.errors.length) {
+            scope.popoverMsg = PasswordValidator.getPopoverMsg();
+            scope.passwordErrors = result.errors;
+            modelCtrl.$setValidity('strength', false);
+            return undefined;
+          } else {
+            scope.popoverMsg = '';
+            modelCtrl.$setValidity('strength', true);
+            return password;
+          }
+        });
+      }
+    };
+}]);
--- a/modules/users/client/directives/password-verify.client.directive.js
+++ b/modules/users/client/directives/password-verify.client.directive.js
@@ -1,0 +1,33 @@
+'use strict';
+
+angular.module('users')
+  .directive("passwordVerify", function() {
+    return {
+      require: "ngModel",
+      scope: {
+        passwordVerify: '='
+      },
+      link: function(scope, element, attrs, modelCtrl) {
+        scope.$watch(function() {
+          var combined;
+          if (scope.passwordVerify || modelCtrl.$viewValue) {
+            combined = scope.passwordVerify + '_' + modelCtrl.$viewValue;
+          }
+          return combined;
+        }, function(value) {
+          if (value) {
+            modelCtrl.$parsers.unshift(function(viewValue) {
+              var origin = scope.passwordVerify;
+              if (origin !== viewValue) {
+                modelCtrl.$setValidity("passwordVerify", false);
+                return undefined;
+              } else {
+                modelCtrl.$setValidity("passwordVerify", true);
+                return viewValue;
+              }
+            });
+          }
+        });
+     }
+    };
+});
--- a/modules/users/client/services/password-validator.client.service.js
+++ b/modules/users/client/services/password-validator.client.service.js
@@ -1,0 +1,19 @@
+'use strict';
+
+// PasswordValidator service used for testing the password strength
+angular.module('users').factory('PasswordValidator', ['$window',
+  function ($window) {
+    var owaspPasswordStrengthTest = $window.owaspPasswordStrengthTest;
+
+    return {
+      getResult: function (password) {
+        var result = owaspPasswordStrengthTest.test(password);
+        return result;
+      },
+      getPopoverMsg: function () {
+        var popoverMsg = "Please enter a passphrase or password with greater than 10 characters, numbers, lowercase, upppercase, and special characters.";
+        return popoverMsg;
+      }
+    };
+  }
+]);
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -23,7 +23,7 @@ describe('User Model Unit Tests:', function () {
       displayName: 'Full Name',
       email: 'test@test.com',
       username: 'username',
-      password: 'password',
+      password: 'M3@n.jsI$Aw3$0m3',
       provider: 'local'
     };
     // user2 is a clone of user1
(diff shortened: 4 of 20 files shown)


modules/users/client/directives/password-validator.client.directive.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+angular.module('users')
+  .directive('passwordValidator', ['$window', function($window) {
+    return {
+      require: 'ngModel',
+      link: function(scope, element, attrs, modelCtrl) {
+        modelCtrl.$parsers.unshift(function (password) {
+          var owaspPasswordStrengthTest = $window.owaspPasswordStrengthTest;
```
mleanos: Can we have a client service for `owasp`? Some user's might want to extend it, or have a bit more control over it's behavior.

mleanos:
@jloveland This all looks pretty good. I left a line comment. I haven't taken this for a test drive yet, but I plan to sometime tonight. Overall, I like what I see. Great job!

jloveland:
Ok, I will add a client service tonight.

jloveland:
ok, added a client service, check it out.

mleanos:
@jloveland I'll look at this again tonight. Got side-tracked yesterday with the jshint issue.

modules/users/tests/server/user.server.model.tests.js
```diff
@@ -217,9 +216,87 @@ describe('User Model Unit Tests:', function () {
...
+
+      _user1.validate(function (err) {
+        err.errors.password.message.should.equal("The password must be fewer than 128 characters.");
+        done();
+      });
+    });
+
+    it('should not allow more than 3 or more repeating characters ', function (done) {
+      var _user1 = new User(user1);
+      _user1.password = 'L0veSexxxSecre+God';
```
mleanos: might want to change that :)

mleanos:
@jloveland I took this for a test drive, and everything is working perfectly. I see you added tests as well :) 

I left a line comment about one of the passwords you have set in one of the new tests. The only other suggestion might be to remove the pop-over when the validation is at 100%. This is merely a suggestion; I just wanted to point it out.

Everything else looks really good. I think this is good to go. Thank you for knocking this out. It's really a great enhancement.

modules/users/tests/server/user.server.model.tests.js
jloveland: haha..yeah, I copied the test passwords from the owasp lib [here](https://github.com/nowsecure/owasp-password-strength-test/blob/master/test.js#L30), I can change that to be less *interesting*..

jloveland:
I can also disable the pop-over once the test passes.

jloveland:
this is still WIP. I noticed that
1. I missed the reset password form.
2. password reset request fails because of `undefined` password
So, I am working on these now.

I need some advice on how to implement closing the popover once the password field is valid. I tried `popover-is-open="{{userForm.password.$error}}"` and that was causing conflicts..

jloveland:
I added password validation and verification, fixed tests, etc. I did not implement disabling the popover, maybe that can be done on a different PR? 

Also, as you can see, I had some issues with squashing my commits properly, so I pushed anyway. Maybe someone can help me with the rebase?

simison:
Great! Testing this out. 

~~UX wise it would be clearer if those requirements would be visible already from the beginning, not only when I hit "sign up". Now it gives blue 80% (=good?) strength for complicated but short passwords.~~

__Edit:__ Oh, there indeed is a popup telling all that. It just doesn't appear until I type something into the field and then click out and focus again to the input.

Incidentally I have an easy to use Angular module for this stuff. :-) https://github.com/trustroots/trustpass Check if you like it.

jloveland:
I just squashed my commits, let me know what you think. @simison thanks for the suggestion. maybe we can evaluate your library and refactor with another PR once this is merged?

simison:
@jloveland sure! I'd try to fix that modal not appearing on first focus anyway. I think I've had that before too when trying to add content dynamically into popups/tooltips. Make sure you have the latest bootstrap-ui though.

mleanos:
@jloveland I've tested this out again, with the password recovery too. Looks good to me. Do you still need help with the rebasing?

To get my PR branches rebased, I usually get my local master branch current, with upstream (meanjs), and then checkout my PR branch, and perform `git rebase master` Sometimes, I'll get a warning that my origin & local have diverged. In which case, I just push to my origin with `--force`

jloveland:
@mleanos I think I got the rebase worked out. please let me know if something looks wrong.

@simison I saw a lot of references to similar issues when researching the *dynamic* popover. I could solve the popover issue by reverting to *static* popover...placing the message in the [html](https://github.com/jloveland/mean/blob/owasp-password-strength/modules/users/client/views/authentication/signup.client.view.html#L37) like so:
`popover="Please enter a passphrase or password with greater than 10 characters, numbers, lowercase, uppercase, and special characters."`

instead of the dynamic approach:
`popover="{{popoverMsg}}"`

This is not ideal, but it would fix the problem. What do you think?

mleanos:
@jloveland I'm not an expert with Git, so correct me if I'm wrong. I think having those other commits (other than from you) is the result of not squashing after your rebase. I usually try get my PR's down to one commit.

jloveland:
@mleanos, you were correct, I needed to `git rebase master` then squash commits.

Are there any Angular.js experts out there that can advise me on how to solve the issue with the dynamic popover message not displaying on focus properly? Otherwise, my next change will be to revert to static popover defined in the html.

simison:
It has something to do with dynamic content; try to make sure that variable always has some non-empty string content or try fiddling with `tooltip-enable="true|false"` or `tooltip-is-open="true|false"`.

jloveland:
- fixed the dynamic popover message scope issues by moving the message variable the service.
- made popover disabled when password strength passes
- made button sizes consistent with the rest of the views

mleanos:
@jloveland Very good. Glad you got those issues worked out. I've tested this, and everything is working wonderfully! Thanks. 

I think this is ready to go. @simison How's it look to you?

codydaig:
LGTM 

Thanks @jloveland! :-D

lirantal:
Generally reviewed it, LGTM.

mleanos:
@lirantal I had the same concern over existing user's. I tested this, and these changes didn't have an impact on my existing users.

These changes merely enforce the new password requirements, when a user attempts to change their password; or when a new user signs up.

@jloveland ?

jloveland:
@lirantal and @mleanos this does not affect existing users until they change their password. It will have immediate affect on users who signup.

lirantal:
Cool, let's continue with this then.

modules/users/client/directives/password-verify.client.directive.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+angular.module('users')
+  .directive("passwordVerify", function() {
+    return {
+      require: "ngModel",
```
simison: @jloveland JSHint is complaining about not using single quotes here. Could you fix and send another PL?

modules/users/client/directives/password-verify.client.directive.js
```diff
@@ -0,0 +1,33 @@
...
+          return combined;
+        }, function(value) {
+          if (value) {
+            modelCtrl.$parsers.unshift(function(viewValue) {
+              var origin = scope.passwordVerify;
+              if (origin !== viewValue) {
+                modelCtrl.$setValidity("passwordVerify", false);
+                return undefined;
+              } else {
+                modelCtrl.$setValidity("passwordVerify", true);
```
simison: should be single quotes

modules/users/client/services/password-validator.client.service.js
```diff
@@ -0,0 +1,19 @@
...
+  function ($window) {
+    var owaspPasswordStrengthTest = $window.owaspPasswordStrengthTest;
+
+    return {
+      getResult: function (password) {
+        var result = owaspPasswordStrengthTest.test(password);
+        return result;
+      },
+      getPopoverMsg: function () {
+        var popoverMsg = "Please enter a passphrase or password with greater than 10 characters, numbers, lowercase, upppercase, and special characters.";
```
simison: should be single quotes

jloveland: Thanks @simison , I didn't see this show after running `grunt jshint` on that branch. Output from the command:
```
grunt jshint
Running "jshint:all" (jshint) task
>> 105 files lint free.

Done, without errors.
```
Any thoughts on why I don't get the errors?
I can submit another PR to correct this though...

simison: A good question. I just copied the directive to my meanjs-based project and got this. Is /client/services/ a new folder?

BTW Thanks for writing these!

modules/users/client/directives/password-verify.client.directive.js
ilanbiala: also in the directive name above?

PR #887, opened 2015-09-05
Adding MEAN.JS version information as part of the startup info when app loads

lirantal:
I've been thinking it would be great if we can display the project version number when the app starts as that will easily allow other devs and users to clearly see which MEAN.JS release they are on.

We often get a lot of bug requests and this is usually the first version that comes up...

image

WDYT?

### DIFF
--- a/config/lib/app.js
+++ b/config/lib/app.js
@@ -46,6 +46,7 @@ module.exports.start = function start(callback) {
       if (process.env.NODE_ENV === 'secure') {
         console.log(chalk.green('HTTPs:\t\t\t\ton'));
       }
+      console.log(chalk.green('MEAN.JS version:\t\t\t' + config.meanjs.version));
       console.log('--');
 
       if (callback) callback(app, db, config);
--- a/config/lib/app.js
+++ b/config/lib/app.js
@@ -46,7 +46,9 @@ module.exports.start = function start(callback) {
       if (process.env.NODE_ENV === 'secure') {
         console.log(chalk.green('HTTPs:\t\t\t\ton'));
       }
-      console.log(chalk.green('MEAN.JS version:\t\t\t' + config.meanjs.version));
+      console.log(chalk.green('App version:\t\t\t' + config.meanjs.version));
+      if (config.meanjs['meanjs-version'])
+        console.log(chalk.green('MEAN.JS version:\t\t\t' + config.meanjs['meanjs-version']));
       console.log('--');
 
       if (callback) callback(app, db, config);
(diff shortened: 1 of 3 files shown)


jloveland:
:+1:

codydaig:
@lirantal What if someone changed the version in package.json to match their project's version? Why not have a meanjs-version attribute in package.json? That would also make things easier to managing versions in the generator and sub generators. I already have this in the proposed PR for the generator:
https://github.com/codydaig/generator-meanjs/blob/dev/0.2/app/templates/master/_package.json#L5
"meanjs-version": "0.4.0",

lirantal:
Sure, that's even better.
In which-case we can include both versions maybe (app and mean.js). WDYT?

codydaig:
SGTM

mleanos:
SGTM too! I like the idea of having both attributes displayed.

lirantal:
Great, I updated

lirantal:
image

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "meanjs",
   "description": "Full-Stack JavaScript with MongoDB, Express, AngularJS, and Node.js.",
   "version": "0.4.1",
+  "meanjs-sdversion": "0.4.1",
```
mleanos: Is this a typo? Shouldn't it be `"meanjs-version"`?

lirantal: Thanks, good catch I fixed it.

lirantal:
Fixed typo

codydaig:
LGTM. 

Side note: you may want to set coveralls to allow a certain percentage of decrease because it's failing you right now.

config/lib/app.js
```diff
@@ -46,6 +46,9 @@ module.exports.start = function start(callback) {
       if (process.env.NODE_ENV === 'secure') {
         console.log(chalk.green('HTTPs:\t\t\t\ton'));
       }
+      console.log(chalk.green('App version:\t\t\t' + config.meanjs.version));
```
ilanbiala: how about appVersion, and then we can do config.meanjs.meanVersion to avoid it being a string.

ilanbiala:
@lirantal is this going to be part of 0.4.2?

lirantal:
@ilanbiala part of the next release so I guess 0.4.2. Why, is that an issue?

lirantal:
@codydaig  I see it but it's not a real error, it's just letting you know that coverage decreased (in this case in 0.08%), it didn't fail the build (tests) and it didn't dramatically reduce code coverage (it's a known issue that we don't yet have tests for core stuff outside of the users/articles modules)

ilanbiala:
@lirantal just want to make sure it's in the right milestone for when we finalize the release and check the milestone for open issues.

lirantal:
I updated it...

mleanos:
This LGTM

codydaig:
LGTM

lirantal:
Great, thanks.

PR #890, opened 2015-09-06
Adding sample config/ directory for articles module

lirantal:
Dummy config directory for the articles module to make it easier for developers to understand and realize there's an actual per-module config/ directory/files that can be added

This came up from my gitter conversation with @trainerbill who was asking about modularizing the configuration approach so that modules can have their own config tasks and didn't realize it already existed and requested to have a 'sample' directory/file for the articles module as a general guide.

### DIFF
--- a/modules/articles/server/config/articles.server.config.js
+++ b/modules/articles/server/config/articles.server.config.js
@@ -1,0 +1,14 @@
+  'user strict';
+
+/**
+ * Module dependencies
+ */
+var path = require('path'),
+  config = require(path.resolve('./config/config'));
+
+/**
+ * Module init function.
+ */
+module.exports = function (app, db) {
+
+};


mleanos:
LGTM

ilanbiala:
@lirantal is this for 0.4.2? Wouldn't this be cleaner if we just put a comment in 1 config file in the core or something?

lirantal:
You're right, I usually did just 0.4.x in the past but we can just assign it to the next release in line by default, unless some discussion rises that requires to change it.

trainerbill:
My issue was that I always copy the articles module as a skeleton for making other modules.  I made a config folder and couldn't figure out why it was erroring.  I have a feeling others will do the same if we don't include the config folder in the articles module.

trainerbill:
Maybe we should think about including a "skeleton" module in the application?  one that is designed to be copied and used in a basic CRUD format?

lirantal:
A "skeleton" module sounds as an overkill at the moment, and basically that's the purpose of the articles module. The generator can be used to scaffold such skeletons for app development and with Cody's help we might be getting there.

So does it look good for merging?

trainerbill:
Fair enough.  LGTM

codydaig:
@trainerbill A "skeleton" module is on my todo list for the generator. Which would be even easier than copy and pasting since it will change the name throughout the new module.

lirantal:
Great, sounds good @codydaig

Issue #769, opened 2015-08-06
suggestion: add views in index.html and use templateCache


andreafdaf:
Hi, I've already done this in a project based on mean 0.4.0 so I guess I can make a PR if you want. Let me explain:

### config/assets/default.js
```
views: ['modules/*/client/**/*client.view.html']
```

### config/config.js
```
var initGlobalConfigFiles = function(config, assets) {
...
  config.files.client.html = getGlobbedPaths(assets.client.views, 'public/');
```

### config/lib/express.js
```
module.exports.initLocalVariables = function (app) {
...
  app.locals.htmlFiles = config.files.client.html;
```

### modules/core/server/controllers/core.server.controller.js
```
exports.renderIndex = function(req, res) {
  res.locals.htmlFiles = []
  res.app.locals.htmlFiles.forEach(function (html) {
    var splitted = html.split('/')
    var name = splitted[splitted.length-1].split('.')[0]
    res.locals.htmlFiles.push({
      name: name,
      file: fs.readFileSync(path.resolve(html), 'utf8')
    })
  })
  res.render('modules/core/server/views/index', {
    user: req.user || null
  })
}
```

### modules/core/server/views/layout.server.view.html
```
{% for html in htmlFiles %}
  <script type="text/ng-template" id="{{html.name}}">{{html.file | safe}}</script>
{% endfor %}
```

then just use it like that for a view named app.client.view.html:
```
$stateProvider.
  state('app', {
    url: '/:lang',
    templateUrl: 'app',
    controller: 'AppController'
  })
```

I saw a bunch of discussions about versions, PR and stuff so... tell me what you think and what to do :)

trainerbill:
@andreafdaf I too would like to use templateCache.  I have a PR to return the /client to the include URL's

https://github.com/meanjs/mean/pull/758

I want to get that merged in and then work on a templating solution as well.  Lets start by +1 that PR to get it in and I would work with you on getting templateCache introduced.  It would increase performance as well.

andreafdaf:
Hey @trainerbill ! I'm sorry but I cannot see how is this PR related to what I said... once we introduce templateCache the above way there is no need to worry about the path for the views since they get all "stuffed" into the main HTML file and called by their id when used inside angular

If someone says that PR's are welcome I've got it ready already :+1: 
@lirantal ? @rhutchison ? @ilanbiala ? anyone? :)

codydaig:
[...]

PR #899, opened 2015-09-08
embed views and load them using templateCache

andreafdaf:
[this is where I proposed to embed views and load them using templateCache](https://github.com/meanjs/mean/issues/769)

### DIFF
--- a/config/assets/default.js
+++ b/config/assets/default.js
@@ -34,7 +34,7 @@ module.exports = {
       'modules/*/client/*.js',
       'modules/*/client/**/*.js'
     ],
-    views: ['modules/*/client/views/**/*.html'],
+    views: ['modules/*/client/**/*client.view.html'],
     templates: ['build/templates.js']
   },
   server: {
--- a/config/assets/default.js
+++ b/config/assets/default.js
@@ -34,7 +34,7 @@ module.exports = {
       'modules/*/client/*.js',
       'modules/*/client/**/*.js'
     ],
-    views: ['modules/*/client/**/*client.view.html'],
+    views: ['modules/*/client/views/**/*.html'],
     templates: ['build/templates.js']
   },
   server: {
--- a/modules/core/server/controllers/core.server.controller.js
+++ b/modules/core/server/controllers/core.server.controller.js
@@ -4,24 +4,24 @@ var fs = require('fs'),
   path = require('path'),
   async = require('async');
 
-function readViewFromDisk (viewPath, callback) {
-  var splittedPath = viewPath.split('/');
-  var viewName = splittedPath[splittedPath.length-1].split('.')[0];
-  fs.readFile(path.resolve(viewPath), 'utf8', function (err, file) {
-    if(err) {
-      callback(err);
-    } else {
-      callback(null, {
-        name: viewName,
-        file: file
-      });
-    }
-  });
-}
 /**
  * Render the main application page
  */
 exports.renderIndex = function (req, res) {
+  function readViewFromDisk (viewPath, callback) {
+    var splittedPath = viewPath.split('/');
+    var viewName = splittedPath[splittedPath.length-1].split('.')[0];
+    fs.readFile(path.resolve(viewPath), 'utf8', function (err, file) {
+      if(err) {
+        callback(err);
+      } else {
+        callback(null, {
+          name: viewName,
+          file: file
+        });
+      }
+    });
+  }
   async.concat(res.app.locals.htmlFiles, readViewFromDisk, function (err, embeddableViews) {
     if(err)
       return res.status(500).send(err);
--- a/modules/core/server/controllers/core.server.controller.js
+++ b/modules/core/server/controllers/core.server.controller.js
@@ -10,25 +10,27 @@ var fs = require('fs'),
 exports.renderIndex = function (req, res) {
   function readViewFromDisk (viewPath, callback) {
     var splittedPath = viewPath.split('/');
-    var viewName = splittedPath[splittedPath.length-1].split('.')[0];
+    var viewId = splittedPath[splittedPath.length-1].split('.')[0];
     fs.readFile(path.resolve(viewPath), 'utf8', function (err, file) {
       if(err) {
         callback(err);
       } else {
         callback(null, {
-          name: viewName,
-          file: file
+          id: viewId,
+          content: file
         });
       }
     });
   }
-  async.concat(res.app.locals.htmlFiles, readViewFromDisk, function (err, embeddableViews) {
-    if(err)
-      return res.status(500).send(err);
-    res.locals.htmlFiles = embeddableViews;
-    res.render('modules/core/server/views/index', {
-      user: req.user || null
-    });
+  async.map(res.app.locals.htmlFiles, readViewFromDisk, function (err, embeddableViews) {
+    if(err) {
+      res.status(500).send(err);
+    } else {
+      res.locals.htmlFiles = embeddableViews;
+      res.render('modules/core/server/views/index', {
+        user: req.user || null
+      });
+    }
   });
 };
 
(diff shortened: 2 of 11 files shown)


mleanos:
@andreafdaf Are the tests running successfully on your local dev? It appears that the client-side tests are failing.

config/assets/default.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
       'modules/*/client/*.js',
       'modules/*/client/**/*.js'
     ],
-    views: ['modules/*/client/views/**/*.html'],
+    views: ['modules/*/client/**/*client.view.html'],
```
mleanos: Why was this changed?

mleanos:
@andreafdaf I don't know how this implementation of template cache is working. Could you explain it a bit? 

Also, could the client-side tests be failing because the route configurations have changed? Also, the change of the assets client views definition might be related.

config/assets/default.js
andreafdaf: this is to make it a little more flexible, for example... if I have something like "...moduleName/client/directives" and I put templates and js files under the "directives" folder with the old rule they would not be found

of course one can put js from directives under js and html under views, but it makes or doesn't make sense based on just a programmer point of view... so that's why it was changed

andreafdaf:
hi @mleanos sorry but I didn't tested (hehe ._.) since it was working on browser and is a mod taken from a project already running in production so... I guess that the tests must be rewritten in some small part, tests fail in all browsers but views are correctly loaded and cached so it's just a test inner thing (template IDs probably), I guess I found a solution but have no time to work on it right now

anyway it works like that:
- suppose we have an application that grows and grows so we start having a lot of small html files here and there, some really small, and maybe we need lots of them to load a single page and the page might start loading like a puzzle (noticeable on old PCs and slow connections)
- one might say "convert all views to JS templates, blob everything together upon build and add it as a module" but if you're working with someone that just lives on the HTML/CSS side of the world it gets really difficult to make them understand this and wait for a task runner to do all this work before they can see if the new text color is nice
- so there is an easy way: we simply get all views and embed them inside the main HTML page inside script type="text/ng-template" tags and give each one an ID (this ID what is causing the test to fail btw)
- now we can continue to write down our HTML templates as normal HTML files and worry about nothing because when a user makes a request to index all HTML templates are going to be bundled as written above and will all be available to the angular app without having to make a request to the server

some performance gain and nothing changes in development (just calling templates by ID instead of by URL, that's it)

andreafdaf:
ok tests now pass, it was a simple mod (memo to myself: read contributing guidelines more carefully)
I can try to make a couple of things more straightforward (like making the view pattern match every HTML file, or using the whole filename as ID)

anyway... this PR is just a starting point to discuss about this feature, this same result can be achieved in multiple ways so now it's up to the community to decide if it fits or not, must be changed or what :v: 

btw, I'm sorry to see that "Coverage decreased (-0.4%) to 51.842%"... how can I improve it?

config/assets/default.js
mleanos: I'm not sure what the best practice in the cases of where to put the html templates for directives.. but I usually put them in my *views* folder. Perhaps, someone else has some insight into that. But generally, I'd be weary of making changes to a core feature like the assets locations.

mleanos:
@andreafdaf I'm not exactly sure.

We could use some more eyes on this one. @codydaig @ilanbiala @lirantal @rhutchison @trainerbill @simison

config/assets/default.js
andreafdaf: I guess you're right... maybe 'modules/*/client/**/*.html' like is done for 'modules/*/client/**/*.js' would be the most flexible one (and wouldn't break compatibility with code coming from an earlier version)

modules/core/server/controllers/core.server.controller.js
```diff
@@ -1,11 +1,34 @@
...
+        callback(err);
+      } else {
+        callback(null, {
+          name: viewName,
+          file: file
+        });
+      }
+    });
+  }
+  async.concat(res.app.locals.htmlFiles, readViewFromDisk, function (err, embeddableViews) {
```
codydaig: why do we need to concat the files if they are being loaded in individually?

modules/core/server/controllers/core.server.controller.js
```diff
@@ -1,11 +1,34 @@
...
+    if(err)
+      return res.status(500).send(err);
```
codydaig: please put brackets around this line otherwise it looks out of place

codydaig:
@andreafdaf Im not completely familiar with ng-template so i out a couple inline questions above. Ill have some time later today to test and play. I'll touch base afterwards and give you more through feedback.

modules/core/server/controllers/core.server.controller.js
andreafdaf: you've got your point :+1:

modules/core/server/controllers/core.server.controller.js
andreafdaf: sorry, my bad, I've tried to follow your syntax, missed this one :)

codydaig:
@andreafdaf So what's the difference between your approach and an approach using a gulp task like this: https://www.npmjs.com/package/gulp-templatecache?

That gulp task is currently implemented in our gulp file so I was wondering what the difference was, and what the benefits of one over the other are?

andreafdaf:
@codydaig hi! the main difference is... you don't have to use gulp! for example there is also https://www.npmjs.com/package/grunt-angular-templates but it is not into this project, but gulp support it's still not finished as I read...

coding it inside the express app this way, as soon as everyone agrees that "on mean views are called like that and placed right here", a developer new to mean don't have to do anything, just follow some naming conventions and templateCache will work out of the box with no external dependencies

also this bundling happens every time one calls "index" making life easier on development if you don't use file watchers (this can be changed, reading the files when express app starts and embedding always those same files... or maybe it can be configurable)

so it does the same thing, except it doesn't rely on any particular task runner... I don't know about you but I don't like to force users to use a task runner, use a task runner over another, having to use both grunt and gulp or having to use a task runner at all to do something that comes even simpler when actually coded

ps: just to say something else, not that one request changes things, but using the task runner approach there is one extra request, the one to load the templates module... this way the templates are inside the main HTML so no requests

codydaig:
@andreafdaf So the issue with putting everything inside the index.html page is that the index.html page cannot be cached because it injects the user object directly into the page. But compiling the templates into a file that get's put into the /public folder allows for production environments to cache the html files, as well as allow nginx (or similar) to server the files instead of relying on express, which is usually he preferred method in production versus relying on express.

andreafdaf:
@codydaig I was writing something else but then you comment popped up! so what about also using https://www.npmjs.com/package/grunt-angular-templates , so that developers can have this same functionality no matter which task runner they choose to use?

codydaig:
@andreafdaf Currently we're switching over to gulp so, I think it's fine leaving support for this on the gulp side. I apologize for the initial link, the this is the one we're actually using. I forgot we did a name swap in the plugins. 
https://github.com/miickel/gulp-angular-templatecache

andreafdaf:
@codydaig dropping grunt? good news! I thought that (in this project) gulp was behind grunt, but you are clearly showing me that there are some tasks that are used only in gulp

so I guess we can close this!

codydaig:
@andreafdaf Gulp isn't fully ahead of grunt yet. But in the near future it will be.

rhutchison:
Ref: https://github.com/meanjs/mean/pull/853

simison:
> (just calling templates by ID instead of by URL, that's it)

 @andreafdaf I've felt it's okay to use the file path as an ID even with templateCache, perhaps just without .html at the end to indicate it's an inline template. Makes it easier to find them.

andreafdaf:
@simison since templateCache is only used in production you were totally right (imho), if development needs the path to find the views (so we put the whole path when we call a templateUrl) then in production the IDs from templateCache should be the same as the path if we don't want to write more code to make this distinction happen (when in production call this ID, when in development call this path)

PR #901, opened 2015-09-10
Fix redirection to previous state after required authentication

igorauad:
`$stateChangeSuccess` event should be signaled in order to allow "recording" of previous state. 

This part of the implementation was missing, I just realized that.

### DIFF
--- a/modules/core/client/app/init.js
+++ b/modules/core/client/app/init.js
@@ -30,7 +30,9 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
         if (Authentication.user !== undefined && typeof Authentication.user === 'object') {
           $state.go('forbidden');
         } else {
-          $state.go('authentication.signin');
+          $state.go('authentication.signin').then(function () {
+            storePreviousState(toState, toParams);
+          });
         }
       }
     }
@@ -38,14 +40,20 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
 
   // Record previous state
   $rootScope.$on('$stateChangeSuccess', function (event, toState, toParams, fromState, fromParams) {
-    if (!fromState.data || !fromState.data.ignoreState) {
+    storePreviousState(fromState, fromParams);
+  });
+
+  // Store previous state
+  function storePreviousState(state, params) {
+    // only store this state if it shouldn't be ignored 
+    if (!state.data || !state.data.ignoreState) {
       $state.previous = {
-        state: fromState,
-        params: fromParams,
-        href: $state.href(fromState, fromParams)
+        state: state,
+        params: params,
+        href: $state.href(state, params)
       };
     }
-  });
+  }
 });
 
 //Then define the init function for starting up the application
--- a/modules/users/tests/client/authentication.client.controller.tests.js
+++ b/modules/users/tests/client/authentication.client.controller.tests.js
@@ -8,6 +8,7 @@
       scope,
       $httpBackend,
       $stateParams,
+      $state,
       $location;
 
     beforeEach(function () {
@@ -58,6 +59,32 @@
           expect(scope.authentication.user).toEqual('Fred');
           expect($location.url()).toEqual('/');
         });
+
+        it('should be redirected to previous state after successful login',
+          inject(function (_$state_) {
+            $state = _$state_;
+            $state.previous = {
+              state: {
+                name: 'articles.create'
+              },
+              params: {},
+              href: '/articles/create'
+            };
+
+            spyOn($state, 'transitionTo');
+            spyOn($state, 'go');
+
+            // Test expected GET request
+            $httpBackend.when('POST', '/api/auth/signin').respond(200, 'Fred');
+
+            scope.signin(true);
+            $httpBackend.flush();
+
+            // Test scope value
+            expect($state.go).toHaveBeenCalled();
+            expect($state.go).toHaveBeenCalledWith($state.previous.state.name, $state.previous.params);
+
+          }));
 
         it('should fail to log in with nothing', function () {
           // Test expected POST request


ilanbiala:
@igorauad did we not have tests for this?

igorauad:
Not yet, @ilanbiala . I don't think so.

ilanbiala:
@igorauad can you also add tests to this PR then?

igorauad:
Sure, @ilanbiala . Just give me some time. I will probably find time tomorrow.

rhutchison:
@igorauad this solution needs some work.  I am not on board with firing the success event, when the state was not successful (it was interrupted).  There should we a better way to preserve the previous (interrupted state) without an event.

The reason I am against reusing the event is because you are not aware of any  other listeners, and what they are doing - you are making the assumption that this will only trigger the listener to record the previous state.

mleanos:
@igorauad I agree with @rhutchison that we don't want to fire the $stateChangeSuccess again. This would end up firing it twice in that cycle. Among the other reasons that Ryan mentioned.

I pulled down your branch locally, and made modifications as an alternative to this solution. Let me know if you think this will suffice? 

https://github.com/mleanos/mean/commit/f86c94814dcf9da5510cee2254bf615a2fa672a8

rhutchison:
@mleanos the commit looks good.

@igorauad can you review his solution and let us know if you think this is a better approach?  If you agree, can you update your PR to reflect those changes, plus any tests you can put around it.

modules/core/client/app/init.js
```diff
@@ -30,7 +30,9 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
         if (Authentication.user !== undefined && typeof Authentication.user === 'object') {
           $state.go('forbidden');
         } else {
-          $state.go('authentication.signin');
+          $state.go('authentication.signin').then(function () {
```
igorauad: We could also use something like `$state.go('authentication.signin', {warn: 'Please, sign in first.'})` here. Then, print that in alert in the `signin` template. What do you think?

mleanos: @igorauad What would that accomplish?

igorauad: @mleanos just UX, for me it seems better to let the user know why the signin page is being displayed rather than the requested one.

ilanbiala: I think most users know nowadays, no?

igorauad: Yes @ilanbiala Ok, in fact that depends on the audience and the application, so we shouldn't worry about it here.

codydaig:
LGTM

ilanbiala:
@igorauad any tests you can throw in?

igorauad:
Yes @ilanbiala , I just need to learn a bit about tests, because I'm not experienced on this yet :) 

These tests should be included in `authentication.client.controller.tests.js`, correct?

If anybody want to contribute first, feel free.

ilanbiala:
@igorauad yeah probably, seems most appropriate there. Add whatever you can and we'll go from there.

igorauad:
@ilanbiala just pushed my first attempt. Not yet successful :( any hints?

mleanos:
@igorauad Do the tests pass locally for you?

igorauad:
@mleanos nop, I'm on the learning curve yet. Just pushed following the advice from @ilanbiala: "Add whatever you can and we'll go from there."

mleanos:
@igorauad Ok, that's fine. I'll try on my end.. I'm not proficient with the Karma tests yet either. Maybe together we'll both learn from this :)

igorauad:
Cool @mleanos :)

ilanbiala:
Error for reference:

```
Expected spy transitionTo to have been called with [ Object({ state: Object({ name: '', url: '^', views: null, abstract: true }), params: Object({  }), href: null }) ] but it was never called.
	    at /home/travis/build/meanjs/mean/modules/users/tests/client/authentication.client.controller.tests.js:82
```

@igorauad I think either the spy isn't on the right thing or scope.signin isn't calling, but I'll look into it more tomorrow when I have time.

modules/users/tests/client/authentication.client.controller.tests.js
```diff
@@ -59,6 +60,28 @@
...
+              params: {},
+              href: '/articles/create'
+            };
+
+            // Test expected GET request
+            $httpBackend.when('POST', '/api/auth/signin').respond(200, 'Fred');
+
+            scope.signin(true);
+            $httpBackend.flush();
+            spyOn($state, 'transitionTo');
```
mleanos: This should be set before `scope.signin(true)` (line 77). However, even after that modification, the test doesn't pass. The actuals returned from the below `expect` don't match up with the previous state defined in this test. Upon further investigation, I logged `$state.previous` to the console in the signin method of the AuthenticationController & it was equal to the previous state defined here in this test.

Anyone else wanna take a shot at this? I'll keep working on it, but probably could use a fresh set of eyes.

mleanos:
@ilanbiala @igorauad I was able to get the test to pass.. It appears that both the `$state.transitionTo` & `$state.go` must be spied on.. I can't tell you why, but after spending quite a bit of time fiddling with this test, this is the only way I got it to work. 

```
it('should be redirected to previous state after successful login',
  inject(function (_$state_) {
    $state = _$state_;
    $state.previous = {
      state: {
        name: 'articles.create'
        },
        params: {},
        href: '/articles/create'
     };

     spyOn($state, 'transitionTo');
     spyOn($state, 'go');

     // Test expected GET request
     $httpBackend.when('POST', '/api/auth/signin').respond(200, 'Fred');
          
     scope.signin(true);
     $httpBackend.flush();

     // Test scope value
     expect($state.go).toHaveBeenCalledWith($state.previous.state.name, $state.previous.params);
}));
```

ilanbiala:
@igorauad can you see if that solves it for you and if so push it?

igorauad:
@ilanbiala just pushed. Yes, it solves. If somebody understand exactly why, that would be interesting to know. Thank you very much @mleanos

mleanos:
@igorauad Looks like timeout errors, during the build. Hopefully, it was due to a hiccup with Mongo. 

@ilanbiala @lirantal Do you have the ability to re-run the CI builds for this PR?

mleanos:
@igorauad It's nice that at least the coveralls increased :)

ilanbiala:
`MongoError: topology was destroyed` and `timeout error`. Not sure why.

lirantal:
Yes, I re-ran the build now, let's see how it goes.

lirantal:
CI LGTM now.

igorauad:
Is there anything else we should do on this?

codydaig:
@igorauad Squash, and it should be good.

mleanos:
LGTM. Just needs squashing.

igorauad:
Done, @codydaig @mleanos

mleanos:
@igorauad Thanks

igorauad:
Thanks, @mleanos

PR #907, opened 2015-09-13
[bug] Missing owasp dependency

mleanos:
This adds the missing owasp-password-strength-test front-end package to
config/assets/production

### DIFF
--- a/config/assets/production.js
+++ b/config/assets/production.js
@@ -15,7 +15,8 @@ module.exports = {
         'public/lib/angular-ui-router/release/angular-ui-router.min.js',
         'public/lib/angular-ui-utils/ui-utils.min.js',
         'public/lib/angular-bootstrap/ui-bootstrap-tpls.min.js',
-        'public/lib/angular-file-upload/angular-file-upload.min.js'
+        'public/lib/angular-file-upload/angular-file-upload.min.js',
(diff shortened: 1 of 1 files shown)


codydaig:
@mleanos It is here, is it not? 
https://github.com/meanjs/mean/blob/master/bower.json#L16

jloveland:
This is already provided by `config/master/assets/default.js`:
https://github.com/meanjs/mean/blob/master/config/assets/default.js#L19

Only if owasp-password-strength-test had a minified lib, we would add it to the production.js right?

codydaig:
Sorry. I was looking at the wrong file. @mleanos is correct. The arrays don't merge in the config. So, for it to be included in the production environment, it must be listed in the production env file. 

LGTM

ilanbiala:
@codydaig @jloveland @mleanos can we get #907 and #909 in one PR because they both really have to do with the same thing and it will keep the commit history cleaner.

codydaig:
@ilanbiala Combined in #909.

mleanos:
@ilanbiala Yea, I'm fine with this being combined to #909

mleanos:
@ilanbiala @codydaig Thanks. Didn't see this was closed, before my last comment.

PR #909, opened 2015-09-13
[bug] Seed Password's strength [Closes #908]

### LINKED ISSUE #908: Running with User Seed is Failed
Warning:  Database seeding is turned on
--
MEAN.JS - Development Environment
Environment:			development
Port:				3000
Database:				mongodb://localhost/jydz-dev
App version:			0.4.1
MEAN.JS version:			0.4.1
--
Failed to add local user
Failed to add local admin

codydaig:


lirantal:
Yeah, looks good, and we missed that with the secure password.
Good thing I've started a local branch to refactor the seed users functionality so it is actually testable! it's coming up :)

ilanbiala:
@lirantal this should actually close #907 and #908.

lirantal:
Thanks, got it.

blueflagbj:
Bug persists, check to remove seed.js encryption like this：
................
 //Add Local User
  User.find({username: 'user'}).remove(function () {
  //  var password = crypto.randomBytes(64).toString('hex').slice(1, 20);
  //  seedUser.password = password;
    var user = new User(seedUser);
    // Then save the user
    user.save(function (err) {
.................

mleanos:
I can confirm this. The password being used for the seeded Users is the randomly generated hex. 

I think we have a couple options to fix.. 

1) Continue to randomly generate the password, but *force* the generated hex to include the required upperCase, lowercase, number, and special character by appending to it. 

`var password = crypto.randomBytes(64).toString('hex').slice(1, 20) + 'aX5$';`

My concern here would be the possibility of 3 or more consecutive characters (which isn't allowed with the password requirements) being generated by crypto. With a little bit of *elbow grease* we can inspect the generated password and make sure it adheres to the password requirements. It might not be pretty though :) 

2) Use the password defined in the seedAdmin & seedUser objects. Here we have the concern that originally prompted the use of crypto; we wouldn't want to reuse the same *hardcoded* password for reasons of security.

mleanos:
@lirantal I know you mentioned you were working on making this database seeding more testable. Have you made any progress on that? What are you thoughts on a possible fix for this?

mleanos:
@blueflagbj Until a fix is merged in, a temporary solution would be to use the password's defined in the User objects on the seed configuration, rather than the randomly generated value.

lirantal:
@blueflagbj posting some code (https://github.com/meanjs/mean/pull/909#issuecomment-140653099) is not equal to reporting a bug.
Can I get an explanation of what's the actual problem?

mleanos:
@lirantal The issue is in db seed configuration. The password being used for the seeded users, is a generated hex string that doesn't pass the owasp test when the model attempts to save. Thus, the users aren't being created.

I submitted #921

lirantal:
Yeah I can read the code too :)
It was a general remark - we usually have little time so I'm only asking to be more elaborate and descriptive and not let us guess.

I'll reply with my comments on #921

mleanos:
Ok :) I agree with that. Thanks.

PR #910, opened 2015-09-13
Hiding the Password Strength progress when the field is empty

jloveland:
Renaming strength meter to requirements meter, hiding when password field is empty, and refactoring directives to use $validators, adding tests
bad pass
getting better
success

### DIFF
--- a/modules/users/client/directives/password-validator.client.directive.js
+++ b/modules/users/client/directives/password-validator.client.directive.js
@@ -11,7 +11,7 @@ angular.module('users')
             var result = PasswordValidator.getResult(password);
             var requirementsIdx = 0;
 
-            // requirements Meter - visual indicator for users
+            // Requirements Meter - visual indicator for users
             var requirementsMeter = [
               { color: 'danger', progress: '20' },
               { color: 'warning', progress: '40' },
@@ -33,6 +33,7 @@ angular.module('users')
               status = false;
             } else {
               scope.popoverMsg = '';
+              scope.passwordErrors = [];
               status = true;
             }
           }
--- a/modules/users/client/directives/password-validator.client.directive.js
+++ b/modules/users/client/directives/password-validator.client.directive.js
@@ -4,39 +4,40 @@ angular.module('users')
   .directive('passwordValidator', ['PasswordValidator', function(PasswordValidator) {
     return {
       require: 'ngModel',
-      link: function(scope, element, attrs, modelCtrl) {
-        modelCtrl.$parsers.unshift(function (password) {
-          var result = PasswordValidator.getResult(password);
-          var strengthIdx = 0;
+      link: function(scope, element, attrs, ngModel) {
+        ngModel.$validators.requirements = function (password) {
+          var status = true;
+          if (password) {
+            var result = PasswordValidator.getResult(password);
+            var requirementsIdx = 0;
 
-          // Strength Meter - visual indicator for users
-          var strengthMeter = [
-            { color: 'danger', progress: '20' },
-            { color: 'warning', progress: '40' },
-            { color: 'info', progress: '60' },
-            { color: 'primary', progress: '80' },
-            { color: 'success', progress: '100' }
-          ];
-          var strengthMax = strengthMeter.length;
+            // requirements Meter - visual indicator for users
+            var requirementsMeter = [
+              { color: 'danger', progress: '20' },
+              { color: 'warning', progress: '40' },
+              { color: 'info', progress: '60' },
+              { color: 'primary', progress: '80' },
+              { color: 'success', progress: '100' }
+            ];
 
-          if (result.errors.length < strengthMeter.length) {
-            strengthIdx = strengthMeter.length - result.errors.length - 1;
+            if (result.errors.length < requirementsMeter.length) {
+              requirementsIdx = requirementsMeter.length - result.errors.length - 1;
+            }
+
+            scope.requirementsColor = requirementsMeter[requirementsIdx].color;
+            scope.requirementsProgress = requirementsMeter[requirementsIdx].progress;
+
+            if (result.errors.length) {
+              scope.popoverMsg = PasswordValidator.getPopoverMsg();
+              scope.passwordErrors = result.errors;
+              status = false;
+            } else {
+              scope.popoverMsg = '';
+              status = true;
+            }
           }
-
-          scope.strengthColor = strengthMeter[strengthIdx].color;
-          scope.strengthProgress = strengthMeter[strengthIdx].progress;
-
-          if (result.errors.length) {
-            scope.popoverMsg = PasswordValidator.getPopoverMsg();
-            scope.passwordErrors = result.errors;
-            modelCtrl.$setValidity('strength', false);
-            return undefined;
-          } else {
-            scope.popoverMsg = '';
-            modelCtrl.$setValidity('strength', true);
-            return password;
-          }
-        });
+          return status;
+        };
       }
     };
 }]);
--- a/modules/users/client/directives/password-verify.client.directive.js
+++ b/modules/users/client/directives/password-verify.client.directive.js
@@ -7,27 +7,22 @@ angular.module('users')
       scope: {
         passwordVerify: '='
       },
-      link: function(scope, element, attrs, modelCtrl) {
+      link: function(scope, element, attrs, ngModel) {
+        var status = true;
         scope.$watch(function() {
           var combined;
-          if (scope.passwordVerify || modelCtrl.$viewValue) {
-            combined = scope.passwordVerify + '_' + modelCtrl.$viewValue;
+          if (scope.passwordVerify || ngModel) {
+            combined = scope.passwordVerify + '_' + ngModel;
           }
           return combined;
         }, function(value) {
           if (value) {
-            modelCtrl.$parsers.unshift(function(viewValue) {
+            ngModel.$validators.passwordVerify = function (password) {
               var origin = scope.passwordVerify;
-              if (origin !== viewValue) {
-                modelCtrl.$setValidity('passwordVerify', false);
-                return undefined;
-              } else {
-                modelCtrl.$setValidity('passwordVerify', true);
-                return viewValue;
-              }
-            });
+              return (origin !== password) ? false : true;
+            };
           }
         });
-     }
+      }
     };
 });
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -258,6 +258,26 @@ module.exports = function (grunt) {
     });
   });
 
+  // Drops the MongoDB database, used in e2e testing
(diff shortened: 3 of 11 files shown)


jloveland:
@mleanos

mleanos:
@jloveland This is what I had originally changed to hide the indicator. However, I'd rather be checking the `passwordDetails.newPassword` model. Could you explain what undesired behavior exists, if you return `password` from here? https://github.com/meanjs/mean/blob/master/modules/users/client/directives/password-validator.client.directive.js#L33 
I ended up returning `password` & I couldn't see any negative side effects. I don't think for the purposes of this directive, that it should set the ngModel to be `undefined`. It could have a negative impact over something else relying on the model.

It was brought up to me by someone testing my app, that the wording "Password Strength" is misleading. This isn't really a standard *strength* indicator; it's a requirement meter. How do you feel about changing the wording? Maybe something along the lines of "Minimum Strength Requirement" ?

jloveland:
@mleanos I agree. I was reviewing more docs and see that I can use $validators instead of $parse. This will leave the model in tact as the user types. I will push an update soon.

Also, do you think **Password Requirements** above the progress bar would be clear enough instead of **Password Strength**?

jloveland:
depending on whether we decide to change the name from **Password Strength** to **Password Requirements**, I can change the name of variables to match the language...

mleanos:
@jloveland I think **Password Requirements** would be much more clear.

As for the model, try merely changing the returned value from `undefined` to `password` and play around with it. I did this & couldn't see anything else that behaved differently, other than the model being intact. You might see something that I don't. The `required` validator that is already implemented on that field should be sufficient.

jloveland:
@mleanos I made some updates, renaming the strength meter to requirements meter. I am also hiding the requirements meter by checking the required validator only.

I am convinced that moving from the $parsers pipeline to the $validators pipeline is the best choice with Angular 1.3. see this blog for reference:
http://blog.thoughtram.io/angularjs/2015/01/11/exploring-angular-1.3-validators-pipeline.html

modules/users/client/directives/password-validator.client.directive.js
```diff
@@ -4,39 +4,41 @@ angular.module('users')
...
-          var strengthMax = strengthMeter.length;
+            // requirements Meter - visual indicator for users
+            var requirementsMeter = [
+              { color: "danger", progress: "20" },
+              { color: "warning", progress: "40"},
+              { color: "info", progress: "60"},
+              { color: "primary", progress: "80"},
+              { color: "success", progress: "100"}
+            ];
+            var requirementsMax = requirementsMeter.length;
```
mleanos: where is this being used? I don't see it.

mleanos:
@jloveland Looks great! I'm still concerned as to why the the ng-model for the password remains `undefined` until the validators pass... Can you explain that a bit to me? I think I'm missing something very fundamental.

My preference would be to not affect the ng-model & leave it intact, but for these purposes it's probably not a huge deal.

lirantal:
@ilanbiala I added you to review this one

ilanbiala:
@lirantal are you okay with the change? If so, then @jloveland please add tests. I'd prefer from now on to prevent bugs and issues that we require tests with any PRs, because it already helps to make reviewing easier and already shows some validity of the code.

modules/users/client/directives/password-verify.client.directive.js
```diff
@@ -7,25 +7,24 @@ angular.module('users')
...
           return combined;
         }, function(value) {
           if (value) {
-            modelCtrl.$parsers.unshift(function(viewValue) {
+            ngModel.$validators.passwordVerify = function (password) {
               var origin = scope.passwordVerify;
-              if (origin !== viewValue) {
-                modelCtrl.$setValidity("passwordVerify", false);
-                return undefined;
+              if (origin !== password) {
```
ilanbiala: Can we just clean this up to be a ternary op?

ilanbiala:
@jloveland another thing, I remember @rhutchison doing something with ng-messages. @rhutchison was that cleaner?

lirantal:
I'm ok with this.
I'll also add that if you are submitting a PR please add screenshots, which is extra important when you are doing frontend PRs. If I did that for my server-side PRs then you should have no problems doing it for frontend.

modules/users/client/directives/password-validator.client.directive.js
jloveland: good catch! it's not used, I will remove.

modules/users/client/directives/password-verify.client.directive.js
jloveland: sure!

jloveland:
@mleanos, `ngModel = undefined` is the default behavior by design for the `$validators` pipeline. Please review this issue: https://github.com/angular/angular.js/issues/1412

jloveland:
@ilanbiala and @rhutchison I added the errors returned from password-validator within the ng-messages block. for example:
https://github.com/meanjs/mean/blob/master/modules/users/client/views/authentication/signup.client.view.html#L40
do you see anything wrong with this approach?

mleanos:
@jloveland Thanks for the reference. I think for the purposes of this exact implementation (used for Password validation), we can go with it. 

However, if we implement `$validators` somewhere else in the project, we should pay attention to this behavior. I can see some cases where the ngModel value would need to be retained even when not valid. 

This LGTM, and is ready to be merged from my perspective.

modules/users/client/directives/password-verify.client.directive.js
```diff
@@ -7,25 +7,24 @@ angular.module('users')
...
               } else {
-                modelCtrl.$setValidity("passwordVerify", true);
-                return viewValue;
+                return true;
               }
-            });
+            };
           }
         });
      }
```
mleanos: @jloveland Slight indentation issue.

mleanos:
For reference.. If the behavior of the ngModel being set to `undefined` is undesired, we could use the `allowInvalid` ng-model-options setting. 
```
ng-model-options="{ allowInvalid: true }"
```
This will allow the `ng-model` to retain the value, while still marking the model as invalid from the `$validator`. Again, I don't think this option is needed for this particular case. But wanted to drop this here for anyone coming along later.

jloveland:
@ilanbiala I am trying to develop karma and e2e tests for this. I am seeing this error with protractor: `Error: Angular could not be found on the page http://localhost:3000/authentication/signin : retries looking for angular exceeded`. Is this a bug with mean.js? or am I doing something wrong?

ilanbiala:
@jloveland I think you have something set up incorrectly. In a separate commit, can you add what you have so far for tests?

ilanbiala:
@jloveland is the testing issue fixed with the PR you have open right now?

jloveland:
@ilanbiala e2e testing issues have been resolved. 
I am planning on adding unit tests for the directives to this PR. 
Should I also add e2e tests here? or would it be easier to review if I added those e2e tests to a different PR?

ilanbiala:
@jloveland add all the tests to this PR.

package.json
```diff
@@ -110,6 +110,7 @@
     "karma-ng-html2js-preprocessor": "^0.1.2",
     "karma-phantomjs-launcher": "~0.2.0",
     "load-grunt-tasks": "^3.2.0",
+    "protractor": "git://github.com/angular/protractor.git#master",
```
jloveland: for some reason, this isn't doing the trick with node v.10 even though this PR was thought to fix the issue.
https://github.com/angular/protractor/pull/2591/files
https://travis-ci.org/angular/protractor/jobs/84351567

Anyone have thoughts on this?

ilanbiala:
@jloveland what do you mean?

package.json
jloveland: @ilanbiala I am seeing that travis is failing to run using node v0.10. it looks like the error is due to `node_modules/grunt-protractor-runner/node_modules/protractor/lib/configParser.js:102:29)` which seems to be an embedded protractor version..so, I suppose we need to wait until grunt-protractor-runner upgrades or we can use my [fork](https://github.com/jloveland/grunt-protractor-runner/tree/protractor-parse-error)..not sure what the best option is..

jloveland:
@ilanbiala sorry for all the chatter here..never mind, I was looking at an old build...looks like all tests pass

ilanbiala:
@jloveland just curious to see if any functionality wasn't covered in the tests, do you think you could try adding https://github.com/karma-runner/karma-coverage to Grunt and Gulp in a PR so we can verify?

jloveland:
@ilanbiala I am also curious about that. I can add another PR for karma coverage. 

I was having issues writing unit tests for the directives. For the password-validator and password-verify directives, I am having trouble determining how to mock the services that the directive depends on...also, it doesn't seem to me that the directives are being loaded in the tests. I added e2e tests instead of unit tests because of the challenges I am having with figuring out how to unit test the directive that is dependent on services..

ilanbiala:
@jloveland let's start with a PR for karma/e2e coverage, and then we can see specifically what else needs to be tested and how to rework the tests so they are clean and intuitive.

mleanos:
@jloveland Per our Gitter discussion...

I verified the directives are being loaded with the tests. However, just like you, I was experiencing the `password-validator` $parser wasn't executing. I had to manually execute the $parser function on the model, inside a test like so.. 

`scope.userForm.password.$parsers[0](scope.credentials.password);` - may not be the correct way to handle this, but at least it's a hint at what we need to do. Might want to put this into a function inside the test file for reuse.

I verified that `scope.passwordErrors` was being properly populated based on the validation.

Also, to answer your question about whether or not we should put this into a separate PR. I think so. I think we should put Directive tests into their own files, since they behave differently than the other tests. For instance, we could add a new test file to *modules/users/tests/client* called *password-validator.client.directive.tests.js*

Let me know what you think about this.

jloveland:
Thanks to the huge help from @mleanos, I was able to add tests for the password-validator and passowrd-verify directives.

@ilanbiala I have a PR coming soon with karma test coverage. Does anyone out there want to take on the e2e coverage PR?

mleanos:
@jloveland I was happy to help out! 

The tests look great. However, I did run into an issue with the E2E tests. I got the same error, as the build. 
https://travis-ci.org/meanjs/mean/jobs/85249711#L715

All in all, this looks really good. +8.5 is nothing to sneeze at :) Great job!

jloveland:
@mleanos thanks for reviewing. I am fixing the e2e test now..

mleanos:
Everything is passing now! Thanks for all the hard work you put into this @jloveland Our test coverage is getting there.

Now that there's so many more E2E tests, it's entertaining to sit back & watch them run :) 

LGTM

lirantal:
Good job guys :+1:

jloveland:
I believe this is ready to merge.

codydaig:
LGTM :+1:

PR #911, opened 2015-09-14
Fix typo

simonlinj:
indent and typo fixes


codydaig:
@simonlinj Good catch!

simison:
Wait, how come JS linter didn't notice this?

https://github.com/meanjs/mean/blob/master/.jshintrc#L14

ilanbiala:
@simison not sure exactly how the rules work, maybe it's a regex and not an exact match check?

PR #922, opened 2015-09-18
NodeJS v4 support - fixing the failed build

lirantal:
Adding the required support to properly build the nodejs v4 edition.

Source and credit originated from: https://github.com/lirantal/meanjs/pull/1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,23 @@ node_js:
   - 0.10
   - 0.12
   - 4
+# NodeJS v4 requires gcc 4.8
 env:
-  - NODE_ENV=travis
+  - NODE_ENV=travis CXX="g++-4.8" CC="gcc-4.8"
 matrix:
   allow_failures:
     - node_js: 4
 services:
   - mongodb
+# gcc 4.8 requires ubuntu-toolchain-r-test
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - g++-4.8
+    - gcc-4.8
+    - clang
 before_install:
   - gem update --system
   - gem install sass --version "=3.3.7"
--- a/modules/users/tests/server/user.server.routes.tests.js
+++ b/modules/users/tests/server/user.server.routes.tests.js
@@ -97,7 +97,15 @@ describe('User CRUD tests', function () {
             }
 
             signoutRes.redirect.should.equal(true);
-            signoutRes.text.should.equal('Moved Temporarily. Redirecting to /');
+
+            // NodeJS v4 changed the status code representation so we must check
+            // before asserting, to be comptabile with all node versions.
+            if (process.version.indexOf('v4') === 0) {
+              signoutRes.text.should.equal('Found. Redirecting to /');
+            } else {
+              signoutRes.text.should.equal('Moved Temporarily. Redirecting to /');
+            }
+
             return done();
           });
       });
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "passport-twitter": "^1.0.2",
     "phantomjs": ">=1.9.0",
     "serve-favicon": "^2.3.0",
-    "socket.io": "^1.3.5",
+    "socket.io": "^1.3.7",
     "swig": "^1.4.2",
     "validator": "^3.41.2"
   },
@@ -79,7 +79,7 @@
     "grunt-mocha-test": "~0.12.7",
     "grunt-mocha-istanbul": "^2.4.0",
     "grunt-ng-annotate": "^1.0.1",
-    "grunt-node-inspector": "~0.2.0",
+    "grunt-node-inspector": "~0.3.0",
     "grunt-nodemon": "~0.4.0",
     "grunt-protractor-runner": "^2.0.0",
     "gulp": "^3.9.0",


ilanbiala:
@lirantal 4.0 still failed. Check Travis. Also, if we fix Node 4.0 support, then we should remove the allow_failures rule for it.

lirantal:
yep, still more work to do there

lirantal:
there's an open issue (https://github.com/socketio/socket.io/issues/2228) with socket.io ability to build on node v4 now, I've been trying some magic but so far no luck. Another option for us is to add shrinkwrap (which I think is a good idea so I'll take a look at it) to control our own sub-dependency versions if we need to.

mleanos:
:+1: On skrinkwrap. I was looking at that a while ago, and thought it was a good solution for this project.

vielmetti:
socket.io@1.3.7 is said to have node.js 4.x support. https://twitter.com/SocketIO/status/645755967934697472

vielmetti:
I was also having issues with one of the dev dependencies, grunt-node-inspector, which is at 0.2.0.

lirantal:
yep, saw that and also Edward from the original PR gave me a ping about it.
I'll update the PR soon enough.

.travis.yml
```diff
@@ -4,13 +4,23 @@ node_js:
   - 0.10
   - 0.12
   - 4
+# NodeJS v4 requires gcc 4.8
 env:
-  - NODE_ENV=travis
+  - NODE_ENV=travis CXX="g++-4.8" CC="gcc-4.8"
 matrix:
```
ilanbiala: If we are attempting to get Node 4 passing, we should remove this section.

ilanbiala:
@lirantal any progress on this?

.travis.yml
lirantal: I don't think that's the reason for it, but rather node-inspector which is failing.

lirantal:
I pushed an update to a newer grunt-node-inspector version which seem to be the culprit, let's see how it goes.

ilanbiala:
@lirantal 1 test failed. Look at the bottom https://travis-ci.org/meanjs/mean/jobs/83342174

Also, if we are going to make sure Node 4 builds and works, please remove the allow failure part for it because it shouldn't be there anymore.

lirantal:
Yeah I just noticed that it finished with a fail for a test, which is related apparently to different messages between node/other package versions:
```
1) User CRUD tests should be able to login successfully and logout successfully:
      Uncaught AssertionError: expected 'Found. Redirecting to /' to be 'Moved Temporarily. Redirecting to /'
      + expected - actual
      +"Moved Temporarily. Redirecting to /"
      -"Found. Redirecting to /"
```

I'm not sure if we want to rely 100% now on node v4 so I say let's keep the allow_failures there still.. As you saw there are many core packages like grunt-node-inspector and others that didn't yet even sync with the node v4 version to become compatible.

codydaig:
@lirantal LGTM. I say merge this in, to get people closer to Node 4.0 support. It's been a popular issue lately.

ilanbiala:
@lirantal I'd rather not merge and keep allow failures, it is deceiving.

lirantal:
@ilanbiala what's deceiving? I don't get it. It's not like the project states it is compatible with v4.
We always had `allow_failures` on for v4 and I think it should be kept this way for a while until the ecosystem becomes more aware and compatible with v4.

ilanbiala:
@lirantal most people don't know about allow_failures, and they won't look closely enough to notice that v4 doesn't have to pass for the badge to show passing.

lirantal:
That's why we as moderators should look into that.
You suggest that remove allow_failures and let the build fail entirely if some PR doesn't meet v4 compatibility but that potentially has the possibility of delaying many PRs from being accepted because all the PRs will always fail, and the fix for node v4 might be out of our reach and not so quickly available.

I don't see the point in failing the build entirely for v4 compatibility for the time being.

codydaig:
I don't think it should fail on node v4 until we start to officially support it.

lirantal:
That's what I'm saying... it currently fails on some test as well that we need to fix to make the build pass v4.

bmcorser:
I was getting a bunch of explosions with `npm install` for this repo, but I can to run without erroring if I run `npm install grunt-node-inspector` beforehand as in https://github.com/node-inspector/v8-debug/issues/7#issuecomment-142390322. That issue suggests that https://github.com/ChrisWren/grunt-node-inspector/ references an out of date `node-inspector` package ... is that what this PR sets out to resolve?

bmcorser:
Hmm seems like v4 isn't quite supported and I should go back to v0.12 ...

lirantal:
@bmcorser indeed.

bmcorser:
@lirantal :+1: ty

lirantal:
```
  1) User CRUD tests should be able to login successfully and logout successfully:
      Uncaught AssertionError: expected 'Found. Redirecting to /' to be 'Moved Temporarily. Redirecting to /'
      + expected - actual
      +"Moved Temporarily. Redirecting to /"
      -"Found. Redirecting to /"
```

The specific test that fails here is because NodeJS v4 changed their status codes for 302 from "Moved Temporarily" to "Found".

I was thinking to support both v4 and prior versions we can check for either of the strings, so I'll update the tests unless you have another idea.

lirantal:
Since the problem here is the actual nodejs version how about we have a case for each version in the test? i.e:
```javascript
if (process.version.indexOf('v4'))
  signoutRes.text.should.equal('Found. Redirecting to /');
else // it's otherwise v0.12, or v0.11 then
  signoutRes.text.should.equal('Moved Temporarily. Redirecting to /');
```

WDYT?

ilanbiala:
@lirantal is it "v4" or "4."? I guess we can do that.

lirantal:
Cool, I'll make the changes and commit so we can see the CI tests pass.

lirantal:
@ilanbiala node v4 is passing just fine now.

.travis.yml
```diff
@@ -25,8 +35,4 @@ notifications:
     on_success: change  # options: [always|never|change] default: always
     on_failure: always  # options: [always|never|change] default: always
     on_start: never     # options: [always|never|change] default: always
-cache:
-  apt: true
-  directories:
-  - node_modules/
-  - public/lib/
+
```
ilanbiala: Why did this get removed?

codydaig:
@lirantal It is failing on the e2e tests, but grunt is continuing and not producing the error too fail the build. Not sure if that is related or not.

ilanbiala:
Here is where things go awry: https://travis-ci.org/meanjs/mean/jobs/83604705#L1555

.travis.yml
lirantal: @ilanbiala I removed it in the beginning when working on this PR for node v4 compatibility
I re-added it again now.

lirantal:
The E2E test fails for both v0.10 and v0.12, it's not related to node v4.
I opened a separate issue to deal with it: https://github.com/meanjs/mean/issues/961

This PR LGTM to continue with.

ilanbiala:
LGTM, squash and merge!

codydaig:
LGTM

Thanks @lirantal!

lirantal:
Squashed and merged.
Thanks guys!

PR #924, opened 2015-09-19
Fixed #923 - password reset links SSL support

### LINKED ISSUE #923: Reset Password URL not using https
when `config.secure.ssl` is `true`, the password reset url should be `https:// ` [here](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js#L62)

lirantal:
Closes #923 - making password reset links work for both http and https configuration.


codydaig:
LGTM

mleanos:
LGTM

jloveland:
:+1:

Issue #926, opened 2015-09-21
Allow for session.secret to be read from local env config


jloveland:
enhancement: put the `sessionSecret` as a `local.js` env config so we can configure it without accidentally committing to code.
`sessionSecret: process.env.SESSION_SECRET || 'MEAN',`
https://github.com/meanjs/mean/blob/master/config/env/default.js#L25

codydaig:
@jloveland Seems like this was resolved wth #927. If not, feel free to reopen! :-D

PR #927, opened 2015-09-21
adding ability to configure session.secret in local env config

jloveland:
fixing #926

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -88,6 +88,28 @@ var validateSecureMode = function (config) {
 };
 
 /**
+ * Validate Session Secret parameter is not set to default in production
+ */
+var validateSessionSecret = function (config, testing) {
+
+  if (process.env.NODE_ENV !== 'production') {
+    return true;
+  }
+
+  if (config.sessionSecret === 'MEAN') {
+    if (!testing) {
+      console.log(chalk.red('+ WARNING: It is strongly recommended that you change sessionSecret config while running in production!'));
+      console.log(chalk.red('  Please add `sessionSecret: process.env.SESSION_SECRET || \'super amazing secret\'` to '));
+      console.log(chalk.red('  `config/env/production.js` or `config/env/local.js`'));
+      console.log();
+    }
+    return false;
+  } else {
+    return true;
+  }
+};
+
+/**
  * Initialize global configuration files
  */
 var initGlobalConfigFolders = function (config, assets) {
@@ -169,7 +191,7 @@ var initGlobalConfig = function () {
   // production or development environment. If test environment is used we don't merge it with local.js
   // to avoid running test suites on a prod/dev environment (which delete records and make modifications)
   if (process.env.NODE_ENV !== 'test') {
-    config = _.merge(config, (fs.existsSync(path.join(process.cwd(), 'config/env/local.js')) && require(path.join(process.cwd(), 'config/env/local.js'))) || {});  
+    config = _.merge(config, (fs.existsSync(path.join(process.cwd(), 'config/env/local.js')) && require(path.join(process.cwd(), 'config/env/local.js'))) || {});
   }
 
   // Initialize global globbed files
@@ -181,9 +203,13 @@ var initGlobalConfig = function () {
   // Validate Secure SSL mode can be used
   validateSecureMode(config);
 
+  // Validate session secret
+  validateSessionSecret(config);
+
   // Expose configuration utilities
   config.utils = {
-    getGlobbedPaths: getGlobbedPaths
+    getGlobbedPaths: getGlobbedPaths,
+    validateSessionSecret: validateSessionSecret
   };
 
   return config;
--- a/modules/core/tests/server/core.server.config.tests.js
+++ b/modules/core/tests/server/core.server.config.tests.js
@@ -10,10 +10,10 @@ var should = require('should'),
   config = require(path.resolve('./config/config')),
   seed = require(path.resolve('./config/lib/seed'));
 
-describe('Configuration tests', function () {
+describe('Configuration Tests:', function () {
   this.timeout(10000);
 
-  describe('Testing default seedDB:', function () {
+  describe('Testing default seedDB', function () {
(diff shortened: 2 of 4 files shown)


config/env/production.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     // Enable mongoose debug mode
     debug: process.env.MONGODB_DEBUG || false
   },
+  sessionSecret: process.env.SESSION_SECRET || 'MEAN',
```
lirantal: This isn't really required because it's already coming from default.js config

config/config.js
```diff
@@ -169,7 +184,7 @@ var initGlobalConfig = function () {
   // production or development environment. If test environment is used we don't merge it with local.js
   // to avoid running test suites on a prod/dev environment (which delete records and make modifications)
   if (process.env.NODE_ENV !== 'test') {
-    config = _.merge(config, (fs.existsSync(path.join(process.cwd(), 'config/env/local.js')) && require(path.join(process.cwd(), 'config/env/local.js'))) || {});  
+    config = _.merge(config, (fs.existsSync(path.join(process.cwd(), 'config/env/local.js')) && require(path.join(process.cwd(), 'config/env/local.js'))) || {});
```
lirantal: the only change here was removing the trailing white space, right?

config/config.js
```diff
@@ -88,6 +88,21 @@ var validateSecureMode = function (config) {
 };
 
 /**
+ * Validate Session Secret parameter is not set to default in production
```
lirantal: Nice, I like this addition.
Can you update the message to be a bit more verbose in telling the user where to change it? i.e `config/env/production.js`

config/config.js
jloveland: @lirantal, that's correct

jloveland:
@lirantal, updated please let me know what you think

mleanos:
LGTM. I also like the addition of `validateSessionSecret` :) Thanks!

lirantal:
LGTM

codydaig:
LGTM.

Thanks! :-D

config/config.js
```diff
@@ -88,6 +88,23 @@ var validateSecureMode = function (config) {
...
+ */
+var validateSessionSecret = function (config) {
```
ilanbiala: Why a variable?

ilanbiala:
Can we get some tests for this?

jloveland:
@ilanbiala I will add some tests.

jloveland:
@ilanbiala added tests

config/config.js
```diff
@@ -88,6 +88,28 @@ var validateSecureMode = function (config) {
 };
 
 /**
+ * Validate Session Secret parameter is not set to default in production
+ */
+var validateSessionSecret = function (config, testing) {
```
ilanbiala: Can you explain why this is a variable and not a named function?

lirantal: @ilanbiala that entire file is handled with variable functions like that: https://github.com/jloveland/mean/blob/local-session-secret/config/config.js , the validateSessionSecret is not an exception.

codydaig:
LGTM with the new tests

config/config.js
ilanbiala: @lirantal why is an anonymous function assigned to a variable better than a named function?

codydaig: @ilanbiala I'm pretty sure that was a "standard" that a lot of projects picked up and ran with for Node. Currently the rest of the file and server side js files are this way, and for consistency should be kept this way until a refactor is done, which IMO is outside the scope of this PR

lirantal: @ilanbiala you're missing the point. I'm not saying it's better or worse, I'm saying that the entire config.js file (https://github.com/jloveland/mean/blob/local-session-secret/config/config.js) makes use of the same style functions, i.e:
```javascript
var getGlobbedPaths = function (globPatterns, excludes) {}
var validateEnvironmentVariable = function () {}
var validateSecureMode = function (config) {}
var validateSessionSecret = function (config, testing) {}
var initGlobalConfigFolders = function (config, assets) {}
and so on
```

I don't think that this PR or others should turn into code style PRs. If we want to change this we can opt for another PR to update to better code styles.

ilanbiala: @lirantal agreed, but let's add this to the list of things to fix up in the formatting issue you are going to create.

lirantal: Cool, np.

lirantal:
@jloveland can you please rebase so we can resolve the conflicts and merge?

jloveland:
this is ready to merge

lirantal:
Coverage increased, great work @jloveland ! :)

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -118,7 +118,43 @@ describe('Configuration tests', function () {
...
+      config.utils.validateSessionSecret(conf, true).should.equal(true);
+      // set env back to test
+      process.env.NODE_ENV = 'test';
+      done();
+    });
+
+    it('should accept default session secret when running in test', function (done) {
+      var conf = { sessionSecret: 'MEAN' };
+      config.utils.validateSessionSecret(conf, true).should.equal(true);
+      done();
```
ilanbiala: This should be `return done()`. However, it doesn't look like there is any asynchronous code being tested, so why are we using done?

ilanbiala: @lirantal @jloveland if this is indeed synchronous (which I think it is), then we should revert this PR and remove the callback because it's unnecessary.

lirantal: @jloveland can you open a new PR to address the `return done()` standard as Ilan suggested?

@ilanbiala I agree with the async observation although for the sake of consistency and later updates to the tests it's ok to keep the done callback. After all, the code is not incorrect, but rather just uses a callback.

jloveland:
@ilanbiala thanks for the review!

@lirantal I can submit a new PR to fix the issue @ilanbiala pointed out.

Issue #934, opened 2015-09-25
Email validation


jloveland:
I am writing some e2e test and found that 123@123 is a valid email according to HTML5 and [RFC 822](https://www.ietf.org/rfc/rfc0822.txt), but we don't support it on the back end.

If we wanted to support this we could set `validator.isEmail('test@test', {require_tld: false})` in `user.server.model`

If we don't want 123@123 to be accepted on the front end, we could check for this on the front end with a pattern like so:
`pattern="[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,63}$"`

reference: http://stackoverflow.com/questions/20573488/why-does-html5-form-validation-allow-emails-without-a-dot

lirantal:
@ilanbiala why 0.4.x milestone? I thought we are avoiding that and always settings all PRs to the next version, unless we know it "needs more time" and so we can set it to to the one after it (i.e: 0.5)

@jloveland I don't specifically mind whether we force the tld option or not.

mleanos:
@jloveland I agree with this statement from the 2nd comment on that SO's accepted answer.. 
>If somebody is entering something unusual as their email, just accept it - chances are they know better than you.

And since it is possible to have an email without a "." in it, shouldn't we accept it as well?

lirantal:
That's why I said I'm ok with it - it is possible for example to utilize meanjs for internal enterprise projects where you may have alias addresses like root@localhost which are valid.

ilanbiala:
@lirantal I figured we should put this up for discussion and it actually is a breaking change, so it should go in 0.5.0.

lirantal:
@ilanbiala what do you refer to by a "breaking change"?

ilanbiala:
We are changing the functionality of email validation, so if users updated to this code as 0.4.2, then their code wouldn't function as before and it might break things, hence a breaking change.

jloveland:
@ilanbiala, how is allowing html5 valid emails (i.e. root@localhost) a breaking change? The application will not stop working, it will allow valid emails. If they don't want to allow root@localhost or test@test, they can remove the `{require_tld: false}`.

ilanbiala:
It's different than before, which breaks the expected functionality. Before we didn't allow emails without a `.` or whatever, and now we do, which changes how things work. That would be a breaking change.

mleanos:
[...]

PR #936, opened 2015-09-25
fixing e2e tasks in grunt and gulp, adding e2e tests for user module

### LINKED ISSUE #929: E2E tests not executing
after running `gulp protractor`
```
Error: Angular could not be found on the page http://localhost:3000/authentication/signin : retries looking for angular exceeded
```
from the angular [docs](https://docs.angularjs.org/guide/e2e-testing):
> ## Caveats
> Protractor does not work out-of-the-box with apps that bootstrap manually using angular.bootstrap. You must use the ng-app directive.

It looks like we manually bootstrap [here](https://github.com/meanjs/mean/blob/master/modules/core/client/app/init.js#L71)
Do we need to add `np-app` somewhere?

jloveland:
fixes #929 and #934

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,10 @@
   "scripts": {
     "start": "grunt",
     "test": "grunt test",
-    "postinstall": "bower install --config.interactive=false"
+    "postinstall": [
+      "bower install --config.interactive=false",
+      "./node_modules/protractor/bin/webdriver-manager update"
+    ]
   },
   "dependencies": {
     "acl": "~0.4.4",
@@ -65,6 +68,8 @@
     "validator": "^3.41.2"
   },
   "devDependencies": {
+    "chromedriver": "^2.19.0",
+    "coveralls": "^2.11.4",
     "grunt-concurrent": "^2.0.0",
     "grunt-contrib-copy": "~0.8.0",
     "grunt-contrib-csslint": "~0.4.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -17,7 +17,10 @@
   "scripts": {
     "start": "grunt",
     "test": "grunt test",
-    "postinstall": "bower install --config.interactive=false"
+    "postinstall": [
+      "bower install --config.interactive=false",
+      "./node_modules/protractor/bin/webdriver-manager update"
```
mleanos: Can this be added as a Grunt task that is coupled with the protractor task? I'm concerned that not all users will want to deal with protractor & the e2e tests at the time of package install.

mleanos: Something like this in the Grunt file would probably suffice.. 
https://github.com/meanjs/mean/blob/master/gulpfile.js#L198

mleanos:
@jloveland I think there's a bit too much going on with this PR. I'm worried it will take some time to review. 

Would it be possible to just add the protractor task to the Grunt file, and fix the existing e2e tests? Perhaps in a separate PR; that way you can keep all this work here. With that approach we would get the bug fix for the Protractor tests out quickly, and then we could start building out the e2e tests.

package.json
jloveland: What do you think about this?

https://www.npmjs.com/package/grunt-protractor-webdriver

mleanos: If we use that package, would we still be required to run the server with our protractor test task? I couldn't tell from the documentation.

Either way, `grunt-protractor-webdriver` looks promising. I'm gonna take a shot at testing it.

jloveland:
@mleanos, I agree, I will separate into different PRs.

package.json
jloveland: @mleanos I didn't like `grunt-protractor-webdriver` after testing it out..take a look at https://github.com/meanjs/mean/pull/939 for the addition of the webdriver update to the grunt file.

PR #937, opened 2015-09-25
SeedDB Refactoring

lirantal:
Refactoring the seeddb logic to work with promises all over due to all the async behavior.
This lays the ground for further work on adding tests for SeedDB functionality.

### DIFF
--- a/config/lib/mongoose.js
+++ b/config/lib/mongoose.js
@@ -9,11 +9,13 @@ var config = require('../config'),
   mongoose = require('mongoose');
 
 // Load the mongoose models
-module.exports.loadModels = function () {
+module.exports.loadModels = function (callback) {
   // Globbing model files
   config.files.server.models.forEach(function (modelPath) {
     require(path.resolve(modelPath));
   });
+
+  if (callback) callback();
 };
 
 // Initialize Mongoose
--- a/config/lib/seed.js
+++ b/config/lib/seed.js
@@ -2,93 +2,146 @@
 
 var mongoose = require('mongoose'),
   chalk = require('chalk'),
-  crypto = require('crypto'),
-  User = mongoose.model('User');
+  crypto = require('crypto');
 
-console.log(chalk.bold.red('Warning:  Database seeding is turned on'));
-
-var seedUser = {
-  username: 'user',
-  password: 'User_Password1!',
-  provider: 'local',
-  email: 'user@localhost.com',
-  firstName: 'User',
-  lastName: 'Local',
-  displayName: 'User Local',
-  roles: ['user']
-};
-
-var seedAdmin = {
-  username: 'admin',
-  password: 'Admin_Password1!',
-  provider: 'local',
-  email: 'admin@localhost.com',
-  firstName: 'Admin',
-  lastName: 'Local',
-  displayName: 'Admin Local',
-  roles: ['user', 'admin']
-};
-
-//If production only seed admin if it does not exist
-if (process.env.NODE_ENV === 'production') {
-  //Add Local Admin
-  User.find({username: seedAdmin.username}, function (err, users) {
-    if (users.length === 0) {
-      var user = new User(seedAdmin);
-
-      // generate a random password and save 
-      User.generateRandomPassphrase()
-      .then(saveUser(user))
-      .catch(reportError);
-
-    } else {
-      console.log(seedAdmin.username + ' user exists');
-    }
-  });
-} else {
-
-  //Add Local User
-  User.find({username: seedUser.username}).remove(function () {
-    var user = new User(seedUser);
-
-    // generate a random password and save
-    User.generateRandomPassphrase()
-    .then(saveUser(user))
-    .catch(reportError);
-  });
-
-  //Add Local Admin
-  User.find({username: seedAdmin.username}).remove(function () {
-    var user = new User(seedAdmin);
-
-    // generate a random password and save
-    User.generateRandomPassphrase()
-    .then(saveUser(user))
-    .catch(reportError);
+function removeUser (user) {
+  return new Promise(function (resolve, reject) {
+    var User = mongoose.model('User');
+    User.find({username: user.username}).remove(function (err) {
+        if (err) {
+          reject(new Error('Database Seeding:\t\t\tFailed to remove local ' + user.username));
+        }
+        resolve();
+    });
   });
 }
 
-// save the specified user with the password provided from the resolved promise
-function saveUser(user) {
-  return function (password) {
-    // set the new password
-    user.password = password;
+function saveUser (user) {
+  return function() {
+    return new Promise(function (resolve, reject) {
+      // Then save the user
+      user.save(function (err, theuser) {
+        if (err) {
+          reject(new Error('Database Seeding:\t\t\tFailed to add local ' + user.username));
+        } else {
+          resolve(theuser);
+        }
+      });
+    });
+  };
+}
 
-    // Then save the user
-    user.save(function (err) {
-      if (err) {
-        console.log('Database Seeding:\t\t\tFailed to add local ' + user.username);        
+function checkUserNotExists (user) {
+  return new Promise(function (resolve, reject) {
+    var User = mongoose.model('User');
+    User.find({username: user.username}, function (err, users) {
+        if (err) {
+          reject(new Error('Database Seeding:\t\t\tFailed to find local account ' + user.username));
+        }
+
+        if (users.length === 0) {
+          resolve();
+        } else {
+          reject(new Error('Database Seeding:\t\t\tFailed due to local account already exists: ' + user.username));
+        }
+    });
+  });
+}
+
+function reportSuccess (password) {
+  return function (user) {
+    return new Promise(function (resolve, reject) {
+      console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+      resolve();
+    });
+  };
+}
+
+// save the specified user with the password provided from the resolved promise
+function seedTheUser (user) {
+  return function (password) {
+    return new Promise(function (resolve, reject) {
+
+      var User = mongoose.model('User');
+      // set the new password
+      user.password = password;
+
+      if (user.username === 'admin' && process.env.NODE_ENV === 'production') {
+        checkUserNotExists(user)
+        .then(saveUser(user))
+        .then(reportSuccess(password))
+        .then(function () {
+          resolve();
+        })
+        .catch(reportError);
       } else {
-        console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+        removeUser(user)
+        .then(saveUser(user))
+        .then(reportSuccess(password))
+        .then(function () {
+          resolve();
+        })
+        .catch(reportError);
       }
     });
   };
 }
 
 // report the error
-function reportError(err) {
+function reportError (err) {
   console.log();
-  console.log('Database Seeding:\t\t\t Failed to generate random password');
   console.log(err);
   console.log();
 }
+
+module.exports.start = function start() {
+  var User = mongoose.model('User');
+  return new Promise(function (resolve, reject) {
+    var seedUser = {
+      username: 'user',
+      password: 'User_Password1!',
+      provider: 'local',
+      email: 'user@localhost.com',
+      firstName: 'User',
+      lastName: 'Local',
+      displayName: 'User Local',
+      roles: ['user']
+    };
+
+    var seedAdmin = {
+      username: 'admin',
+      password: 'Admin_Password1!',
+      provider: 'local',
+      email: 'admin@localhost.com',
+      firstName: 'Admin',
+      lastName: 'Local',
+      displayName: 'Admin Local',
+      roles: ['user', 'admin']
+    };
+
+    var user = null;
+    var adminAccount = new User(seedAdmin);
+    var userAccount = new User(seedUser);
+
+    //If production only seed admin if it does not exist
+    if (process.env.NODE_ENV === 'production') {
+      User.generateRandomPassphrase()
+      .then(seedTheUser(adminAccount))
+      .then(function () {
+          resolve();
+        })
+      .catch(reportError);
+    } else {
+      // Add both Admin and User account
+
+      User.generateRandomPassphrase()
+      .then(seedTheUser(userAccount))
+      .then(User.generateRandomPassphrase)
+      .then(seedTheUser(adminAccount))
+      .then(function () {
+          resolve();
+        })
+      .catch(reportError);
+    }
+  });
+};
--- a/config/lib/app.js
+++ b/config/lib/app.js
@@ -6,15 +6,18 @@
 var config = require('../config'),
   mongoose = require('./mongoose'),
   express = require('./express'),
-  chalk = require('chalk');
+  chalk = require('chalk'),
+  seed = require('./seed');
+
+function seedDB() {
+  if (config.seedDB) {
+    console.log(chalk.bold.red('Warning:  Database seeding is turned on'));
+    seed.start();
+  }
+}
 
 // Initialize Models
-mongoose.loadModels();
-
-//SeedDB
-if (config.seedDB) {
-  require('./seed');
-}
+mongoose.loadModels(seedDB);
 
 module.exports.loadModels = function loadModels() {
   mongoose.loadModels();
--- a/modules/core/tests/server/core.server.config.tests.js
+++ b/modules/core/tests/server/core.server.config.tests.js
@@ -1,0 +1,123 @@
+'use strict';
+
+/**
+ * Module dependencies.
+ */
+var should = require('should'),
+  mongoose = require('mongoose'),
+  User = mongoose.model('User'),
+  path = require('path'),
+  config = require(path.resolve('./config/config')),
+  seed = require(path.resolve('./config/lib/seed'));
+
+describe('Configuration tests', function () {
+
+  describe('Testing default seedDB:', function () {
+    before(function(done) {
+      User.remove(function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+
+    after(function(done) {
+      User.remove(function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+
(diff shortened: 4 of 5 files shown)


lirantal:
@mleanos care to review?

config/lib/seed.js
```diff
@@ -2,87 +2,39 @@
...
+      }
 
-    // generate a random password and save
-    User.generateRandomPassphrase()
-    .then(saveUser(user))
-    .catch(reportError);
-  });
-}
+      // For admin user account we do not remove them
+      if (user.username === 'admin' && users.length === 0) {
```
mleanos: I think this should be checking for a length greater than zero, or just `users.length`

Even with that fix, this still won't work. We're already in the callback of the `remove()`, so it's too late to check if we should be removing this user.

I think it might be a good idea to move the remove functionality outside of the `saveUser` method. We probably don't want to enter this function unless we are sure we want to add the new user.

mleanos:
@lirantal I left a line comment. Overall, I like what you've done. Other than what I pointed out, everything is working pretty well.. Tested this with the *production* env as well. 

One thing, that may just be preference, is the use of the callbacks for the promises. Rather than using *closures*. Is there a specific reason to use one way over the other?

Overall, I really like these changes & how you tied the seeding into the app & mongoose config. That seems like the best way to make this testable and scaleable.

lirantal:
@mleanos  Thanks, I'll address the admin user issue for the production environment state.

What do you suggest we use instead of closures and callbacks? Can you show an example?

lirantal:
@mleanos I updated the PR with the fix for the admin user account.

lirantal:
Check it out, I also added meanjs-core server tests so we can test our own code base.
This is of course just the beginning with the Seed DB feature so there will be more to come.

lirantal:
Closing and re-opening this PR because it seems that Travis CI was not working at all.

codydaig:
@lirantal Travis is having technical issues right now. 
https://www.traviscistatus.com/incidents/1r8qwcwdfk90

lirantal:
Ahh, explains it :)

config/lib/mongoose.js
```diff
@@ -9,11 +9,13 @@ var config = require('../config'),
...
 
 // Load the mongoose models
-module.exports.loadModels = function () {
+module.exports.loadModels = function (callback) {
   // Globbing model files
   config.files.server.models.forEach(function (modelPath) {
     require(path.resolve(modelPath));
   });
+
+  if (callback) callback();
```
ilanbiala: I'd prefer this returns a promise.

config/lib/seed.js
```diff
@@ -2,87 +2,60 @@
...
-  });
+        if (users.length === 0) {
+
+          // Then save the user
+          user.save(function (err) {
+            if (err) {
+              reject(new Error('Database Seeding:\t\t\tFailed to add local ' + user.username));
+            } else {
+              console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+              resolve();
```
ilanbiala: Why not return the saved user?

config/lib/seed.js
```diff
@@ -2,87 +2,60 @@
...
-  return function (password) {
-    // set the new password
-    user.password = password;
+        // Then save the user
+        user.save(function (err) {
+          if (err) {
+            reject(new Error('Database Seeding:\t\t\tFailed to add local ' + user.username));
+          } else {
+            console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+            resolve();
```
ilanbiala: Why not return the saved user?

config/lib/seed.js
```diff
@@ -92,3 +65,85 @@ function reportError(err) {
   console.log(err);
   console.log();
 }
+
+function performSeedUser(user) {
+
+  var User = mongoose.model('User');
+
+  return new Promise(function(resolve, reject) {
```
ilanbiala: either put a space here or remove the one from above, but do whatever we defined in the jshint or in the last formatting PR.

config/lib/seed.js
```diff
@@ -92,3 +65,85 @@ function reportError(err) {
...
+
+    var user = null;
+
+    //If production only seed admin if it does not exist
+    if (process.env.NODE_ENV === 'production') {
+      //Add Local Admin
+
+      user = new User(seedAdmin);
+
+      User.generateRandomPassphrase()
```
ilanbiala: I would do:
```
User
  .generateRandomPassphrase()
  .then(<code>)
  .catch(<code>);
```

config/lib/seed.js
```diff
@@ -92,3 +65,85 @@ function reportError(err) {
...
+          resolve();
+        })
+        .catch(reportError);
+
+    } else {
+
+      //Add Local User
+      user = new User(seedUser);
+
+      User.generateRandomPassphrase()
```
ilanbiala: Same here

config/lib/seed.js
```diff
@@ -92,3 +65,85 @@ function reportError(err) {
...
+      })
+      .then(function() {
+        var user = new User(seedAdmin);
+
+        User.generateRandomPassphrase()
+        .then(function(password) {
+          return saveUser(user, password);
+        })
+        .then(function() {
+          resolve();
```
ilanbiala: Why not just resolve in the previous one?

config/lib/seed.js
```diff
@@ -92,3 +65,85 @@ function reportError(err) {
...
+    }
+
+  });
+
+}
+
+module.exports.start = function start() {
+
+  return new Promise(function(resolve, reject) {
+    resolve(performSeedUser());
```
ilanbiala: Why not just have the start return a resolve of the function? This seems too complicated.

config/lib/seed.js
mleanos: I commented on the use of the closures here. So I figured I'd clarify here, since it's related to @ilanbiala comment...

@lirantal Here's an example of what I meant... 
```
User.generateRandomPassphrase()
.then(saveUser(user))
.then(completeSeedUser) // you can resolve your local promise in this function
.catch(reportError);

// saveUser would look like this
function saveUser (user) {
  return function (password) { // this is what I meant about using a closure, rather than a function
    ........
  }
}
```
Essentially, what you have is the same & it works the same. Maybe it's just a preference. However, it may make the code more readable & you can put all logic into their own functions & closures. For instance, if you return the user with the resolve from `saveUser` you could take the success reporting out of `saveUser` & into their own function where you only have to manage that code in one place.

mleanos:
@lirantal I've tested this, and the issue with the remove is fixed. However, in the original version of this seeding, the *admin* & *user* accounts were removed & re-created each time when the node env is *Development*. With these changes, you're not re-creating the *admin* account if in the *Development* env. Was that intentional? 

In addition to my line comment, I think you can do a couple things differently to reduce the amount of code & keep the logic's SOC in tact...

What about moving the remove functionality outside of the `saveUser` function? If you burden a new `removeUser` function with the task of determining whether or not to remove the user (based on env, and admin already existing) then your `saveUser` can be simplified. Also, by doing so you can reduce the code of your `performSeedUser` function even further. 

For example.. 
```
// resolve with success, or reject when there's an error or we don't want to touch, or add, the user
function removeUser(user) {
  return new Promise...
}
```
And the body of `performSeedUser` will look something like this.. 
```
removeUser(user)
.then(generatePassword)
.then(saveUser(user))
.then(userSeedSuccess)
.then(function () {
  // add your admin user here
})
.catch(reportError);
```
An alternative to the last call in the above chain, you can refactor `performSeedUser` to accept a *user* and call it from `start` each time you'd like to add a seeded user.

With all that said, I think this is a great start. And I agree that this can be the foundation for a grandeur refactoring of the back-end to make it more testable. How we implement these changes here can set the style for how we refactor, and implement promises, throughout the rest of the back-end. That's part of the reason why I'm so interested in getting this *right*, or at least understanding the style.

lirantal:
Thanks for the feedback guys.
I'm all about doing things right, so let's work through this PR together to get everything out right as it should be.

config/lib/seed.js
lirantal: @ilanbiala I chose to remove spaces between function arguments and the function keyword.

config/lib/mongoose.js
lirantal: Let's do this in another PR and keep the promises refactoring contained to the seed functionality.

config/lib/seed.js
lirantal: @mleanos the saveUser() must return a promise because it's doing an async operation.
If I leave the saveUser() as is in your original code then this breaks the whole point of the promises. The saveUser() function must be a thenable so I can chain it, that's why I refactored it to return a promise.

mleanos: I didn't mean that you shouldn't return a promise from the `saveUser` function... I just meant that you could add the closure, as in my example & return the promise from within that. To clarify my example..
```
function saveUser (user) {
  return function (password) { // this is what I meant about using a closure, rather than a function
    return new Promise...
  }
}
```
I realize that point wasn't very clear. The reason I prefer using the closures this way, is so you don't have to use a function from the `.then()`'s. I think it allows for more reusable code & is more readable.

lirantal: yeah I figured this is what you had in mind.
I'm not so sure I am in complete agreement about the readability, because that's just moving closing from one place to another. Just read again that block of code in saveUser and see how confusing that is - a function, that returns a function, that returns a promise. Goodluck debugging that :)

lirantal: @mleanos I'm on gitter, let's continue the refactoring there.

mleanos: @lirantal Also, see my example in my main comment below. I'm assuming everything is returning a promise.

ilanbiala: @lirantal @mleanos what I think the code should look like is:
```
fetchData()
  .then(prepareDataForCsv)
  .then(writeToCsv)
  .then(function() {
    console.log('your csv has been saved');
  });
```

Take a look at the article I pulled it from and what the function definitions look like: http://javascriptplayground.com/blog/2015/02/promises/

lirantal: @ilanbiala the desire is understood, I have discussed elaborately with Michael about it.
As far as I see it's not possible to pass multiple arguments to the callback function in a promise thenable, and so we can't just do:
```javascript
generatePassword()
.then(saveUser)
```
because that will pass only the password from the previous promise.
To also pass in the `user` argument we need to either do a closure based call like I did in this original PR or use the function invocation inside saveUser which will look like this:
```javascript
function saveUser(user) {
  return function(password) {
    return new Promise() .. {
  }
}
```
which I find quite ugly.

Is there another way of doing it? otherwise we need to choose one of the above, or refactor the generatePassword method to possibly take the user argument and return it along with the password.

config/lib/seed.js
ilanbiala: @lirantal is that consistent with what we did for the formatting refactor?

config/lib/seed.js
ilanbiala: @lirantal I'd rather make it nice and clean and pass along:
```
{
  user: user,
  password: password
}
```

mleanos: I don't we should passa user into the generatePassword function, just to return it with the password. In most cases we won't have a user to be concerned with, at the time of generating a password; if we do, this function shouldn't care or *know* about it. 

@lirantal Why do think that example of using a closure like that is "ugly"? Isn't that a standard way of using such closures, to solve this specific type of problem?

ilanbiala: @mleanos the password is always tied to a user because you wouldn't generate a seed password if there is no user for it to go with. Also, I agree with @lirantal that the closures are ugly and unintuitive. Most people would just pass the user and pass in 1 object instead of using a closure.

mleanos: @ilanbiala Generating a password doesn't have anything to do with a User; nor do we need to have a user in context when creating one. 

The reason that we wouldn't want to introduce a user to the generate password function, is so that the function can just be concerned with what it is intended to do (Autonomy). Which is to generate a random password. There could be cases where we would just want to generate a number of random passwords, and do something with them that is not tied to any User(s). And in this case, we don't even know if the User that we defined already will even make use of the password. The password & user don't (and shouldn't) care about each other until we're ready to assign the password to the user & save it.

As for the closure. I think we'd be missing out on some very key benefits of using Promises, by not introducing a closure here. Otherwise, we have to introduce some other way of handing the user off to the `saveUser` function. Thus, it might be a reason for you wanting to pass it into the generate password method; which from what I said earlier, you can gather that I deem to be breaking the notion of SOC.

mleanos: Also, I think it reads really well & is very intuitive. It seems to be one of the most basic usages of a closure, so this should be familiar enough to most developers.

lirantal:
@ilanbiala and @mleanos 
I've refactored and fixed formatting styles to conform with the comments, please review.

P.S. Going with the closures inside the functions is preferred IMO on moving the user object around.

config/lib/seed.js
```diff
@@ -2,93 +2,145 @@
...
-    var user = new User(seedUser);
+}
 
-    // generate a random password and save
-    User.generateRandomPassphrase()
-    .then(saveUser(user))
-    .catch(reportError);
-  });
+function saveUser (user) {
+  return function() {
```
mleanos: You don't need the closure here, if you're not using it to pass any parameters.

config/lib/seed.js
```diff
@@ -2,93 +2,145 @@
...
+      // set the new password
+      user.password = password;
 
-    // Then save the user
-    user.save(function (err) {
-      if (err) {
-        console.log('Database Seeding:\t\t\tFailed to add local ' + user.username);        
+      if (user.username === 'admin' && process.env.NODE_ENV === 'production') {
+        checkUserNotExists(user)
+        .then(saveUser(user))
```
mleanos: You forgot to add `reportSuccess` to the chain after successfully resolving `saveUser`.

mleanos:
@lirantal This looks really good! It's much cleaner & easier to follow. And I think it can be extended further in the future if need be.

I left a couple of minor comments. I'm going to spend some time with the tests now. I think we're pretty much there with this. I'm a little concerned that `seedTheUser` is still dependent on checking the environment & username, but I don't think we have a choice here with this specific use case; and it's a very minor concern for me.

One thing we might want to introduce at this time, is adding seed db options to the environmental configs. For instance, we could add the user info to the config. I can see users wanting to control that specifically based on the environment that the app is running in. 

I had an issue with the username being set to "admin" in my development environment. I already had a user with "admin" for the username. The seeding removed this user & I lost context of data that was related to this user. Now might be as good of a time as any to do this, if we can see the need for this to be added to the config.

config/lib/seed.js
ilanbiala: @mleanos it may be familiar, but it isn't intuitive without reading it multiple times, so I'm not a fan of that approach.

config/lib/seed.js
lirantal: Right, I added it now.

config/lib/seed.js
lirantal: We need the closure there so that saveUser can act as a promise, and thus it will call the resolve() with the data variable of `theuser`.
Try commenting out the closure function and test to see what I mean.

lirantal:
@ilanbiala @mleanos ready. even increased our coverage by 2%! ;-)
@mleanos about further refactoring to environment variables, how about we work on that in another PR?

codydaig:
@lirantal @ilanbiala @mleanos 
- I prefer closures as well!
- Great job! Looks great. :-D
- Let's leave the env variables refactor for another PR.
- I think travis is failing on 0.10 because native promises wern't supported until 0.12. If we want to keep 0.10 support, we need to use Q or Bluebird.

lirantal:
Great, thanks Cody.
With regards to versions, v0.10 did receive a backport for Promises support with an update for the rest of the V8 features so we're ok with v0.10 users and no need to deprecate support for that. The build failed for some other reason with db timeouts so I restarted it and it passes just fine now.

codydaig:
@lirantal Good to know. Thanks!

LGTM

config/lib/seed.js
mleanos: @lirantal I see what you mean. However, I think this behavior is stemming from the previous promise resolve in the chain. In both uses of the `saveUser` method, the previous promise is resolving with no data: `resolve();`

Hence, the `saveUser` method is expecting to handle that resolve, but if you don't have that closure there it expects that your `user` parameter of `saveUser` is coming from the resolve (which would be *undefined*).

Let's say you resolved `removeUser` with data: `resolve(user);` 
The next operation in the chain could look like this `.then(saveUser)`; in this case you wouldn't need that empty closure since it would get the `user` from the resolve.

Ok... Now with all that said. I think what you have is the correct way to handle it. I just wanted to clarify what my findings were, as to why we're seeing this behavior & need for the empty closure; and I'm sure you're aware of most of this. Furthermore, I think you're absolutely correct to resolve most of those functions with no data. For instance, it wouldn't make sense to resolve `removeUser`, and we want to maintain the `user` object from the scope that the chain is in. 

I'm very pleased with how this turned out! Thanks for working through this with me. I've definitely learned a few things. I think this is ready.

mleanos:
@lirantal I left a comment about the empty closure. I agree with your use of the closure there. However, I wanted to clarify why I think we're seeing that behavior.

I'm very pleased with how this turned out! Thanks for working through this with me. I've definitely learned a few things. I think this is ready. LGTM.

mleanos:
@lirantal One thing I forgot to mention.. The tests are logging the seed db events to the console. This is disruptive to the output of the tests. Let's work on that in a separate PR, so we can get this out.

config/lib/seed.js
lirantal: Exactly.

lirantal:
Sure. I noticed that too but didn't have much idea how to further work it out.
You're welcome to submit an enhancement PR with more tests as well.

But hey, we already increased coverage by 2% to 60%, that's awesome! :)

codydaig:
:+1:

mleanos:
@lirantal Once this is merged, I'll work on adding the user info the configs, and adding tests :)

And yea, pretty awesome, we're at 60%! Great job leading the charge on the test coverage!!

PR #941, opened 2015-09-29
Footer

trainerbill:
Add a footer view/controller

### DIFF
--- a/modules/core/client/views/footer.client.view.html
+++ b/modules/core/client/views/footer.client.view.html
@@ -5,6 +5,6 @@
     </div>
   </div>
   <div class="row">
-    <div class="col-md-4 col-md-offset-4">&copy; {{vm.year}} MEAN.js | All rights reserved</div>
+    <div class="col-md-4 col-md-offset-4">&copy; {{::vm.year}} MEAN.js | All rights reserved</div>
   </div>
 </div>
--- a/modules/core/client/views/footer.client.view.html
+++ b/modules/core/client/views/footer.client.view.html
@@ -1,0 +1,10 @@
+<div class="container" ng-controller="FooterController as vm">
+  <div class="row">
+    <div class="col-md-12">
+      <hr/>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-4 col-md-offset-4">&copy; {{vm.year}} MEAN.js | All rights reserved</div>
+  </div>
+</div>
--- a/modules/core/client/controllers/footer.client.controller.js
+++ b/modules/core/client/controllers/footer.client.controller.js
@@ -1,0 +1,12 @@
+(function () {
+  'use strict';
+  angular
+    .module('core')
+    .controller('FooterController', FooterController);
+
+    function FooterController() {
+      var vm = this;
+
+      vm.year = new Date().getFullYear();
+    }
+})();
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -43,16 +43,18 @@
     </section>
   </section>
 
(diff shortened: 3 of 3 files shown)


ilanbiala:
Why?

modules/core/client/views/footer.client.view.html
```diff
@@ -0,0 +1,10 @@
+<div class="container" ng-controller="FooterController as vm">
+  <div class="row">
+    <div class="col-md-12">
+      <hr/>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-4 col-md-offset-4">&copy; {{vm.year}} MEAN.js | All rights reserved</div>
```
jloveland: should use [app.title](https://github.com/meanjs/mean/blob/master/config/env/default.js#L5) instead of the MEAN.js text.

simison:
Seems a bit odd one, but if this goes trough I'd change `{{ vm.year }}` to `{{ ::vm.year }}` — [one-time-binding syntax](http://toddmotto.com/angular-one-time-binding-syntax/).

trainerbill:
@ilanbiala 

1.  I manually add it to most projects I start with MEAN.js
2.  Most applications have a footer
3.  It's a small PR
4.  I was bored on a plane.

Why not?

modules/core/client/views/footer.client.view.html
trainerbill: @jloveland Agreed as should the header.  How would you go about getting the server config to the client?  I think this opens up a larger can of worms that we need to discuss.

simison: >How would you go about getting the server config to the client?

a very good question.

For instance:

1. Have settings object next to user object at [the layout](https://github.com/meanjs/mean/blob/master/modules/core/server/views/layout.server.view.html#L48).

2. Feed `$window.settings` object for Modules as a [Constant](https://github.com/johnpapa/angular-styleguide#style-y241) (_"Use constants for values that do not change and do not come from another service."_) or via Service.

codydaig:
@trainerbill I'm not opposed to this PR, it seems like most people would have a footer. IMO, getting the config to the client side is outside the scope of this PR. So...

LGTM

trainerbill:
@simison I agree that getting server config to the client this should definitely be some functionality introduced to the project.  Things in the header and footer would benefit from this type of functionality.  

I want to keep this PR small and just introduce a footer.  I like your :: binding comment and will change that.  Can you submit a PR/issue to introduce server config to the client?

ilanbiala:
@lirantal I feel like most people scrap the UI, I'm not really sure whether we should add more or not. Anyway this is just copying the header setup, which I think most users can do if they want to.

lirantal:
@ilanbiala I agree, I don't have anything specific to say pro or con for this PR. I feel pretty neutral about it, and if it doesn't really add any real value to the framework then probably it's not worth to add more codebase that we will need to later on maintain and support.

trainerbill:
@lirantal Its only 23 lines of code.  I think the project needs a footer.  Yeah we can make people do it themselves but its a pretty small change, and I think will benefit people in the future.  IMO

ilanbiala:
@trainerbill If we keep saying "it's only x lines of code", eventually we'll be at an astoundingly large number. Don't use that as evidence of adding something in any project, because that's not how development should be driven.

codydaig:
@trainerbill Make it a sub generator in the generator project to generate a footer!

lirantal:
@trainerbill I'm in alignment with Ilan. I'm not sure it even qualifies as a valid sub-generator enhancement but it's not like we're short of stuff to work on

PR #309, opened 2014-12-17
Removed sudo from npm install grunt

kfuchs:

PR #942, opened 2015-09-29
Note for users that they may need to use sudo for installing the pack

SibuStephen:
Users note (For one using linux ), this note will provide an overview that sudo might be required for installing some of the packages as such.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Note: You may need to use sudo for OSX, *nix, BSD etc and run your command shell
 for installing the following commands.
 
 Make sure you have installed all of the following prerequisites on your development machine:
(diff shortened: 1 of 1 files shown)


simison:
BTW, NodeJS download URL is wrong, should probably be this instead: https://nodejs.org/en/download/

SibuStephen:
@simison  Please review, the changes as per saying.

lirantal:
No packages should be installed as sudo, we already discussed it.
See https://github.com/meanjs/mean/pull/309 and https://github.com/meanjs/mean/pull/350

SibuStephen:
@lirantal , Yes got it.

Issue #947, opened 2015-09-30
Update multer middleware in config/lib/express.js


gustavodemari:
Hi guys,

The multer library has changed, and also they've changed the way of using multer as a middleware in express.

[Old way](https://github.com/expressjs/multer/blob/c36cf0c33656ef3e2ba76a6118de14e7b61e3c6d/README.md#usage):
```js
var express = require('express')
var multer  = require('multer')

var app = express()
app.use(multer({ dest: './uploads/'}))
```

[New way](https://github.com/expressjs/multer#usage):
```javascript
var express = require('express')
var multer  = require('multer')
var upload = multer({ dest: 'uploads/' })

var app = express()

app.post('/profile', upload.single('avatar'), function (req, res, next) {
  // req.file is the `avatar` file
  // req.body will hold the text fields, if there were any
})

app.post('/photos/upload', upload.array('photos', 12), function (req, res, next) {
  // req.files is array of `photos` files
  // req.body will contain the text fields, if there were any
})

var cpUpload = upload.fields([{ name: 'avatar', maxCount: 1 }, { name: 'gallery', maxCount: 8 }])
app.post('/cool-profile', cpUpload, function (req, res, next) {
  // req.files is an object (String -> Array) where fieldname is the key, and the value is array of files
  //
  // e.g.
  //  req.files['avatar'][0] -> File
  //  req.files['gallery'] -> Array
  //
  // req.body will contain the text fields, if there were any
})
```

As you can see below, we are using the old way of usage in the [config/lib/express.js#L92-L97](https://github.com/meanjs/mean/blob/master/config/lib/express.js#L92-L97)
```js
  // Add multipart handling middleware
  app.use(multer({
    dest: './uploads/',
    inMemory: true
  }));
};
```

At the same time, we are using the parameter inMemory:true, that can lead our apps to run out of memory as pointed out here https://github.com/expressjs/multer#memorystorage

Also, I tried to remove the parameter inMemory (it's the same to set inMemory:false) and had an issue changing the profile image, because [our controller](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.profile.server.controller.js#L51-L90) uses [the buffer (req.files.file.buffer)](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.profile.server.controller.js#L59) to change the profile picture.

So, my suggestion is to update the way we use multer, and maybe, remove the parameter inMemory:true and adjust the controller to don't use req.files.file.buffer

[...]

PR #948, opened 2015-09-30
updating profile upload with a new version of multer

gustavodemari:
as described in #947

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -28,5 +28,13 @@ module.exports = {
   sessionKey: 'sessionId',
   sessionCollection: 'sessions',
   logo: 'modules/core/client/img/brand/logo.png',
-  favicon: 'modules/core/client/img/brand/favicon.ico'
+  favicon: 'modules/core/client/img/brand/favicon.ico',
+  uploads: {
+    profileUpload: {
+      dest: './modules/users/client/img/profile/uploads/', // Profile upload destination path
+      limits: {
+        fileSize: 1*1024*1024 // Max file size in bytes (1 MB)
+      }
+    }
+  }
 };
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -8,6 +8,8 @@ var _ = require('lodash'),
   path = require('path'),
   errorHandler = require(path.resolve('./modules/core/server/controllers/errors.server.controller')),
   mongoose = require('mongoose'),
+  multer = require('multer'),
+  config = require(path.resolve('./config/config')),
   User = mongoose.model('User');
 
 /**
@@ -54,15 +56,20 @@ exports.update = function (req, res) {
 exports.changeProfilePicture = function (req, res) {
   var user = req.user;
   var message = null;
+  var upload = multer(config.uploads.profileUpload).single('newProfilePicture');
+  var profileUploadFileFilter = require(path.resolve('./config/lib/multer')).profileUploadFileFilter;
+  
+  // Filtering to upload only images
+  upload.fileFilter = profileUploadFileFilter;
 
   if (user) {
-    fs.writeFile('./modules/users/client/img/profile/uploads/' + req.files.file.name, req.files.file.buffer, function (uploadError) {
-      if (uploadError) {
+    upload(req, res, function (uploadError) {
+      if(uploadError) {
         return res.status(400).send({
           message: 'Error occurred while uploading profile picture'
         });
       } else {
-        user.profileImageURL = 'modules/users/client/img/profile/uploads/' + req.files.file.name;
+        user.profileImageURL = config.uploads.profileUpload.dest + req.file.filename;
 
         user.save(function (saveError) {
           if (saveError) {
--- a/modules/users/tests/server/user.server.routes.tests.js
+++ b/modules/users/tests/server/user.server.routes.tests.js
@@ -847,6 +847,65 @@ describe('User CRUD tests', function () {
       });
   });
 
+  it('should be able to change profile picture if signed in', function (done) {
+    agent.post('/api/auth/signin')
+      .send(credentials)
+      .expect(200)
+      .end(function (signinErr, signinRes) {
+        // Handle signin error
+        if (signinErr) {
+          return done(signinErr);
+        }
+
+        agent.post('/api/users/picture')
+          .attach('newProfilePicture', './modules/users/client/img/profile/default.png')
+          .send(credentials)
+          .expect(200)
+          .end(function (userInfoErr, userInfoRes) {
+            // Handle change profile picture error
+            if (userInfoErr) {
+              return done(userInfoErr);
+            }
+
+            userInfoRes.body.should.be.instanceof(Object);
+            userInfoRes.body.profileImageURL.should.be.a.String();
+            userInfoRes.body._id.should.be.equal(String(user._id));
+
+            return done();
+          });
+      });
+  });
+
+  it('should not be able to change profile picture if not signed in', function (done) {
+    agent.post('/api/users/picture')
+      .attach('newProfilePicture', './modules/users/client/img/profile/default.png')
+      .send(credentials)
+      .expect(400)
+      .end(function (userInfoErr, userInfoRes) {
+        done(userInfoErr);
+      });
+  });
+
+  it('should not be able to change profile picture if attach a picture with a different field name', function (done) {
+    agent.post('/api/auth/signin')
+      .send(credentials)
+      .expect(200)
+      .end(function (signinErr, signinRes) {
+        // Handle signin error
+        if (signinErr) {
+          return done(signinErr);
+        }
+
+        agent.post('/api/users/picture')
+          .attach('fieldThatDoesntWork', './modules/users/client/img/profile/default.png')
+          .send(credentials)
+          .expect(400)
+          .end(function (userInfoErr, userInfoRes) {
+            done(userInfoErr);
+          });
+      });
+  });
+
   afterEach(function (done) {
     User.remove().exec(done);
   });
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -9,7 +9,6 @@ var config = require('../config'),
   bodyParser = require('body-parser'),
   session = require('express-session'),
   MongoStore = require('connect-mongo')(session),
-  multer = require('multer'),
   favicon = require('serve-favicon'),
   compress = require('compression'),
   methodOverride = require('method-override'),
@@ -88,12 +87,6 @@ module.exports.initMiddleware = function (app) {
   // Add the cookie parser and flash middleware
   app.use(cookieParser());
   app.use(flash());
-
-  // Add multipart handling middleware
-  app.use(multer({
-    dest: './uploads/',
-    inMemory: true
-  }));
 };
 
 /**
--- a/config/lib/multer.js
+++ b/config/lib/multer.js
@@ -1,0 +1,8 @@
+'use strict';
+
+module.exports.profileUploadFileFilter = function (req, file, cb) {
+  if (file.mimetype !== 'image/png' && file.mimetype !== 'image/jpg' && file.mimetype !== 'image/jpeg' && file.mimetype !== 'image/gif') {
+    return cb(new Error('Only image files are allowed!'), false);
+  }
+  cb(null, true);
+};
--- a/modules/users/client/controllers/settings/change-profile-picture.client.controller.js
+++ b/modules/users/client/controllers/settings/change-profile-picture.client.controller.js
@@ -7,7 +7,8 @@ angular.module('users').controller('ChangeProfilePictureController', ['$scope', 
 
     // Create file uploader instance
     $scope.uploader = new FileUploader({
-      url: 'api/users/picture'
+      url: 'api/users/picture',
+      alias: 'newProfilePicture'
     });
 
     // Set file uploader image filter
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "mocha": "~1.20.0",
     "mongoose": "^4.0.6",
     "morgan": "^1.6.1",
-    "multer": "0.1.8",
+    "multer": "~1.0.5",
     "nodemailer": "^1.4.0",
     "owasp-password-strength-test": "^1.3.0",
     "passport": "~0.2.2",


ilanbiala:
@gustavodemari please add tests for any applicable route, controller, and model part of this.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -87,10 +89,9 @@ exports.changeProfilePicture = function (req, res) {
...
   }
 };
-
 /**
  * Send User
  */
 exports.me = function (req, res) {
   res.json(req.user || null);
-};
+};
```
codydaig: Needs a New Line

gustavodemari:
My Travis-CI build is not passing due a timeout when getting all users in /api/users with an admin role.
I will try for the last time to reopen this PR to rebuild again.

mleanos:
@gustavodemari What's the reason for upgrading Multer here? I understand your concerns for the `inMemory` option. However, it seems for the purposes of using this for the Profile Image upload, it may not be much of a concern. 

Other than that, is there a good reason for this upgrade? I've spent quite a bit of time with Multer & angular-file-upload; I had my own implementation of a profile image upload when my project was still on the 0.3.3 version of MEANJS. 

I'm not necessarily opposed to this, but I'd like to understand it a bit more & spend some time testing the newer version of Multer.

mleanos:
@gustavodemari I've tested this and it doesn't appear to work with Firefox. The image is uploading correctly, and updating the user data with the new image url. However, I think it may have to do with the URL encoding. I've tested in Chrome, and it's working fine.

multer-profile-image-upgrade-mean 948

gustavodemari:
@mleanos I was having an issue to upload files in other routes and destinations, also, I was trying to upload larger files than pictures like .csv files in a instance of 512 MB of RAM, so I had a concern about inMemory.
Furthermore, I've tested in Firefox and it's working fine.
captura de tela de 2015-10-01 21 48 48

mleanos:
@gustavodemari It's still not working for me in Firefox. The issue seems to be that profileImageURL is saved as `modules\\users\\client\\img\\profile\\uploads\\78e540f60d2b73406420825e0d239ee3`, which is fine but the browser will end up decoding this as `modules%5Cusers%5Cclient%5Cimg%5Cprofile%5Cuploads%5C78e540f60d2b73406420825e0d239ee3`

Chrome seems to be able to handle this, but not Firefox. Before this change, profileImageURL was saved in the database as `modules/users/client/img/profile/testavatar.jpg`. I'd suggest saving the path with this format. If not, then you'd have to URL decode it before you display it. However, that would break any existing profile images.

Can you share what the path looks like for you in Firefox? The above screen shot only had the name of the image. Not the full path. That might help us understand why it's working for you, and not for me.

Can anyone else confirm this works for them in Firefox?

gustavodemari:
@mleanos Firefox 41.0 on Ubuntu 14.04.3 trusty

captura de tela de 2015-10-02 01 30 29

Also, I looked at /api/users/me and the profileImageURL was correct, see below.
```json
{
    "_id": "560c52d70eea21c17224bfb1",
    "displayName": "Gustavo De Mari",
    "provider": "local",
    "username": "gustavodemari",
    "__v": 0,
    "updated": "2015-10-02T00:47:46.138Z",
    "created": "2015-09-30T21:23:35.014Z",
    "roles": ["user"],
    "profileImageURL": "modules/users/client/img/profile/uploads/cabdd49cda2eb8a0bbcb6f937986eaf8",
    "email": "guhdemari@gmail.com",
    "lastName": "De Mari",
    "firstName": "Gustavo"
}
```

Are you sure that you are using my PR version?

mleanos:
@gustavodemari The data you provided isn't from the raw database is it? It looks like you've already queried it using mongoose; based on the date format. 

This is how my user data looks in my database after I update the profile picture... As you can see the URL saved is not the same as yours. I actually used Chrome for this particular update.. Although, again, Chrome can handle this URL, but Firefox can't.

```json
{
    "_id": {
        "$oid": "55d483767cc47ac40c01ab75"
    },
    "displayName": "Michael Leanos",
    "provider": "local",
    "username": "mleanos",
    "created": {
        "$date": "2015-08-19T13:24:06.484Z"
    },
    "roles": [
        "user"
    ],
    "profileImageURL": "modules\\users\\client\\img\\profile\\uploads\\7bd0a0ac2bd0e57527b31abc2e94936f",
    "lastName": "Leanos",
    "firstName": "Michael",
    "__v": 0,
    "updated": {
        "$date": "2015-09-25T07:19:37.151Z"
    }
}
```

I'm 100% certain I'm using your branch. 
WIndows 7
Firefox 41.0.1
Mongodb: 3.0.6

mleanos:
@gustavodemari My guess is that either Multer, or a package that it's using, is saving the path in format that is incompatible with either Windows and/or Firefox. Chrome being able to resolve it fine, doesn't surprise me. But we need to support Firefox on Windows.

I'm going to look into Multer's dependencies to see if I can determine where the issue is coming from. Do you have any way of testing this on Windows & Firefox?

gustavodemari:
@mleanos The data came from the HTTP request on /api/users/me.
Now, take a look at the raw database:
```json
{
	"_id" : ObjectId("560c52d70eea21c17224bfb1"),
	"salt" : "xxxxxxxxxxxxxxxxx",
	"displayName" : "Gustavo De Mari",
	"provider" : "local",
	"username" : "gustavodemari",
	"created" : ISODate("2015-09-30T21:23:35.014Z"),
	"roles" : [
		"user"
	],
	"profileImageURL" : "modules/users/client/img/profile/uploads/8a1ca188e4e0e4dcb7404976dfa9d6d0",
	"password" : "xxxxxxxxxxxxxxxxxxxxxx",
	"email" : "guhdemari@gmail.com",
	"lastName" : "De Mari",
	"firstName" : "Gustavo",
	"__v" : 0,
	"updated" : ISODate("2015-10-02T00:47:46.138Z")
}
```

Also, looking in your data, the field "updated" have an outdated date, which leads me to believe that  you are using an outdated version:
```json
"updated": {
        "$date": "2015-09-25T07:19:37.151Z"
    }

```

So, IHMO, isn't a browser problem, because the test are passing and I personally cloned my repo again, tested on Firefox and Chrome and everything is ok.

Can you try to git clone https://github.com/gustavodemari/mean.git and test again?

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -54,15 +55,16 @@ exports.update = function (req, res) {
...
-    fs.writeFile('./modules/users/client/img/profile/uploads/' + req.files.file.name, req.files.file.buffer, function (uploadError) {
-      if (uploadError) {
+    upload(req, res, function (err){
+      if(err) {
         return res.status(400).send({
           message: 'Error occurred while uploading profile picture'
         });
       } else {
-        user.profileImageURL = 'modules/users/client/img/profile/uploads/' + req.files.file.name;
+        user.profileImageURL = req.file.path;
```
mleanos: Thanks to @bastianwegge for pointing this out..

This line will not work in Windows with the current implementation of Multer, because of how it is storing the path. It's not accommodating for Windows paths. 

Can you update this line to use the following instead?
`user.profileImageURL = 'modules/users/client/img/profile/uploads/' + req.file.filename;` 

I've tested this and it works in Firefox as well.

mleanos:
@gustavodemari I just left a line comment. I think my suggestion is a safer way to calculate the path for the profile images.

https://github.com/meanjs/mean/pull/948/files#r41007858

mleanos:
@gustavodemari The `updated` field isn't updated during the change profile image method. That's why my `updated` field was "outdated". If it makes sense to add it, then perhaps you can do so now since you're modifying that code.

gustavodemari:
@mleanos I've tested on Chrome and Firefox on Windows, and it is working.

working_mean

Also, in your case, the updated field has to be recent, because as you showed in your screenshot, the success message "Profile Picture Changed Successfully" makes clearly that the [user.save method](https://github.com/gustavodemari/mean/blob/update-multer-profile-upload/modules/users/server/controllers/users/users.profile.server.controller.js#L69) was executed with success.

mleanos:
@gustavodemari Our user model `.pre('save')` hook doesn't update the `updated` field, neither does the `changeProfilePicture` method, so that field doesn't get updated when a user changes their profile picture. If you think otherwise, can you provide the reference?

I'm using the most up to date code base from master, and from your branch. I'm 100% confident about that. And it's simply not working for me in Firefox.

The reason this isn't working for me, is because the process that I'm running the application in. I'm running it directly from my Windows command prompt; I'm guessing you're running the app through *bash* or something similar, when you're testing on Windows.

None of our users, nor myself, should be required to run their app in an environment that can handle the `req.file.path`. We need to accommodate these other types of environments, when dealing with *paths*. The best way to handle this is to change this line 
https://github.com/meanjs/mean/pull/948/files#diff-028054ee3eec2278e53157ccf4668fe4R67 to 
`user.profileImageURL = 'modules/users/client/img/profile/uploads/' + req.file.filename;`

This is pretty close to how it was before it was changed with this PR. This solves the issue for me, and ensures this code works for all our users.

gustavodemari:
@mleanos About the updated field, you're right, sorry.

So, about running through bash or something similar...no, I'm not, I am using the command prompt (check the video below).

https://www.youtube.com/watch?v=usT32SrzJck

mleanos:
@gustavodemari In that video, you're using Chrome. I don't have any issues with Chrome. However, it does appear that the profileImageURL that is saved, is correct. If you say it's working for you in Firefox, I trust that.

The thing is, that it's genuinely not working for me. The profileImageURL is not saving in the correct format for me. For whatever reason. It's obviously caused by a difference in my environment. Since I'm having this issue, we can expect that others will also. 

Let's just go ahead and change that line to what I suggested. It's the safest way to make sure we support all users. Is there a specific reason that you want to use `req.file.path`, and not my suggestion?

bastianwegge:
@gustavodemari seriously, there is someone having a serious issue with the code you submitted (since it's not working for him) and you're arguing with him altough he gave you a detailed description what to do to make it work for you and for him. What's the point? Just go ahead and use a variable for the path in that controller and use that in Line 59 and 65. You both invested about 5 days of your sparetime in this PR ... isn't that enough ?

gustavodemari:
@mleanos Jump to 1:23 forward, I'm using Firefox.

My reason is just to understand what is happening, and now we know that isn't a browser problem, so, relax, I will update the code using your suggestion soon.

@bastianwegge As I said, I will update the code using @mleanos suggestion, relax, lol.

mleanos:
@gustavodemari Thank you for updating. I've tested this, and confirm it works as expected for me in Firefox & Chrome.

I'm all about trying to figure out the root cause of an issue together. It just felt like you were telling me that I was testing this wrong, and that since it works on your end, it should work for me. We can't always rely on that. I realize being able to reproduce the error is an important part of the debugging process. However, we knew the issue was related to the *path*. Paths behave differently from one environment to the next. 

I appreciate you helping out here. So don't let our back & forth discourage you going forward.

mleanos:
@ilanbiala I think this would introduce a breaking change. Perhaps it should go into the 0.5.0 queue?

I'm still exploring the implications of this upgrade; as it pertains to existing users. Particularly, Multer has been removed from the express config. Personally, it seems like a good thing to be able to remove Multer from this part of the app's config. However, I just don't know a great deal about Multer & the new features. Some users might be relying on Multer existing in the Express middleware, for purposes other than the profile image.

Any thoughts?

codydaig:
@gustavodemari Be sure to squash your commits! Otherwise, looks good. :-D Thanks!

gustavodemari:
@codydaig Thanks. Code squashed.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -54,15 +55,16 @@ exports.update = function (req, res) {
 exports.changeProfilePicture = function (req, res) {
   var user = req.user;
   var message = null;
+  var upload = multer({ dest: './modules/users/client/img/profile/uploads/' }).single('newProfilePicture');
 
   if (user) {
-    fs.writeFile('./modules/users/client/img/profile/uploads/' + req.files.file.name, req.files.file.buffer, function (uploadError) {
-      if (uploadError) {
+    upload(req, res, function (err){
```
ilanbiala: formatting

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -54,15 +55,16 @@ exports.update = function (req, res) {
...
+      if(err) {
```
ilanbiala: formatting

modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -262,6 +262,31 @@ describe('User CRUD tests', function () {
...
+        // Handle signin error
+        if (signinErr) {
+          return done(signinErr);
+        }
+
+        agent.post('/api/users/picture')
+          .attach('newProfilePicture', './modules/users/client/img/profile/default.png')
+          .send(credentials)
+          .expect(200)
+          .end(function (changePictureErr, changePictureRes) {
```
ilanbiala: What is the response?

modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -262,6 +262,31 @@ describe('User CRUD tests', function () {
...
+            // Handle change profile picture error
+            if (changePictureErr) {
```
ilanbiala: What is the error? A string?

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -54,15 +55,16 @@ exports.update = function (req, res) {
 exports.changeProfilePicture = function (req, res) {
   var user = req.user;
   var message = null;
+  var upload = multer({ dest: './modules/users/client/img/profile/uploads/' }).single('newProfilePicture');
```
lirantal: why did we remove folder configuration from the more generic area in express (https://github.com/meanjs/mean/pull/948/files#diff-16e0a4c137ae50cf68b1d2c3c914be85L93) to the controller?

It makes sense the an upload folder is a configuration item and someone should be able to configure it, even actually to take it out from express and put it as part of the env configuration, but definitely not to hard code it in the controller.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -54,15 +55,16 @@ exports.update = function (req, res) {
...
-    fs.writeFile('./modules/users/client/img/profile/uploads/' + req.files.file.name, req.files.file.buffer, function (uploadError) {
-      if (uploadError) {
+    upload(req, res, function (err){
+      if(err) {
         return res.status(400).send({
           message: 'Error occurred while uploading profile picture'
         });
       } else {
-        user.profileImageURL = 'modules/users/client/img/profile/uploads/' + req.files.file.name;
+        user.profileImageURL = 'modules/users/client/img/profile/uploads/' + req.file.filename;
```
lirantal: Again, don't hard code it. Let's use a config option for this.

lirantal:
@gustavodemari Good job adding tests with this PR.

modules/users/tests/server/user.server.routes.tests.js
gustavodemari: When is an upload error, the message is "Error occurred while uploading profile picture", so is a string.
Otherwise, is an user error when saving.

modules/users/server/controllers/users/users.profile.server.controller.js
gustavodemari: Instead,
```javascript
function (err){
```
I need to fix to:
```javascript
function (err) {
```
?

modules/users/server/controllers/users/users.profile.server.controller.js
gustavodemari: In this line, specifically, before I wrote just 
```js 
user.profileImageURL = req.file.path 
```
But, @mleanos said that he is having an issue with this line, as discussed above.

modules/users/server/controllers/users/users.profile.server.controller.js
gustavodemari: I did this way because I was having an issue when I needed to configure another upload path for another route.
Basically, when was an middleware in ``` config/lib/express.js ```, all uploads goes to the ```/uploads``` folder. If I needed to create another resource in my API where I upload a file and want to store in a different folder, I wasn't able to do it using multer.
Also, I agree with you that we can put the destination path in env, but, if some people, like me, want to add new destinations, how we can configure to support many destinations? Any suggestion?

gustavodemari:
@ilanbiala @lirantal Thanks for the feedbacks. I added some notes in yours notes.

modules/users/tests/server/user.server.routes.tests.js
ilanbiala: Assert that then.

modules/users/server/controllers/users/users.profile.server.controller.js
lirantal: Supporting multiple upload destinations isn't a problem.
Let's have a configuration variable like:
```javascript
uploads: {
  profileUploadDir: './modules/users/client/img/profile/uploads/'
```
and then from the code you can just do:
```javascript
var upload = multer({ dest: config.uploads.profileUploadDir }).single('newProfilePicture');
```

modules/users/server/controllers/users/users.profile.server.controller.js
lirantal: See my note above.
@mleanos is correct that
```javascript
user.profileImageURL = req.file.path 
```

is not correct. You should rather include the full path and not rely on the req.file.path variable for the location of the image file. Use the config variable as described above in my comment.

modules/users/server/controllers/users/users.profile.server.controller.js
gustavodemari: Fixed

gustavodemari:
@lirantal Thanks for your suggestion. So, it is ok now?

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -8,6 +8,8 @@ var _ = require('lodash'),
   path = require('path'),
   errorHandler = require(path.resolve('./modules/core/server/controllers/errors.server.controller')),
   mongoose = require('mongoose'),
+  multer = require('multer'),
+  defaultEnvConfig = require(path.resolve('./config/env/default')),
```
mleanos: I think you could just use the global config here..
`config = require(path.resolve('./config/config'))`

A user might add this setting to one of their env configs (i.e. Production) to override the default setting. WDYT?

mleanos:
@gustavodemari I left a line comment. But other than that, this looks good to me. I've tested this, and verified everything is working as expected. Thanks!

modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -262,6 +262,35 @@ describe('User CRUD tests', function () {
...
+            // Handle change profile picture error
+            if (userInfoErr) {
+              return done(userInfoErr);
+            }
+
+            userInfoRes.body.should.be.instanceof(Object);
+            userInfoRes.body.profileImageURL.should.be.a.String();
+            userInfoRes.body._id.should.be.equal(String(user._id));
+
+          return done();
```
mleanos: Indent here

modules/users/server/controllers/users/users.profile.server.controller.js
ilanbiala: Yeah I think that suggestion is good.

config/env/default.js
```diff
@@ -28,5 +28,8 @@ module.exports = {
   sessionKey: 'sessionId',
   sessionCollection: 'sessions',
   logo: 'modules/core/client/img/brand/logo.png',
-  favicon: 'modules/core/client/img/brand/favicon.ico'
+  favicon: 'modules/core/client/img/brand/favicon.ico',
+  uploads: {
+    profileUploadDir: './modules/users/client/img/profile/uploads/'
```
ilanbiala: . not needed

modules/users/server/controllers/users/users.profile.server.controller.js
lirantal: Indeed.

config/env/default.js
gustavodemari: Doesn't work without ./

modules/users/server/controllers/users/users.profile.server.controller.js
gustavodemari: Great, I will update with your suggestion.

modules/users/tests/server/user.server.routes.tests.js
gustavodemari: Done

config/env/default.js
ilanbiala: How do favicon and logo and the other ones work?

mleanos: @ilanbiala I think Multer requires it. If you look at the Express config diff, you'll see it was there. 

If that's the case, then we either have to add the *./* here, or add it here https://github.com/meanjs/mean/pull/948/files#diff-028054ee3eec2278e53157ccf4668fe4R59

@gustavodemari Maybe add a comment to the code to make it clear.

This way may be cleaner. WDYT?

lirantal: We still need to make changes to the config object here and also include a reference to the configuration option per environment and not hard code the upload dir destination.

gustavodemari: @ilanbiala 
Seems to me that favicon.ico is using ```./``` and isn't using ```config.favicon```
https://github.com/meanjs/mean/blob/master/config/lib/express.js#L68
Maybe, we need to adjust this.

@lirantal 
I can't understand your expectations, can you give more details?

lirantal: @gustavodemari that comment is irrelevant anymore with the recent changes. Good job.

modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -262,6 +262,35 @@ describe('User CRUD tests', function () {
     });
   });
 
+  it('should be able to change profile picture if signed in', function (done) {
```
ilanbiala: Can we also add a test for not being able to change profile picture if not signed in.

ilanbiala:
@gustavodemari LGTM except for 1 test that we should definitely add. Also, please add a couple tests to make sure the error handling works as expected. Feel free to check coveralls to see what lines aren't covered to know what specific test cases you may have missed. Also, we just need to get this rebased before we can merge.

gustavodemari:
@ilanbiala Added the test, now I will check the conflicts.

config/env/default.js
```diff
@@ -28,5 +28,19 @@ module.exports = {
...
+      limits: {
+        fileSize: 1*1024*1024 //Max file size in bytes
+      },
+      fileFilter: function (req, file, cb) {
+        if (file.mimetype !== 'image/png' && file.mimetype !== 'image/jpg' && file.mimetype !== 'image/jpeg' && file.mimetype !== 'image/gif') {
+            return cb(new Error('Only image files are allowed!'), false);
+        }
+        cb(null, true);
+      }
+    }
```
gustavodemari: I've made some change on the uploads config to add some other necessary things, like file size limit and filetype filtering

config/env/default.js
```diff
@@ -28,5 +28,19 @@ module.exports = {
   sessionKey: 'sessionId',
   sessionCollection: 'sessions',
   logo: 'modules/core/client/img/brand/logo.png',
-  favicon: 'modules/core/client/img/brand/favicon.ico'
+  favicon: 'modules/core/client/img/brand/favicon.ico',
+  uploads: {
+    profileUpload: {
+      dest: './modules/users/client/img/profile/uploads/', //Profile upload destination path
+      limits: {
+        fileSize: 1*1024*1024 //Max file size in bytes
```
ilanbiala: Why 1 MB? Also, please put that number in the comments and a space after //

gustavodemari: IHMO, 1 MB is ok for profile uploading. Also, I'll add the comment

config/env/default.js
lirantal: I like it, good stuff.
The function callback looks a bit odd in the configuration object, but can you think of another alternative?

gustavodemari: Yes, I agree that looks a bit odd.
But, the another alternative that I think is possible, is to define the file filter in the controller, but, seems to me that is the worst way, because doesn't scale when we've a scenario where we need to reuse the same filter in multiple controllers

mleanos: @gustavodemari What about adding this filter as an angular filter? I also think the filter could be refactored to be generic. Perhaps, add a parameter for the accepted file types? The list of file types seem more appropriate as a env config. 

Another option

simison: Whoop, please no app logic at config files.

@gustavodemari you can keep the codebase DRY by putting it somewhere else and exporting/importing it when needed, e.g. as a middleware.

mleanos: I misunderstood what this fileFilter is being used for. Is this a requirement of Multer? 

Since this is on the server side, can it be a reusable function?

config/env/default.js
```diff
@@ -28,5 +28,19 @@ module.exports = {
   sessionKey: 'sessionId',
   sessionCollection: 'sessions',
   logo: 'modules/core/client/img/brand/logo.png',
-  favicon: 'modules/core/client/img/brand/favicon.ico'
+  favicon: 'modules/core/client/img/brand/favicon.ico',
+  uploads: {
+    profileUpload: {
+      dest: './modules/users/client/img/profile/uploads/', //Profile upload destination path
+      limits: {
+        fileSize: 1*1024*1024 //Max file size in bytes (1 MB)
```
ilanbiala: space after `//`

config/env/default.js
```diff
@@ -28,5 +28,19 @@ module.exports = {
   sessionKey: 'sessionId',
   sessionCollection: 'sessions',
   logo: 'modules/core/client/img/brand/logo.png',
-  favicon: 'modules/core/client/img/brand/favicon.ico'
+  favicon: 'modules/core/client/img/brand/favicon.ico',
+  uploads: {
+    profileUpload: {
+      dest: './modules/users/client/img/profile/uploads/', //Profile upload destination path
```
ilanbiala: space after `//`

ilanbiala:
@gustavodemari last two things, then this is ready to go.

config/env/default.js
lirantal: First of all, this is indeed a server-side config thing.
Secondly, there's another down-side to put it like that which is that we now force that filefilter everywhere else multer is used and that is wrong because for uploading some other kind of assets you wouldn't want to force only images.

@gustavodemari please recommend another alternative.

gustavodemari: @mleanos I added this filter in the backend because the previous version allowed to upload any type of file using a POST in  /api/users/picture.
Also, ```filefilter``` is an [option](https://github.com/expressjs/multer#multeropts) of multer and I believe that this code can't be used on Angular.

@lirantal We don't force to use ```filefilter``` everywhere, because after these changes, multer will operate in a [independent way on the controller](https://github.com/gustavodemari/mean/commit/736efbf9ab563c4e2d1941ebb6d134322570e98d#diff-028054ee3eec2278e53157ccf4668fe4R59).

@simison Any suggestion where to put?

Some alternatives (feel free to give your opinion or to add suggestions):

Alternative 1:
Remove fileFilter function from config/env/default.js and define in the controller

config/env/default.js
```javascript
uploads: {
    profileUpload: {
      dest: './modules/users/client/img/profile/uploads/', // Profile upload destination path
      limits: {
        fileSize: 1*1024*1024 // Max file size in bytes (1 MB)
      }
    }
  }
```
modules/users/server/controllers/users/users.profile.server.controller.js
```javascript
var upload = multer(config.uploads.profileUpload).single('newProfilePicture');
upload.fileFilter = function (req, file, cb) {
  if (file.mimetype !== 'image/png' && file.mimetype !== 'image/jpg' && file.mimetype !== 'image/jpeg' && file.mimetype !== 'image/gif') {
        return cb(new Error('Only image files are allowed!'), false);
  }
  cb(null, true);
};
```
---
Alternative 2:
Remove fileFilter function from config/env/default.js, create config/lib/multer.js, define a profileUploadFilter in config/lib/multer.js and import config/lib/multer.js on user profile controller

 config/env/default.js
```javascript
uploads: {
    profileUpload: {
      dest: './modules/users/client/img/profile/uploads/', // Profile upload destination path
      limits: {
        fileSize: 1*1024*1024 // Max file size in bytes (1 MB)
      }
    }
  }
```

config/lib/multer.js

```javascript
module.exports.profileUploadFileFilter = function (req, file, cb) {
    if (file.mimetype !== 'image/png' && file.mimetype !== 'image/jpg' && file.mimetype !== 'image/jpeg' && file.mimetype !== 'image/gif') {
        return cb(new Error('Only image files are allowed!'), false);
    }
    cb(null, true);
  };
```

modules/users/server/controllers/users/users.profile.server.controller.js
```javascript
var profileUploadFileFilter = require(path.resolve('./config/lib/multer')).profileUploadFileFilter;
...
var upload = multer(config.uploads.profileUpload).single('newProfilePicture');
upload.fileFilter = profileUploadFileFilter;
```

mleanos: @gustavodemari I like the 2nd option. 

What do you think about adding a env config setting for a list of accepted file types for `profileUpload`? In the Multer configuration file, you can retrieve the list of accepted file types from the env config. 

Also, I think using lodash in your `fileFilter` function to check the current file's *mimetype* against the list. WDYT?

ilanbiala:
@gustavodemari can you fix up the last couple things and rebase so we can merge?

config/env/default.js
lirantal: @gustavodemari  2nd option is definitely better. I'd even move the file upload limit to that specific multer function in `config/lib/multer` because it is very specific.

gustavodemari:
@ilanbiala I think it's ready. Thanks.

codydaig:
LGTM

PR #949, opened 2015-09-30
[feat] Upgrade Forever Version

codydaig:


ilanbiala:
@codydaig I already have a PR for updating all the deps (hopefully we break nothing...), but Forever has no changelog and there were some releases that were unpublished for whatever reason, so I'm waiting for a changelog. Feel free to follow along at #950.

ilanbiala:
@

ilanbiala:
If we end up doing this, it will be part of #950.

PR #956, opened 2015-10-01
Mocha test timeouts [fixes #955]

### LINKED ISSUE #955: Mocha intermittent timeouts
We seem to be having intermittent issues with the Mocha tests timing out; including the Travis CI builds failing due to these timeouts. This has been disrupting the flow of contributions with our PR's builds failing. Personally, when I've committed changes to a PR, I have been crossing my fingers that no tests timeout. 

I suggest to implement suite level timeout increases; described here http://mochajs.org/#timeouts. This could be a permanent or temporary solution. Either way, it may save us a lot of headache when running through the CI builds.

I have a branch that I've been testing this with, and it seems to have solved the issue with timeouts. I'm setting the timeout using `this.timeout(10000);` for each server test suite (just once per file). It's a small change to each server test file. 

I wanted to open the discussion before I submitted a PR. Perhaps there is a bigger issue at play here, rather than just latency issues. Any thoughts, or concerns over this approach?

@lirantal @ilanbiala @codydaig @rhutchison @trainerbill @simison @bastianwegge @jloveland

mleanos:
Added a timeout of 10000 ms to each server test file. This is an attempt
to solve the timeout issues that we're experiencing with the Mocha
tests. Especially, this is hoping to address the build fails that are
caused by such timeouts.

Issue is described in https://github.com/meanjs/mean/issues/955

### DIFF
--- a/modules/articles/tests/server/article.server.model.tests.js
+++ b/modules/articles/tests/server/article.server.model.tests.js
@@ -17,6 +17,8 @@ var user, article;
  * Unit tests
  */
 describe('Article Model Unit Tests:', function () {
+  this.timeout(10000);
+
   beforeEach(function (done) {
     user = new User({
       firstName: 'Full',
--- a/modules/articles/tests/server/article.server.routes.tests.js
+++ b/modules/articles/tests/server/article.server.routes.tests.js
@@ -17,6 +17,8 @@ var app, agent, credentials, user, article;
  * Article routes tests
  */
 describe('Article CRUD tests', function () {
+  this.timeout(10000);
+
   before(function (done) {
     // Get application
     app = express.init(mongoose);
--- a/modules/chat/tests/server/chat.socket.tests.js
+++ b/modules/chat/tests/server/chat.socket.tests.js
@@ -4,5 +4,7 @@
  * Chat socket tests
  */
 describe('Chat Socket Tests:', function () {
+  this.timeout(10000);
+
   // TODO: Add chat socket tests
 });
--- a/modules/core/tests/server/core.server.config.tests.js
+++ b/modules/core/tests/server/core.server.config.tests.js
@@ -11,6 +11,7 @@ var should = require('should'),
   seed = require(path.resolve('./config/lib/seed'));
 
 describe('Configuration tests', function () {
+  this.timeout(10000);
 
   describe('Testing default seedDB:', function () {
     before(function(done) {
--- a/modules/users/tests/server/user.server.model.tests.js
+++ b/modules/users/tests/server/user.server.model.tests.js
@@ -16,6 +16,8 @@ var user1, user2, user3;
  * Unit tests
  */
 describe('User Model Unit Tests:', function () {
+  this.timeout(10000);
+
   before(function () {
     user1 = {
       firstName: 'Full',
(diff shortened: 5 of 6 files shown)


mleanos:
@lirantal That's unfortunate :( The build didn't fail due to test timeouts. It was caused by "network connectivity" though. 

And chance of re-running the build?

codydaig:
@mleanos You can close and reopen the PR to force Travis to rerun.

codydaig:
@mleanos Also, that could indicate that the timeout was due to network connectivity issues and an issue with Travis, not with our code.

mleanos:
@lirantal @ilanbiala Any thoughts on fast tracking this one? I really think it's gonna help out with all the build fails we're having.

lirantal:
@mleanos I'm ok, but let's see if we get some better ideas from @bastianwegge or anyone else on how to deal with it.

lirantal:
BTW looks like other projects are using similar method (namely meanio, and mongoose https://github.com/Automattic/mongoose/search?utf8=%E2%9C%93&q=%22this.timeout%22&type=Code)

mleanos:
@lirantal Fair enough. And yea, it seems like a straight forward method. Especially, since we can't predict the environmental factors that might contribute to these issues.

I'm definitely curious to understand what @bastianwegge had in mind though.

bastianwegge:
I investigated into the last failing builds on travis and found out that the `node-pre-gyp` ist mostly the reason for the build to fail, when it's trying to rebuild node.

Would you point me to any failing build you're according to? Since I can't reproduce it neither on my local jenkins nor on travis.

Is this a point where mongoose is sometimes taking longer?

mleanos:
@bastianwegge We've been having trouble with the Travis CI builds when submitting, or committing to, a PR. More often than not the issue seems that the test is taking longer than 2000ms, as indicated on the Travis logs. 

There may be other issues that are causing a periodic build fail, but mostly they've been the timeout errors, on any given test. There doesn't seem to be a pattern with which tests fail.

bastianwegge:
Ok maybe I'm missing out on something. The idea of "let's raise the timeout" didn't look good to me at first.

mleanos:
As @lirantal pointed out, it seems that others are using the timeout increase as a method to combat these Travis issues. It's been very frustrating to have the builds fail for no apparent reason, other than a test timeout. So I think this may be a fine solution for now. Perhaps, it's something on Travis' side.

lirantal:
@bastianwegge the timeouts aren't related to node build, they are happening as part of our tests suite when it runs. Some of them are related to interacting with the server API and others are related to database operations

bastianwegge:
@lirantal the timeouts michael is having locally are due to his connection from the local machine to mongolab. He's doing it right testing against a service instead of a local database to avoid issues in that case and he convinced me that it makes sense to set up a higher timeout. We never ran into such issues since our Jenkins has a mongodb locally installed that gets cleared before every run. Do you know what database travis uses?

lirantal:
@bastianwegge we got those timeouts as well on Travis, and I personally also getting them from time to time with a local mongodb installed here

codydaig:
@bastianwegge I believe travis is using a local mongo instance inside the container that the tests are ran in.

bastianwegge:
Yeah, seems to be just fine if nobody's concerned, I just wanted to jump to make you think about it. Go ahead and merge it, let's see if this breakes down the whole "timout" discussion. Otherwise I'd love to investigate with you into it ! :+1:

lirantal:
Cool.

PR #957, opened 2015-10-01
Seed options - logResults

mleanos:
Added an options object to the database seed configuration. Currently, the only option implemented is `logResults`; set to 'true' by default.

Setting the logResults option to `false` in the core configuration server test suite.

The initial reason for this option is to disable the console output of the seed db events, while running the test coverage. However, this can be useful in other cases. For instance, if the user doesn't want the passwords displayed in the console where they may be comprimised.

### DIFF
--- a/config/lib/seed.js
+++ b/config/lib/seed.js
@@ -1,15 +1,20 @@
 'use strict';
 
-var mongoose = require('mongoose'),
+var _ = require('lodash'),
+  config = require('../config'),
+  mongoose = require('mongoose'),
   chalk = require('chalk'),
   crypto = require('crypto');
+
+// global seed options object
+var seedOptions = {};
 
 function removeUser (user) {
   return new Promise(function (resolve, reject) {
     var User = mongoose.model('User');
     User.find({username: user.username}).remove(function (err) {
         if (err) {
-          reject(new Error('Database Seeding:\t\t\tFailed to remove local ' + user.username));
+          reject(new Error('Failed to remove local ' + user.username));
         }
         resolve();
     });
@@ -22,7 +27,7 @@ function saveUser (user) {
       // Then save the user
       user.save(function (err, theuser) {
         if (err) {
-          reject(new Error('Database Seeding:\t\t\tFailed to add local ' + user.username));
+          reject(new Error('Failed to add local ' + user.username));
         } else {
           resolve(theuser);
         }
@@ -36,13 +41,13 @@ function checkUserNotExists (user) {
     var User = mongoose.model('User');
     User.find({username: user.username}, function (err, users) {
         if (err) {
-          reject(new Error('Database Seeding:\t\t\tFailed to find local account ' + user.username));
+          reject(new Error('Failed to find local account ' + user.username));
         }
 
         if (users.length === 0) {
           resolve();
         } else {
-          reject(new Error('Database Seeding:\t\t\tFailed due to local account already exists: ' + user.username));
+          reject(new Error('Failed due to local account already exists: ' + user.username));
         }
     });
   });
@@ -51,7 +56,7 @@ function checkUserNotExists (user) {
 function reportSuccess (password) {
   return function (user) {
     return new Promise(function (resolve, reject) {
-      console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+      if (seedOptions.logResults) console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
       resolve();
     });
   };
@@ -66,14 +71,16 @@ function seedTheUser (user) {
       // set the new password
       user.password = password;
 
-      if (user.username === 'admin' && process.env.NODE_ENV === 'production') {
+      if (user.username === seedOptions.seedAdmin.username && process.env.NODE_ENV === 'production') {
         checkUserNotExists(user)
         .then(saveUser(user))
         .then(reportSuccess(password))
         .then(function () {
           resolve();
         })
-        .catch(reportError);
+        .catch(function (err) {
+          reject(err);
+        });
       } else {
         removeUser(user)
         .then(saveUser(user))
@@ -81,47 +88,49 @@ function seedTheUser (user) {
         .then(function () {
           resolve();
         })
-        .catch(reportError);
+        .catch(function (err) {
+          reject(err);
+        });
       }
     });
   };
 }
 
 // report the error
-function reportError (err) {
-  console.log();
-  console.log(err);
-  console.log();
+function reportError (reject) {
+  return function (err) {
+    if (seedOptions.logResults) {
+      console.log();
+      console.log('Database Seeding:\t\t\t' + err);
+      console.log();
+    }
+    reject(err);
+  };
 }
 
-module.exports.start = function start() {
+module.exports.start = function start(options) {
+  // Initialize the default seed options
+  seedOptions = _.clone(config.seedDB.options, true);
+
+  // Check for provided options
+
+  if (_.has(options, 'logResults')) {
+    seedOptions.logResults = options.logResults;
+  }
+
+  if (_.has(options, 'seedUser')) { 
+    seedOptions.seedUser = options.seedUser; 
+  }
+
+  if (_.has(options, 'seedAdmin')) {
+    seedOptions.seedAdmin = options.seedAdmin;
+  }
+
   var User = mongoose.model('User');
   return new Promise(function (resolve, reject) {
-    var seedUser = {
-      username: 'user',
-      password: 'User_Password1!',
-      provider: 'local',
-      email: 'user@localhost.com',
-      firstName: 'User',
-      lastName: 'Local',
-      displayName: 'User Local',
-      roles: ['user']
-    };
 
-    var seedAdmin = {
-      username: 'admin',
-      password: 'Admin_Password1!',
-      provider: 'local',
-      email: 'admin@localhost.com',
-      firstName: 'Admin',
-      lastName: 'Local',
-      displayName: 'Admin Local',
-      roles: ['user', 'admin']
-    };
-
-    var user = null;
-    var adminAccount = new User(seedAdmin);
-    var userAccount = new User(seedUser);
+    var adminAccount = new User(seedOptions.seedAdmin);
+    var userAccount = new User(seedOptions.seedUser);
 
     //If production only seed admin if it does not exist
     if (process.env.NODE_ENV === 'production') {
@@ -130,7 +139,7 @@ module.exports.start = function start() {
       .then(function () {
           resolve();
         })
-      .catch(reportError);
+      .catch(reportError(reject));
     } else {
       // Add both Admin and User account
 
@@ -141,7 +150,7 @@ module.exports.start = function start() {
       .then(function () {
           resolve();
         })
-      .catch(reportError);
+      .catch(reportError(reject));
     }
   });
 };
--- a/config/lib/seed.js
+++ b/config/lib/seed.js
@@ -56,7 +56,9 @@ function checkUserNotExists (user) {
 function reportSuccess (password) {
   return function (user) {
     return new Promise(function (resolve, reject) {
-      if (seedOptions.logResults) console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+      if (seedOptions.logResults) {
+        console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+      }
       resolve();
     });
   };
@@ -73,24 +75,24 @@ function seedTheUser (user) {
 
       if (user.username === seedOptions.seedAdmin.username && process.env.NODE_ENV === 'production') {
         checkUserNotExists(user)
-        .then(saveUser(user))
-        .then(reportSuccess(password))
-        .then(function () {
-          resolve();
-        })
-        .catch(function (err) {
-          reject(err);
-        });
+          .then(saveUser(user))
+          .then(reportSuccess(password))
+          .then(function () {
+            resolve();
+          })
+          .catch(function (err) {
+            reject(err);
+          });
       } else {
         removeUser(user)
-        .then(saveUser(user))
-        .then(reportSuccess(password))
-        .then(function () {
-          resolve();
-        })
-        .catch(function (err) {
-          reject(err);
-        });
+          .then(saveUser(user))
+          .then(reportSuccess(password))
+          .then(function () {
+            resolve();
+          })
+          .catch(function (err) {
+            reject(err);
+          });
       }
     });
   };
@@ -135,22 +137,22 @@ module.exports.start = function start(options) {
     //If production only seed admin if it does not exist
     if (process.env.NODE_ENV === 'production') {
       User.generateRandomPassphrase()
-      .then(seedTheUser(adminAccount))
-      .then(function () {
+        .then(seedTheUser(adminAccount))
+        .then(function () {
           resolve();
         })
-      .catch(reportError(reject));
+        .catch(reportError(reject));
     } else {
       // Add both Admin and User account
 
       User.generateRandomPassphrase()
-      .then(seedTheUser(userAccount))
-      .then(User.generateRandomPassphrase)
-      .then(seedTheUser(adminAccount))
-      .then(function () {
+        .then(seedTheUser(userAccount))
+        .then(User.generateRandomPassphrase)
+        .then(seedTheUser(adminAccount))
+        .then(function () {
           resolve();
         })
-      .catch(reportError(reject));
+        .catch(reportError(reject));
     }
   });
 };
--- a/modules/core/tests/server/core.server.config.tests.js
+++ b/modules/core/tests/server/core.server.config.tests.js
@@ -10,6 +10,11 @@ var should = require('should'),
   config = require(path.resolve('./config/config')),
   seed = require(path.resolve('./config/lib/seed'));
 
+/**
+ * Globals
+ */
+var user1, admin1, userFromSeedConfig, adminFromSeedConfig;
+
 describe('Configuration Tests:', function () {
   this.timeout(10000);
 
@@ -17,7 +22,32 @@ describe('Configuration Tests:', function () {
     before(function(done) {
       User.remove(function(err) {
         should.not.exist(err);
+
+        user1 = {
+          username: 'user_config_test',
+          provider: 'local',
+          email: 'user_config_test_@localhost.com',
+          firstName: 'User',
+          lastName: 'Local',
+          displayName: 'User Local',
+          roles: ['user']
+        };
+
+        admin1 = {
+          username: 'admin_config_test',
+          provider: 'local',
+          email: 'admin_config_test_@localhost.com',
+          firstName: 'Admin',
+          lastName: 'Local',
+          displayName: 'Admin Local',
+          roles: ['user', 'admin']
+        };
+
+        userFromSeedConfig = config.seedDB.options.seedUser;
+        adminFromSeedConfig = config.seedDB.options.seedAdmin;
+
         return done();
+
       });
     });
 
@@ -26,6 +56,22 @@ describe('Configuration Tests:', function () {
         should.not.exist(err);
         return done();
       });
+    });
+
+    it('should have seedDB configuration set for "regular" user', function(done) {
+      (typeof userFromSeedConfig).should.not.equal('undefined');
+      should.exist(userFromSeedConfig.username);
+      should.exist(userFromSeedConfig.email);
+
+      return done();
+    });
+
+    it('should have seedDB configuration set for admin user', function(done) {
+      (typeof adminFromSeedConfig).should.not.equal('undefined');
+      should.exist(adminFromSeedConfig.username);
+      should.exist(adminFromSeedConfig.email);
+
+      return done();
     });
 
     it('should not be an admin user to begin with', function(done) {
@@ -44,26 +90,26 @@ describe('Configuration Tests:', function () {
       });
     });
 
-    it('should set NODE_ENV to production and seedDB turned on so admin account must exist', function(done) {
+    it('should seed ONLY the admin user account when NODE_ENV is set to "production"', function(done) {
 
       // Save original value
       var nodeEnv = process.env.NODE_ENV;
       // Set node env ro production environment
       process.env.NODE_ENV = 'production';
 
-      User.find({username: 'admin'}, function(err, users) {
+      User.find({username: adminFromSeedConfig.username}, function(err, users) {
 
         // There shouldn't be any errors
         should.not.exist(err);
         users.should.be.instanceof(Array).and.have.lengthOf(0);
 
-        seed.start().then(function() {
-          User.find({username: 'admin'}, function(err, users) {
+        seed.start({ logResults: false }).then(function() {
+          User.find({username: adminFromSeedConfig.username}, function(err, users) {
             should.not.exist(err);
             users.should.be.instanceof(Array).and.have.lengthOf(1);
 
-            var admin = users.pop();
-            admin.username.should.equal('admin');
+            var _admin = users.pop();
+            _admin.username.should.equal(adminFromSeedConfig.username);
 
             // Restore original NODE_ENV environment variable
             process.env.NODE_ENV = nodeEnv;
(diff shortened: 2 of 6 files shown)


codydaig:
@mleanos I'd prefer this be an option in the env config.

mleanos:
@codydaig I originally thought the same thing. However, we'd have to explicitly set the option to be specifically for when the tests run.

For example: 
```javascript
seedDb: {
  logResultsDuringTests: process.env.logResultsDuringTests || false
}
```
Maybe using options at the seed config, and env config would work.. that way someone could set the option in the env config if they want it applied everywhere. Or they could leave the env config as the default of `true`, and then set it using the options parameter for the seed config.

The key thing is that we need to provide a way to mute the console messages when the test suite is running vs when the db is being seeded.

I think I still prefer having the options parameter. But I'm not opposed to having a "global" env config as well.

mleanos:
Another thing to keep in mind, is that the Seed Db may, eventually, be used in other parts of the application. Hence, why we have the `.start` exported. So we'd want to accommodate an option to mute console messages on any given usage.

config/lib/seed.js
```diff
@@ -4,6 +4,11 @@ var mongoose = require('mongoose'),
   chalk = require('chalk'),
   crypto = require('crypto');
 
+// set default seed options
+var seedOptions = {
+  logResults: true // log the results of the seeding to the console
```
mleanos: @codydaig What if we set the default here, from the env config? While keeping the seedOptions object.

lirantal: That's a lot better, yes. It means we are not inventing another configuration option but instead re-using and existing one.

lirantal:
@mleanos will this logging option support also logging the data (the seed user/pass that was created) into the app log file?

mleanos:
@lirantal We could have an option for that as well. Or have *logging options* that may include a list of where to log the results.. `['console', 'app logs'`.. etc. We don't currently have the app log file functionality associated with the seeding, but we could introduce it.

The key point for me is to be able to control the logging, or any other options that we may introduce, on any given seed operation.. Meaning, that when someone calls `seed.start` from multiple areas of the application, they can specify options for each of the usages.

mleanos:
Also, I think having default seed options coming from the env configs would a good idea. WDYT?

mleanos:
Made some modifications, based on suggestions from @codydaig & @lirantal 

The seed options are now defaulting from env configs. Also, changed the seedDB env config setting to an object, with an `options` field.

Also, fixed an issue with how the env config was reading the seed setting from the processes env variable; see commit message for more details.

bastianwegge:
Coming from a ruby/C# Scene the whole seeding thing seems to get a lot bigger than it has to be. @lirantal & @codydaig what do you think about creating mocks that can be used in the tests for example for a user? In that case you'd test the mocks using your model-tests, would have only two places (model&mock) to reconfigure when changing something and you'd have the seeding right away. That would obviate the need to tests the seeding. Talked about that with @mleanos earlier, but we were disconnected while talking.

config/lib/seed.js
```diff
@@ -89,12 +95,16 @@ function seedTheUser (user) {
...
+  if (seedOptions.logResults) {
+    console.log();
+    console.log(err);
+    console.log();
+  }
 }
 
-module.exports.start = function start() {
+module.exports.start = function start(options) {
+  if (options && options.hasOwnProperty('logResults')) seedOptions.logResults = options.logResults;
```
lirantal: How about using lodash's .has method? it's a shorthand way of doing it and we can use it consistently across the project?

lirantal:
@mleanos how about we move the seeded user details outside of the seed file and into the env variables as well as config options? (@bastianwegge is that what you meant by mock users?)

bastianwegge:
This is what I meant:

Imagine a MockFile like this `user.model.mock.js` using faker.js.

```
'use strict';
var faker = require('faker');

function user(credentials) {
	if (credentials === undefined) {
		credentials = {
			username: 'username',
			password: 'password'
		};
	}

	return {
		displayName: faker.name.findName(),
		email: faker.internet.email(),
		username: credentials.username,
		password: credentials.password,
		provider: 'local',
		roles: ['user']
	};
}

module.exports = {
	User: user
};
```

You would require this in your tests that would create something like this:

```
describe('User Model Unit Tests:', function() {

  afterEach(function(done) {
    User.remove().exec(done);
  });

  beforeEach(function() {
    user = userMock.User();
  });

  describe('Method Save', function() {
    it('should begin with no users', function(done) {
      User.find({}, function(err, users) {
        users.should.have.length(0);
        done();
      });
    });

    it('should be able to save without problems', function(done) {
      var _user = new User(user);

      _user.save(function(err) {
        should.not.exist(err);
        done();
      });
    });
  })
});
```

This way you would have tested your mongoose-Schema and your mock. Then you could just go ahead and create something like a seeding-service like the following `seeds.js` without the necessity of testing it because you're using tested services:

```
function addTestUser(done) {
	var _user = new User(userMock.User());
	_user.save(done);
}
```

Any questions?

config/lib/seed.js
ilanbiala: Please use {} for if statements.

mleanos:
@lirantal I will address your comments today. I already have a branch that moves the User details out of the seed configuration and puts it into the env configs. I'll add it to this PR.

@bastianwegge What you're suggesting is outside the scope of what this PR is intended for. I think you make some valid points, and I'd like to discuss it further; and get everyone else involved. This would be a big change in most of the server side tests, so it's probably best for a separate PR. Thanks for the idea; I think we can incorporate something like your example.

mleanos:
@lirantal @codydaig  I addressed your comments, and added some tests to the core server config test suite.

1) Moved the seeded user details to the env configs (development, production, and test). 
2) Added the seeded user details to the seedDB config options; defaults to use the env config.
3) Added additional tests to the core server configuration test suite
4) Added the usage of lodash's `.has` method to check the existence of the custom seed options.

@ilanbiala I actually prefer the way the code reads using the inline if statements, in this case. Otherwise, we would have something like this
```javascript
if (options) {
    seedOptions.logResults = options.logResults || seedOptions.logResults;
    seedOptions.seedUser = options.seedUser || seedOptions.seedUser;
    seedOptions.seedAdmin = options.seedAdmin || options.seedAdmin;
  }
```
I think what I have now, is much cleaner, and we only touch the seedOptions when the options are provided. I can change this based on your suggestion, if it goes against our coding styles. I'm just not aware that we have a desired style in cases like this. 

Something that occurred to me while writing the tests, was that we can only test against the *test* env seed settings; even though we're manually changing the *process.env.NODE_ENV*. This probably isn't a problem, but I wanted to point it out. 

One last thing, is the possibility of the seedDB being turned on for the *test* environment. Should we allow this? Or should we set the `seedDB.seed` setting to `false` in the *test* env config? This isn't a big deal to me, but it brought up the issue of whether or not we can expect a user of this project to run the application using the *test* environment. Just something that seems ambiguous to me. I can see this go both ways.

mleanos:
BTW, I'm surprised the coverage only went up .02%. I had such high hopes for an increase here :( haha 

I'm just glad the build succeeded :)

mleanos:
@lirantal Was there a specific reason this is here? I don't see a usage. I forgot to mention it before
https://github.com/meanjs/mean/pull/957/files#diff-7dfe67f86c4c0773b98c0422a0cb606aL122

lirantal:
Yeah I see it's not being used, can you fix it?

lirantal:
@mleanos why do we have both `userFromSeedConfig` and `user1`? are you testing the config options are actually set? I don't think that's required...

mleanos:
@lirantal Yes. I can remove that unused var. 

The `userFromSeedConfig` is used to test the seeding using the *default* seed user details set in the in env config. `user1` is used to test the passing of the seed user details into the `.start` method using the options param.

If we don't test against the *default* settings, then we can only test the seed using the passed in user details, using the options param.

What you're saying is that we can't guarantee that the env config will contain the seed user details? It would be nice to test the seeding using the *default* settings from the env config though. 

How about we test for the existence of the user details in the env config, and if present then we can proceed to test using those settings? Otherwise, we just allow the test to pass since we don't have the user details to test against.

lirantal:
Got it. I'm ok with the tests as is.

mleanos:
@ilanbiala How do you feel about this? https://github.com/meanjs/mean/pull/957/files#diff-7dfe67f86c4c0773b98c0422a0cb606aR105 

See my comment above about it https://github.com/meanjs/mean/pull/957#issuecomment-145201321

mleanos:
@codydaig @ilanbiala @lirantal How do the seedDB env config settings look to you?

I'm wondering if I need to add something like `process.env.MONGO_SEED_USER_USERNAME` to both the admin and user account details. Would probably only need this for username & email.. WDYT?

Also, would it be appropriate to add an exported function to the seed configuration, to retrieve the options? I'm thinking that it would be better to rely on the seed configuration to interact with the seedDB options from other parts of the application. For instance, the server config test suite; rather than retrieving the options from the application config like here https://github.com/meanjs/mean/pull/957/files#diff-27fe56fc024d6e9d46e8c361ad7879e4R45

The idea I'm going with here is to add some additional sanitization to the options. Perhaps, to gracefully handle when the seedDB options aren't present in the env config. I'm trying to get the seedDB feature to be more extensible & reusable; but also not too opinionated. Any thoughts?

ilanbiala:
@mleanos no, that isn't okay in my opinion, and honestly having one strict style is better. Also, not having brackets leads to confusion and other issues that you may be aware of.

mleanos:
@ilanbiala Are you specifically saying that you prefer something like this?

```javascript
if (_.has(options, 'logResults')) { seedOptions.logResults = options.logResults };
if (_.has(options, 'seedUser')) { seedOptions.seedUser = options.seedUser };
if (_.has(options, 'seedAdmin')) { seedOptions.seedAdmin = options.seedAdmin };
```

Or do you mean you'd rather have it broken up into multiple lines?

```javascript
if (_.has(options, 'logResults')) { 
  seedOptions.logResults = options.logResults 
};

if (_.has(options, 'seedUser')) { 
  seedOptions.seedUser = options.seedUser 
};

if (_.has(options, 'seedAdmin')) { 
  seedOptions.seedAdmin = options.seedAdmin 
};
```

If the latter, I'd be concerned with readability. Although, we could put this logic in a separate function like `initSeedOptions(options);` that sets the global seedDB options object. Any ideas on that?

ilanbiala:
The latter, and put it in whatever way makes sense. But the bracket requirement should be strict.

lirantal:
I think ```process.env.MONGO_SEED_USER_USERNAME``` is a good idea to add to the default env files. I don't think we need more exported functions on the seed file at the moment.

mleanos:
@ilanbiala @lirantal Ok. Thanks.

And to clarify, we're ok with requiring that existence of the seedDB env config in order for the tests to run & pass?

ilanbiala:
@mleanos I think so, but as opposed to what?

lirantal:
Right.

mleanos:
@ilanbiala I guess since we're adding these settings to the *test* env config, they *should* always be there unless someone removes it; or for whatever reason doesn't get the updated *test* env settings once this is merged. In this case, the tests won't run properly. However, I'm probably being too cautious :) 

People do some funny things with their env configs; I know this, because I've done it myself. But I'm sure we're fine in this case. Thanks for the clarification.

mleanos:
Updated per the requested changes, and squashed.

@codydaig @ilanbiala @lirantal Final review?

ilanbiala:
@mleanos is there any way the tests can be broken down. Also, if you are concerned about the coverage increase being too small, take a look at https://coveralls.io/jobs/8743576 and add any other tests that go through the red lines.

mleanos:
@ilanbiala Can you be more specific on what you think is too complicated? I added 4 tests, but 2 of them are essentially copies of what was previously there. I'm more than happy to work on those tests, but I need a little more context as to what you mean.

Thanks for the suggestion on the coverage. I'm still learning the intricacies of Travis. I'll come up with some additional tests that satisfy those other conditions.

ilanbiala:
@mleanos looking over it again, the tests are quite long (longer than I'd prefer at least for the sake of quick reading and checking), but it doesn't seem like we can truly break down the tests into smaller unit tests, so I guess that's fine. See whatever else you can add to cover other paths in the code.

mleanos:
@ilanbiala I added some additional tests. The coverage in the SeedDB configuration is pretty good now. There are still a couple of lines that I can't easily (nor practically) test for. 

There are two rejects that can happen if a database operation unexpectedly fails
https://coveralls.io/builds/3759560/source?filename=..%2Fconfig%2Flib%2Fseed.js#L17
https://coveralls.io/builds/3759560/source?filename=..%2Fconfig%2Flib%2Fseed.js#L44

Also, there are a few lines that report an error to the console. However, I am disabling the console logs during this test suite.
https://coveralls.io/builds/3759560/source?filename=..%2Fconfig%2Flib%2Fseed.js#L100

Overall, I think these tests are pretty solid. Let me know if you see something additional that I can do.

mleanos:
@lirantal I had to change how the rejects were being propagated in the seed configuration. In order to test against the expected errors, I had to be able to send the errors back to the test suite. 

See this diff here to see what I changed. Let me know what you think, and if you see any issues with this.
https://github.com/meanjs/mean/pull/957/files#diff-7dfe67f86c4c0773b98c0422a0cb606aR81
https://github.com/meanjs/mean/pull/957/files#diff-7dfe67f86c4c0773b98c0422a0cb606aR139

lirantal:
@mleanos looks like we can't get totally rid of closures then, huh? :)
I guess it's ok.

mleanos:
@lirantal Are you referring to this?
https://github.com/meanjs/mean/pull/957/files#diff-7dfe67f86c4c0773b98c0422a0cb606aR154

If so, I actually modified that since my last push. I wanted to get feedback first. Do you think this is better? 
```javascript
User.generateRandomPassphrase()
  .then(seedTheUser(userAccount))
  .then(User.generateRandomPassphrase)
  .then(seedTheUser(adminAccount))
  .then(function () {
    resolve();
  })
  .catch(reportError(reject));
```
And then `reportError` will look like this..
```javascript
// report the error
function reportError (reject) {
  return function (err) {
    if (seedOptions.logResults) {
      console.log();
      console.log('Database Seeding:\t\t\t' + err);
      console.log();
    }
    reject(err);
  };
}
```
Maybe this is weird to pass the `reject` callback to the `reportError` function? WDYT?

mleanos:
@lirantal @ilanbiala Rebased and squashed. Was not a fun merge conflict :/ 

I added the change to the `reportError` function as described in my above comment. I can revert that if you don't like it. Let me know what you think. Thanks.

lirantal:
@mleanos yeah I think the change with the functions is better.
LGTM

config/lib/seed.js
```diff
@@ -51,7 +56,7 @@ function checkUserNotExists (user) {
 function reportSuccess (password) {
   return function (user) {
     return new Promise(function (resolve, reject) {
-      console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
+      if (seedOptions.logResults) console.log(chalk.bold.red('Database Seeding:\t\t\tLocal ' + user.username + ' added with password set to ' + password));
```
ilanbiala: braces

config/lib/seed.js
```diff
@@ -36,13 +41,13 @@ function checkUserNotExists (user) {
     var User = mongoose.model('User');
     User.find({username: user.username}, function (err, users) {
         if (err) {
-          reject(new Error('Database Seeding:\t\t\tFailed to find local account ' + user.username));
+          reject(new Error('Failed to find local account ' + user.username));
         }
 
         if (users.length === 0) {
           resolve();
```
ilanbiala: Shouldn't this resolve with something like `true` or `[]` or something more concrete than nothing?

config/lib/seed.js
```diff
@@ -1,15 +1,20 @@
...
+
 function removeUser (user) {
   return new Promise(function (resolve, reject) {
     var User = mongoose.model('User');
     User.find({username: user.username}).remove(function (err) {
         if (err) {
-          reject(new Error('Database Seeding:\t\t\tFailed to remove local ' + user.username));
+          reject(new Error('Failed to remove local ' + user.username));
         }
         resolve();
```
ilanbiala: Resolve with something to affirm, or maybe the removed user.

config/lib/seed.js
```diff
@@ -66,14 +71,16 @@ function seedTheUser (user) {
       // set the new password
       user.password = password;
 
-      if (user.username === 'admin' && process.env.NODE_ENV === 'production') {
+      if (user.username === seedOptions.seedAdmin.username && process.env.NODE_ENV === 'production') {
         checkUserNotExists(user)
         .then(saveUser(user))
```
ilanbiala: indent

config/lib/seed.js
```diff
@@ -66,14 +71,16 @@ function seedTheUser (user) {
...
         .then(reportSuccess(password))
         .then(function () {
           resolve();
```
ilanbiala: What is this resolving to?

config/lib/seed.js
```diff
@@ -81,47 +88,49 @@ function seedTheUser (user) {
...
```
ilanbiala: What is this resolving to?

config/lib/seed.js
```diff
@@ -130,7 +139,7 @@ module.exports.start = function start() {
       .then(function () {
           resolve();
```
ilanbiala: What is this resolving to?

config/lib/seed.js
```diff
@@ -141,7 +150,7 @@ module.exports.start = function start() {
...
```
ilanbiala: What is this resolving to?

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -17,6 +22,30 @@ describe('Configuration Tests:', function () {
...
+          firstName: 'Admin',
+          lastName: 'Local',
+          displayName: 'Admin Local',
+          roles: ['user', 'admin']
+        };
+
+        userFromSeedConfig = config.seedDB.options.seedUser;
+        adminFromSeedConfig = config.seedDB.options.seedAdmin;
+
         done();
```
ilanbiala: return

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -28,6 +57,22 @@ describe('Configuration Tests:', function () {
       });
     });
 
+    it('should have seedDB configuration set for "regular" user', function(done) {
+      (typeof userFromSeedConfig).should.not.equal('undefined');
```
ilanbiala: I don't see anything async here.

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -28,6 +57,22 @@ describe('Configuration Tests:', function () {
...
+      should.exist(userFromSeedConfig.username);
+      should.exist(userFromSeedConfig.email);
+
+      return done();
+    });
+
+    it('should have seedDB configuration set for admin user', function(done) {
+      (typeof adminFromSeedConfig).should.not.equal('undefined');
```
ilanbiala: Async?

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -44,26 +89,26 @@ describe('Configuration Tests:', function () {
...
-      User.find({username: 'admin'}, function(err, users) {
+      User.find({username: adminFromSeedConfig.username}, function(err, users) {
 
         // There shouldn't be any errors
         should.not.exist(err);
         users.should.be.instanceof(Array).and.have.lengthOf(0);
 
-        seed.start().then(function() {
-          User.find({username: 'admin'}, function(err, users) {
+        seed.start({ logResults: false }).then(function() {
```
ilanbiala: split across multiple lines.

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -77,34 +122,152 @@ describe('Configuration Tests:', function () {
...
       process.env.NODE_ENV = 'test';
 
-      User.find({username: 'admin'}, function(err, users) {
+      User.find({username: adminFromSeedConfig.username}, function(err, users) {
+
+        // There shouldn't be any errors
+        should.not.exist(err);
+        users.should.be.instanceof(Array).and.have.lengthOf(0);
+
+        seed.start({ logResults: false }).then(function() {
```
ilanbiala: Multiple lines

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -77,34 +122,152 @@ describe('Configuration Tests:', function () {
...
+          // There shouldn't be any errors
+          should.not.exist(err);
+
+          User.find({ username: { $in: [adminFromSeedConfig.username, userFromSeedConfig.username] } }, function (err, users) {
+
+            // There shouldn't be any errors
+            should.not.exist(err);
+            users.should.be.instanceof(Array).and.have.lengthOf(2);
+
+            seed.start({ logResults: false })
```
ilanbiala: new line for .start

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -77,34 +122,152 @@ describe('Configuration Tests:', function () {
...
+      // Set node env ro production environment
+      process.env.NODE_ENV = 'production';
+
+      User.find({username: admin1.username}, function(err, users) {
+
+        // There shouldn't be any errors
+        should.not.exist(err);
+        users.should.be.instanceof(Array).and.have.lengthOf(0);
+
+        seed.start({ logResults: false, seedAdmin: admin1 }).then(function() {
```
ilanbiala: multiple lines

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -77,34 +122,152 @@ describe('Configuration Tests:', function () {
...
+
+      User.find({username: admin1.username}, function(err, users) {
 
         // There shouldn't be any errors
         should.not.exist(err);
         users.should.be.instanceof(Array).and.have.lengthOf(0);
 
-        seed.start().then(function() {
-          User.find({username: 'admin'}, function(err, users) {
+        seed.start({ logResults: false, seedAdmin: admin1, seedUser: user1 }).then(function() {
```
ilanbiala: multiple lines

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -118,6 +281,86 @@ describe('Configuration Tests:', function () {
...
+      process.env.NODE_ENV = 'production';
+
+      var _admin = new User(adminFromSeedConfig);
+
+      _admin.save(function(err, user) {
+        // There shouldn't be any errors
+        should.not.exist(err);
+        user.username.should.equal(adminFromSeedConfig.username);
+
+        seed.start({ logResults: false })
```
ilanbiala: new line

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -118,6 +281,86 @@ describe('Configuration Tests:', function () {
...
+
+      // Save original value
+      var nodeEnv = process.env.NODE_ENV;
+      // Set node env ro test environment
+      process.env.NODE_ENV = 'test';
+
+      var _user = new User(user1);
+      _user.email = '';
+
+      seed.start({ logResults: false, seedUser: _user })
```
ilanbiala: new line

mleanos:
@ilanbiala I addressed your line comments, with the exception of a couple.

The `resolve()` comments were on *non*-diffs. I don't think we need to `resolve` anything in these cases. We're just concerned with whether or not it was successful, or not. The data that we could return from these Promises wouldn't have any relevance outside the scope of the Promise; So I think it's appropriate to resolve with no data here.

As for the issue of the non-async tests using the `done` callback method, I think it's fine to use it here. It doesn't affect the functionality in any way, but I like the way it reads; it's more clear. Either way, I'm fine with removing them from all the tests in this file.

ilanbiala:
@mleanos if tests aren't async it's unnecessary, and it also indicates whether the tested function is synchronous or asynchronous, so let's make sure we aren't abusing it.

mleanos:
@ilanbiala Those resolves are to satisfy the current promise. You'll notice that each usage is at the end of the chain; we need to resolve each Promise, at each level. 

I'll go ahead and make those tests synchronous.

mleanos:
@ilanbiala Updated the tests. If we need to address the resolves, in regards to your concerns, it will require a bigger refactor of the SeedDB feature that we should do in a separate PR.

mleanos:
@ilanbiala @lirantal Does this look ready to you both?

ilanbiala:
@mleanos if we can't address the resolves in this PR, then let's open one right after because I don't think that is very intuitive. Otherwise, LGTM.

mleanos:
I would rather finalize this PR, and then have a discussion on how we want to proceed with our strategy of using Promises & Closures.

@lirantal Is that fine with you? I'll be on Gitter for a couple more hours, if you'd like to discuss.

lirantal:
Sure.

PR #967, opened 2015-10-05
[bug] Solves User model validation issue with tests [fixes #966]

### LINKED ISSUE #966: User CRUD Tests saving hashed password
While running the server tests, I'm periodically running into an issue with some of the User CRUD tests. I would get a model validation error on the `password` field. Each time, the validation was complaining about the "may not contain sequences of three or more..". I would re-run the tests and may, or may not, run into this issue again. 

Today, I decided to dig a bit deeper into this. What it looks like to me is the following tests are attempting to re-save the hashed password. Once I realized this, the issue appearing only periodically made sense. The hashed password isn't always going to contain 3 or more consecutive characters.

We fixed pretty much the same issue with the User model's `pre('save')` method. But I think we now have to add that same fix to the `pre('validate')` method using `this.isModified('password')`

This is only happening with these tests, since they re-save the user...
https://github.com/meanjs/mean/blob/master/modules/users/tests/server/user.server.routes.tests.js#L129-L265

@lirantal Would you agree with the suggested fix? I ask you because I know you're working directly with the User CRUD tests right now. You may have the most insight at this time.

@jloveland WDYT?

mleanos:
This solves the issue of the User model's `pre('validate')` method, attempting to validate against a password that was not modified.

Adds the `this.isModified('password')` check to the condition.

### DIFF
--- a/modules/users/server/models/user.server.model.js
+++ b/modules/users/server/models/user.server.model.js
@@ -116,7 +116,7 @@ UserSchema.pre('save', function (next) {
  * Hook a pre validate method to test the local password
  */
 UserSchema.pre('validate', function (next) {
-  if (this.provider === 'local' && this.password) {
+  if (this.provider === 'local' && this.password && this.isModified('password')) {
     var result = owasp.test(this.password);
     if (result.errors.length) {
       var error = result.errors.join(' ');


mleanos:
Fix for #966

lirantal:
LGTM and we should quickly push this fix or any alteration for it as this is a bug in `master` now.

jloveland:
LGTM, we should continue the conversation on #966 to make sure our user saving strategy is well understood.

mleanos:
This will definitely handle the bug. 

Lets merge it. And then we can continue our discussion about the strategy with the tests.

PR #972, opened 2015-10-10
Global Mocha timeout

mleanos:
Added the timeout option to the Mocha grunt task; set to 10000.

Removed the individual test suite timeouts, for all server tests.

As discussed in #964

Issue #964, opened 2015-10-05
mochaTest  timeout parameter


vaucouleur:
@mleanos  @lirantal 

Hi,

Worth considering: the timeout parameter can be set directly on the options of the mochaTest task. It avoids duplicating the this.timeout(10000) statements everywhere in the mocha tests.

Something like:

    mochaTest: {
      src: testAssets.tests.server,
      options: {
        reporter: 'dot',
        timeout: 10000
      }

See the Option paragraph  of https://github.com/pghalliday/grunt-mocha-test

mleanos:
My original idea was that it would be better to have the *timeouts* set per suite/test. That way when things stabilize with Travis, we can start customizing the timeouts; increasing/decreasing based on the observed performance for each test (or suite). Perhaps, that isn't the best strategy.

In our current situation everything is set to `10000` anyway, so it might make more sense to add this as a global timeout. Eventually, we'll want to decrease this timeout. Once that happens, we could just add a custom timeout to any tests that are giving us *trouble*.

@lirantal WDYT?

Gulp has the same option.. 
https://github.com/sindresorhus/gulp-mocha#timeout

vaucouleur:
@mleanos  Obviously, we can have both. Instead of mochajs default (2000), we can have a default value that fits better MEANJS needs, and if needed we can adjust the timeout for a specific test.

mleanos:
@vaucouleur I agree. I've tested this for Grunt & Gulp & they seem to work as expected. I'm fine with having this option added.

lirantal:
I think @vaucouleur has a valid point, let's add it as a global parameter for now instead of fiddling with each specific test scenario.

mleanos:
@vaucouleur I can submit a PR for this, unless you're planning to. 

Should we remove the individual test suite timeouts when we add this?

lirantal:
@mleanos @vaucouleur I think it's worth another PR change to move this to the grunt configuration.

vaucouleur:
@mleanos Hi Michael, please go ahead with the PR. If it turns out that you're busy with other things, just let me know and I'll do it.

mleanos:
This can be closed. I did a few rounds of testing after #972 was merged, and both Grunt & Gulp are working with this change.

PR #973, opened 2015-10-10
Global Mocha timeout option for Gulp

mleanos:
Added the timeout option for the Mocha gulp task.

This the Gulp counterpart to #972 & #964


mleanos:
Closing to combine with #972

PR #975, opened 2015-10-12
Added logging of time, username, and origin on failed signin.

bullwinklescousin:
Hi!  My name is Tim and I love what you're doing!  Thank you!  I want to help out but am not sure where to start.  

I noticed or missed locating any logging for failed signin's.  I thought this might help users secure their servers.  

Thoughts? Ideas or places I can start helping out?  Thanks again!

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -58,6 +58,7 @@ exports.signup = function (req, res) {
 exports.signin = function (req, res, next) {
   passport.authenticate('local', function (err, user, info) {
     if (err || !user) {
+      console.log(req._startTime + ', username: ' + req.body.username + ', from: ' + req.headers.origin);
       res.status(400).send(info);
     } else {
       // Remove sensitive data before login


lirantal:
Hi @bullwinklescousin,

That's great that you're joining in with us to collaborate on the development effort for the project.
Your PR actually had me wondering about the logging facility that we added a while back: https://github.com/meanjs/mean/blob/d41490c14860a135b3b0729b863d75eb01882f47/config/env/development.js#L15 but I'm not seeing it utilized anywhere.

Maybe I just overlooked it but my point is that we should move out of `console.log` all over the code base and instead use a logging facility like `morgan` which is already part of the framework.

codydaig:
@lirantal 
Are you saying that this line should utilize the config settings as well? https://github.com/meanjs/mean/blob/d41490c14860a135b3b0729b863d75eb01882f47/config/lib/express.js#L71

mleanos:
@codydaig It indeed looks like that line should be using `log.format` and `log.options` from the env config. Seems like the intended use for the env config log setting.

bullwinklescousin:
Okay! I thought I was just using the logging option incorrectly. I'm studying for a technical interview, but I will inspect things further and review open issues. Thanks!

lirantal:
@bullwinklescousin I think this logging would be beneficial but let's first sort out the bigger logging problem in #982 and then we can get back to this issue/PR.

codydaig:
@bullwinklescousin @lirantal #999 has been merged.

ilanbiala:
Coming back to this issue, I don't think logging on failed signins is something we should do. It is clutter and bots may just try to cause the server to run out of disk space and crash.

jloveland:
I think for security reasons and auditing this is a good idea. I would also want a future PR that would reset the users password and email them due to this activity. WDYT?

http://blog.securitymetrics.com/2015/02/hipaa-compliant-system-logs.html?m=1

codydaig:
@ilanbiala Handeling bots is probably something we should address. like captcha.

mleanos:
@jloveland I think now that we have logging capabilities, the auditing could be done through that, no? The failed login would be captured by the current logging system.

Having a feature that can detect certain behavior, via the logs, might be worthwhile. I just don't see the need to log this type of event to the console; since technically it already is with something like 
`POST /api/auth/signin 400 36.460 ms - 42` 
being logged already. We may want to be more explicit with the returned error code though.

jloveland:
@mleanos, we could audit the logs generated from Morgan to look for 400 or 401 code, but this wouldn't give us the username, only the client info right?
```
::1 - - [29/Oct/2015:01:05:56 +0000] "POST /api/auth/signin HTTP/1.1" 400 42 "https://localhost:8443/authentication/signin" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:41.0) Gecko/20100101 Firefox/41.0"
```

mleanos:
@jloveland You're right. We wouldn't have any useful info about the data sent to the API; username in this case.

Perhaps we could find a library that tracks these types of things, but I don't think a console log would be the right solution. If we see a specific IP making suspicious requests, we could add it to a *blacklist*.

jloveland:
@mleanos I agree. 
@bullwinklescousin have you researched other libraries that can help with this sort of thing?

bullwinklescousin:
I've been slacking since it was my birthday weekend, but I  will do some research by this afternoon. Personally I prefer verbose logging so fail2ban can ban ip's and forensics purposes, but perhaps logging's scope should be restricted and security modules might be better served by mongo. I'll do some research on security modules and solutions.

bullwinklescousin:
A time to live collection for failed auth attempts could be leveraged to minimize code to blacklist / ban after X attempts in the collection. After reading about the 401 PR, I'm tempted to close this in favor of that PR as I'm more concerned about having the ip available over the username (the logger didn't use the env files yet). It was just something I noticed as I was poking and prodding, but I didn't intend to bog us down providing features for the paranoid over something more useful / fun. Wdyt?

lirantal:
@bullwinklescousin would you like to further pursue more work here?

rhutchison:
close

lirantal:
@rhutchison glad to have you back! :)

I agree that this PR as-is is not even on a discussion to be merged, but I do think we can improve logging through the facility we've introduced, so if @bullwinklescousin is interested to invest more work here then I'm open to hear.

bullwinklescousin:
I'm out of my funk and interested in putting in work. I just don't know where to start. I guess I'll start with more research on libraries and then disabling an account after X failed attempts? Sorry for the post Christmas scatter brain.

lirantal:
@bullwinklescousin I don't think we need to mix different features and behaviors.
The topic of this PR is about better logging, but this PR as is can't be merged because we don't just do `console.log` to print debug info. There is a better logging facility which also logs to files.

If you're interested in doing more research and improving logging of information across the app then by all means you're welcome to integrate and build upon the existing architecture for logging and submit a new PR.

Mary Christmas :)

PR #982, opened 2015-10-13
[bug] Logger wasn't using the logger config option

codydaig:
To fix an issue that I brought up in #975. 

The morgan logger in express wasn't utilizing the logger config options, so this PR fixes that.

### DIFF
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -70,7 +70,7 @@ module.exports.initMiddleware = function (app) {
   // Environment dependent middleware
   if (process.env.NODE_ENV === 'development') {
     // Enable logger (morgan)
-    app.use(morgan('dev'));
+    app.use(morgan(config.log.format, config.log.options));
 
     // Disable views cache
     app.set('view cache', false);


mleanos:
LGTM

mleanos:
@ilanbiala Shouldn't this go in with 0.4.2?

This seems like the purpose of adding the `log` settings to the env config was to have this functionality. However, the implementation of the settings in the express middleware was overlooked. Making this a bug.

ilanbiala:
@mleanos whatever the case may be, this will change the functionality of morgan, so it is a breaking change. Anyway, after 0.4.2 we are moving to 0.5.0 to release other new features.

codydaig:
@ilanbiala How does this change the functionality of morgan?

jloveland:
+1 to get this merged in ASAP

ilanbiala:
@codydaig if I downloaded 0.4.1 and used this, then I download 0.4.2 and suddenly the logger works differently, that can cause issues for expected usage. This is a breaking change, even if it seems so insignificant. Anyway, we are jumping straight from 0.4.1 to 0.5.0, so it shouldn't matter too much in which milestone this is.

mleanos:
@ilanbiala If you see how Morgan is implemented currently, you'll see that it won't change how it works at all with this change.. It just changes where the *format* is coming from..

So rather than hard coding "dev" here https://github.com/meanjs/mean/blob/master/config/lib/express.js#L73
It will insert "dev" from the env config; Thus, it will not change the functionality. 
https://github.com/meanjs/mean/blob/master/config/env/development.js#L17

The only possibility of a change in functionality is the options, when in env/production. It is setting the `stream` option for Morgan. This was added for a reason, but if it could cause issues then we should comment that option out by default.

ilanbiala:
@mleanos sorry I misread the PR at first. @codydaig the PR should actually move the morgan setup outside of the env check completely and just to an if statement that checks if the log property is defined and if so adds morgan to the middlewares.

mleanos:
@ilanbiala You bring up a good point. I wonder if we want to fallback to using Morgan's *combined* format, if the env config `log.format` or `log` setting isn't set. If we opt to exclude Morgan in the express middleware, then there won't be anything logged with the requests & responses on the backend.

I think it could make sense to use Morgan no matter what. If a user doesn't want to use Morgan, then they could remove it from the Express config. 

WDYT?

mleanos:
@codydaig @ilanbiala I just noticed something else. Morgan is only being used when the NODE_ENV is *development*. What is the purpose of the `log` settings in the Production env config? Is there another intended use for the setting?

lirantal:
Hey guys,

I originally added this piece.
Here is my original PR for this functionality with the complete code and notes: https://github.com/meanjs/mean/pull/254/files

I believe that it may have gotten discarded when we did the merge from 0.3 to 0.4.
As you can see in the original PR - it always logs, irrelevant to being production or development, and it uses a `logger` wrapper with morgan.
Let me know WDYT

ilanbiala:
@mleanos that's what I was trying to say, we should move it out of that env check and just have it use the config log option and have a default option.

mleanos:
@lirantal The implementation in #254 is exactly what we need here. I was having issues with the Stream option, and it looks as though you solved it with the `logger` config. Too bad this got overlooked in the upgrade. But we got it back now :) 

@codydaig @ilanbiala Can we proceed with Liran's implementation?

ilanbiala:
That implementation looks good, though I don't know exactly why the fs.createReadStream is there.

mleanos:
@ilanbiala That `createWriteStream` is a requirement of the Stream option. 
https://github.com/expressjs/morgan#write-logs-to-a-file

I ran into the issue when testing this PR when setting the NODE_ENV to Production. It will throw an error if there isn't a WriteStream set.

lirantal:
@codydaig do you want to check if you can incorporate my old work at #254 into this PR?

codydaig:
@mleanos Is going to take lead on that. Once he gets a new PR submitted, we can close this one.

codydaig:
Closing in favor of #999

PR #985, opened 2015-10-14
Adding karma coverage to grunt

jloveland:
Adding karma coverage for displaying client side test coverage when running `grunt coverage`.
karma-coverage

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -5,7 +5,13 @@
  */
 var _ = require('lodash'),
   defaultAssets = require('./config/assets/default'),
-  testAssets = require('./config/assets/test');
+  testAssets = require('./config/assets/test'),
+  testConfig = require('./config/env/test'),
+  karmaReporters = ['progress'];
+
+if (testConfig.coverage) {
+  karmaReporters.push('coverage');
+}
 
 // Karma configuration
 module.exports = function (karmaConfig) {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -82,6 +82,7 @@
     "grunt-ng-annotate": "^1.0.1",
     "grunt-node-inspector": "~0.3.0",
     "grunt-nodemon": "~0.4.0",
+    "grunt-protractor-coverage": "^0.2.15",
```

karma.conf.js
```diff
@@ -30,7 +43,27 @@ module.exports = function (karmaConfig) {
 
     // Test results reporter to use
     // Possible values: 'dots', 'progress', 'junit', 'growl', 'coverage'
-    reporters: ['progress'],
+    reporters: karmaReporters,
```
ilanbiala: Why did you move this to a variable?

karma.conf.js
```diff
@@ -30,7 +43,27 @@ module.exports = function (karmaConfig) {
...
+
+    // Configure the coverage reporter
+    coverageReporter: {
+      dir : 'coverage/',
+      reporters: [
```
ilanbiala: Don't we only need html/lcov?

ilanbiala:
@jloveland good start. Can you also add Gulp support?

karma.conf.js
jloveland: I added the variable so I can test for `testConfig.coverage === true` which is set by the grunt coverage task [here](https://github.com/meanjs/mean/pull/985/files#diff-595bf3fa2348192244b0319be33066b8R227)

karma.conf.js
jloveland: we can add or remove options here. I think we can get away with html and text only. Text will output to the console for us. Does anyone want lcov?

jloveland:
@ilanbiala thanks for the review. 

I would like to have the gulp karma coverage added in a different PR. Rationale is the following:
1. Mean.js doesn't fully support gulp yet
2. We don't even have a gulp coverage task for the server yet.

WDYT?

jloveland:
also, what are your thoughts on outputting reports for client and server in `coverage/client` and `coverage/server` respectively?

lirantal:
:+1: on adding coverage report for the client side
Sounds like a good idea to separate them for the future. People might be interested only in one aspect (client vs server) or will want to do different things with them in CI/CD process.

ilanbiala:
:+1:

jloveland:
@lirantal and @ilanbiala fixed conflicts, this should be ready to merge.

ilanbiala:
@lirantal should we see coverage in Coveralls already, or will we have to merge the PR?

lirantal:
IIRC you must merge first because it's just tracking the `master` branch.

PR #987, opened 2015-10-15
[bug] Favicon invalid path [fixes #979]

### LINKED ISSUE #979: Favicon incorrect path
I've experienced an issue with the favicon resolving to an incorrect path. It happens intermittently, and I'm not able to reproduce regularly. However, I have seen others reporting the same issue.

When the issue occurs, the favicon URL resolves to `http://localhost:3000/articlesmodules/core/client/img/brand/favicon.ico` or similar; this will depend on what route I navigate to. So in this example, I was on the Articles route. I've seen it when I was on the Profile route as well.

The problem seems to be with this line here..
https://github.com/meanjs/mean/blob/master/modules/core/server/views/layout.server.view.html#L32

If I remove the `{{url}}` from the `href`, I am unable to reproduce the issue; and not for a lack of effort. Can someone shed some light on why the `{{url}}` was added to this path? 

Also, I'm confused as to what these META tags purposes are. 
https://github.com/meanjs/mean/blob/master/modules/core/server/views/layout.server.view.html#L17-L29

It seems that the `{{url}}` was definitely put there for a specific purpose. I can only guess that it's to accommodate an environmental change. 

I think the favicon path shouldn't need the `{{url}}`, but I'm not sure of the other above mentioned uses. 

I'd like to fix this, but before I submit a possible fix for the favicon, I'd like to get a bit more context & insight from others.

mleanos:
Removed `{{url}}` from the Favicon path. This fixes the intermittent issues with the path resolving to an invalid location.

This seems to fix the bug with the Favicon invalid path. However, as discussed in #979, there are still remaining questions as to why these lines are using `locals.url` and `locals.logo` with the META tags
https://github.com/meanjs/mean/blob/master/modules/core/server/views/layout.server.view.html#L17-L29

Does anyone have any insight into these, and how we should address these issues? I think it would be appropriate to address all these issues in this PR. We just need to decide on how.

### DIFF
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+﻿<!DOCTYPE html>
 <html lang="en">
 <head>
   <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


mleanos:
@lirantal I removed the `{{url}}` from the twitter:image & og:image META tags. 

Can someone shed some light on what `local.url` is intended for? 
https://github.com/meanjs/mean/blob/master/config/lib/express.js#L44

Is this for redirection purposes? In order to know where the user came from?

codydaig:
LGTM

lirantal:
LGTM @mleanos 
As to the locals question: http://expressjs.com/api.html#res.locals - it was added way back from Amos's 0.4 branch IIRC.

mleanos:
@lirantal Thank you for clarifying that. I've rebased & squashed.

If we need to address any of the other META tags, then I can do so. Otherwise, I think this is ready.

PR #989, opened 2015-10-15
Add .eslintrc

### LINKED ISSUE #763: Request for JSCS
Hi @rhutchison,

Can you please create a PR for the adding the JSCS (https://github.com/jscs-dev/node-jscs) support? I'd like to push this in along with JSHint that we have today to enforce the code style standards.

I know that you included part of that in another PR that we asked to take out, so there's no point for me to re-do any of the work you already have pending there.
Thanks!

ilanbiala:
Fixes #763
@lirantal @codydaig @rhutchison

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -66,5 +66,28 @@ module.exports = {
     }
   },
   livereload: true,
-  seedDB: process.env.MONGO_SEED || false
+  seedDB: {
+    seed: process.env.MONGO_SEED === 'true' ? true : false,
+    options: {
+      logResults: process.env.MONGO_SEED_LOG_RESULTS === 'false' ? false : true,
+      seedUser: {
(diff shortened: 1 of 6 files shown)


ilanbiala:
So I added Grunt and Gulp support, which means that we now get errors in our builds. The question is whether we want that to fail our builds (which I think should be the case). If so, I suggest we devote another PR to fixing those issues and make this PR strictly for adding in ESLint.

trainerbill:
Yes it should fail the build and we should submit PR's to fix the entire project.

codydaig:
@ilanbiala I think it should fail the build. However, do we want to fix the issues before we force it to fail? I'd hate for this to get merged in and have every PR after this fail because of eslint errors in other parts of the code.

mleanos:
:+1: On fixing the issues in separate PR(s) before this gets merged.

lirantal:
Agree with the above.

Issue #961, opened 2015-10-05
E2E Tests error out but don't fail the build


lirantal:
```
[4mRunning "protractor:e2e" (protractor) task[24m
webdriver-manager path: /home/travis/build/meanjs/mean/node_modules/grunt-protractor-runner/node_modules/protractor/bin/webdriver-manager
Updating selenium standalone
downloading https://selenium-release.storage.googleapis.com/2.45/selenium-server-standalone-2.45.0.jar...
Updating chromedriver
downloading https://chromedriver.storage.googleapis.com/2.15/chromedriver_linux64.zip...
chromedriver_2.15.zip downloaded to /home/travis/build/meanjs/mean/node_modules/grunt-protractor-runner/node_modules/protractor/selenium/chromedriver_2.15.zip
selenium-server-standalone-2.45.0.jar downloaded to /home/travis/build/meanjs/mean/node_modules/grunt-protractor-runner/node_modules/protractor/selenium/selenium-server-standalone-2.45.0.jar
Starting selenium standalone server...
[launcher] Running 1 instances of WebDriver
Selenium standalone server started at http://172.17.0.142:49446/wd/hub

/home/travis/build/meanjs/mean/node_modules/grunt-protractor-runner/node_modules/protractor/node_modules/selenium-webdriver/lib/atoms/error.js:113
  var template = new Error(this.message);
                 ^
UnknownError: unknown error: Chrome failed to start: crashed
  (Driver info: chromedriver=2.15.322448 (52179c1b310fec1797c81ea9a20326839860b7d3),platform=Linux 3.13.0-40-generic x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 60.14 seconds
Build info: version: '2.45.0', revision: '5017cb8', time: '2015-02-26 23:59:50'
System info: host: 'testing-worker-linux-docker-207121c9-3188-linux-3', ip: '172.17.0.142', os.name: 'Linux', os.arch: 'amd64', os.version: '3.13.0-40-generic', java.version: '1.7.0_76'
Driver info: org.openqa.selenium.chrome.ChromeDriver
    at new bot.Error (/home/travis/build/meanjs/mean/node_modules/grunt-protractor-runner/node_modules/protractor/node_modules/selenium-webdriver/lib/atoms/error.js:113:18)
[...]

PR #993, opened 2015-10-16
Adding configuration to fail grunt task if e2e tests fail

jloveland:
fixing #961, failing grunt task will fail the travis build which is the desired behavior

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -201,7 +201,6 @@ module.exports = function (grunt) {
     protractor: {
       options: {
         configFile: 'protractor.conf.js',
-        keepAlive: true,
         noColor: false,
         webdriverManagerUpdate: true
       },


jloveland:
@codydaig, I purposely failed the build so you can see this config works.

codydaig:
LGTM
@jloveland Thanks! :-D

mleanos:
LGTM too! Thanks!

I am wondering, and just throwing this out there, if we want to add the `keepAlive` setting into the env config? How useful would this be for development purposes?

gruntfile.js
```diff
@@ -201,7 +201,8 @@ module.exports = function (grunt) {
     protractor: {
       options: {
         configFile: 'protractor.conf.js',
-        keepAlive: true,
+        // Stops Grunt process if a test fails
+        keepAlive: false,
```
ilanbiala: The default is false, right? I'd rather just remove it because it's the behavior we expect and Grunt isn't as clear as it could be anyway.

codydaig:
@mleanos I don't see a use case where that would be beneficial of the top of my head. Since the e2e tests run last, you'd still find out if any of your server or client tests failed first before hitting the e2e tests. When would you want to a build to pass even though the e2e tests weren't?

mleanos:
@codydaig I guess my thought was that in a certain environment (development/staging), you may want the Grunt task to continue running even when a E2E test fails.

The idea was mostly just an afterthought :)

jloveland:
For the `keepAlive` config, I was thinking it would be handy to have it there, then add another PR for adding grunt watch so, we can develop e2e tests have it continually run grunt tests after saving the test files..
https://www.credera.com/blog/technology-insights/java/testing-angularjs-part-5-protractor-grunt/

codydaig:
@ilanbiala Are you okay with `keepAlive` being left in there, or do you want it removed before I merge?

ilanbiala:
Remove it please.

jloveland:
@ilanbiala and @codydaig will do.

codydaig:
Thanks @jloveland! :-)

PR #999, opened 2015-10-17
[bug] Enable log options for Morgan that were missed in the 0.4.0 merge

mleanos:
Adds the log options, and format to the Morgan middleware in the Express configuration.

These options are defined in the environment configurations.

The implementation derived from https://github.com/meanjs/mean/pull/254 by @lirantal, which somehow got overlooked when merging 0.4.0 into master.

Related to the discussion in #975, and taking over for @codydaig #982

### DIFF
--- a/config/env/test.js
+++ b/config/env/test.js
@@ -12,6 +12,25 @@ module.exports = {
     // Enable mongoose debug mode
     debug: process.env.MONGODB_DEBUG || false
   },
+  log: {
+    // logging with Morgan - https://github.com/expressjs/morgan
+    // Can specify one of 'combined', 'common', 'dev', 'short', 'tiny'
+    format: process.env.LOG_FORMAT || 'combined',
+    options: {
+      // Stream defaults to process.stdout
+      // Uncomment/comment to toggle the logging to a log on the file system
+      stream: {
+        directoryPath: process.cwd(),
+        fileName: 'access.log',
+        rotatingLogs: { // for more info on rotating logs - https://github.com/holidayextras/file-stream-rotator#usage
+          active: false, // activate to use rotating logs 
+          fileName: 'access-%DATE%.log', // if rotating logs are active, this fileName setting will be used
+          frequency: 'daily',
+          verbose: false
+        }
+      }
+    }
+  },
   port: process.env.PORT || 3001,
   app: {
     title: defaultEnvConfig.app.title + ' - Test Environment'
--- a/config/lib/logger.js
+++ b/config/lib/logger.js
@@ -1,0 +1,109 @@
+﻿'use strict';
+
+var _ = require('lodash'),
+  config = require('../config'),
+  chalk = require('chalk'),
+  fileStreamRotator = require('file-stream-rotator'),
+  fs = require('fs');
+
+// list of valid formats for the logging
+var validFormats = ['combined', 'common', 'dev', 'short', 'tiny'];
+
+// build logger service
+var logger = {
+  getFormat: getLogFormat, // log format to use
+  getOptions: getLogOptions // log options to use
+};
+
+// export the logger service
+module.exports = logger;
+
+/**
+ * The format to use with the logger
+ *
+ * Returns the log.format option set in the current environment configuration
+ */
+function getLogFormat () {
+  var format = config.log && config.log.format ? config.log.format.toString() : 'combined';
+
+  // make sure we have a valid format
+  if (!_.includes(validFormats, format)) {
+    format = 'combined';
+
+    if (process.env.NODE_ENV !== 'test') {
+      console.log();
+      console.log(chalk.yellow('Warning: An invalid format was provided. The logger will use the default format of "' + format + '"'));
+      console.log();
+    }
+  }
+
+  return format;
+}
+
+/**
+ * The options to use with the logger
+ *
+ * Returns the log.options object set in the current environment configuration.
+ * NOTE: Any options, requiring special handling (e.g. 'stream'), that encounter an error will be removed from the options.
+ */
+function getLogOptions () {
+  var options = config.log && config.log.options ? _.clone(config.log.options, true) : {};
+
+  // check if the current environment config has the log stream option set
+  if (_.has(options, 'stream')) {
+
+    try {
+
+      // check if we need to use rotating logs
+      if (_.has(options, 'stream.rotatingLogs') && options.stream.rotatingLogs.active) {
+
+        if (options.stream.rotatingLogs.fileName.length && options.stream.directoryPath.length) {
+
+          // ensure the log directory exists
+          if (!fs.existsSync(options.stream.directoryPath)) {
+            fs.mkdirSync(options.stream.directoryPath);
+          }
+
+          options.stream = fileStreamRotator.getStream({
+            filename: options.stream.directoryPath + '/' + options.stream.rotatingLogs.fileName,
+            frequency: options.stream.rotatingLogs.frequency,
+            verbose: options.stream.rotatingLogs.verbose
+          });
+
+        } else {
+          // throw a new error so we can catch and handle it gracefully
+          throw new Error('An invalid fileName or directoryPath was provided for the rotating logs option.');
+        }
+
+      } else {
+
+        // create the WriteStream to use for the logs
+        if (options.stream.fileName.length && options.stream.directoryPath.length) {
+
+          // ensure the log directory exists
+          if (!fs.existsSync(options.stream.directoryPath)) {
+            fs.mkdirSync(options.stream.directoryPath);
+          }
+
+          options.stream = fs.createWriteStream(options.stream.directoryPath + '/' + config.log.options.stream.fileName, { flags: 'a' });
+        } else {
+          // throw a new error so we can catch and handle it gracefully
+          throw new Error('An invalid fileName or directoryPath was provided for stream option.');
+        }
+      }
+    } catch (err) {
+
+      // remove the stream option
+      delete options.stream;
+
+      if (process.env.NODE_ENV !== 'test') {
+        console.log();
+        console.log(chalk.red('An error has occured during the creation of the WriteStream. The stream option has been omitted.'));
+        console.log(chalk.red(err));
+        console.log();
+      }
+    }
+  }
+
+  return options;
+}
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "crypto": "0.0.3",
     "express": "^4.13.1",
     "express-session": "^1.11.3",
+    "file-stream-rotator": "~0.0.6",
     "forever": "~0.14.2",
     "generate-password": "^1.1.1",
     "glob": "^5.0.13",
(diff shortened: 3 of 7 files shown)


mleanos:
I'm aware of the coverage decrease. However, I just wanted to get this out there for review & work on writing tests in the meantime. I shall do that now :)

mleanos:
I wrote two tests initially for this feature. However, I'm unsure if I went about this correctly. They just seem so simple. Although, the testability of a feature like this seems limited.

It may not make sense to have the log options set in the Test env config. Does anyone have any feedback, or insight into how to enhance the test coverage here?

This brings up an interesting idea I was pondering a while ago.. Would be worthwhile to add a feature to *reinitialize* the global configuration? In instances where the NODE_ENV is manually changed (most likely for test coverage only), it would be nice to test features against the updated configuration settings. Any thought on this?

mleanos:
I reworked the tests, and cleaned up the default settings for the log options & format.

I think this last commit is better, but still would love some feedback.

lirantal:
@mleanos I did a similar "trick" with changing the NODE_ENV from testing to production through out the tests to make sure that the config works properly.

Otherwise LGTM.

mleanos:
@lirantal I liked your strategy of changing the NODE_ENV manually. However, that won't work in this case. Once the logger is required, the config is initialized. Changing the node env will have no impact aftward. 

What if I added a `init` method to the logger service? There I could add the `require` of the config. Is that a route worth going down?

lirantal:
Don't think so because that's usually a bad practice to `require` within functions/if because node can't cache that and whatever other side effects it will have.
We can go with what we have for now until we come up with another strategy.

mleanos:
SGTM. Thank you. 

Any opinion on if we should add the log settings to the Test env config? This way we can test the Stream option as well.

config/lib/logger.js
```diff
@@ -0,0 +1,46 @@
...
+  var options = config.log && config.log.options ? config.log.options : {};
+
+  // check if the current environment config has the log stream option set
+  if (_.has(options, 'stream')) {
+    try {
+      // create the WriteStream to use for the logs
+      options.stream = fs.createWriteStream(process.cwd() + '/' + config.log.options.stream, { flags: 'a' });
+    } catch (err) {
+      // remove the stream option
+      delete options.stream;
```
codydaig: Do we want to log an error to the console instead of just silently catching and handeling the error?

mleanos: That's a good point. I think logging to the console would be appropriate here.

config/lib/logger.js
```diff
@@ -0,0 +1,46 @@
...
+// export the logger service
+module.exports = logger;
+
+/**
+ * The format to use with the logger
+ *
+ * Returns the log.format option set in the current environment configuration
+ */
+function getLogFormat () {
+  return config.log && config.log.format ? config.log.format : 'combined';
```
codydaig: Why is this defaulting to combined? The current default is 'dev' so we want to change that?

mleanos:
I just committed quite a bit of changes. I added a bit more error handling, and modified some logic of how I was returning the log settings. For instance, before I was returning references to the env config settings. I think that was wrong. Now I'm making copies.

I also added a couple more tests, and added the log settings to the Test env config.

Further review would be greatly appreciated.

config/env/test.js
```diff
@@ -12,6 +12,15 @@ module.exports = {
     // Enable mongoose debug mode
     debug: process.env.MONGODB_DEBUG || false
   },
+  log: {
+    // Can specify one of 'combined', 'common', 'dev', 'short', 'tiny'
+    format: process.env.LOG_FORMAT || 'combined',
+    // Stream defaults to process.stdout
+    // Uncomment to enable logging to a log on the file system
```
jloveland: @mleanos what is this comment for? doesn't the option below log to the file system?

mleanos: It was just a left over comment from when I copied the settings from the development config. 

I can remove it, but it also does make sense to leave it?

lirantal: It seems confusing. Let's move the comments in line 18 to https://github.com/mleanos/mean/blob/morgan-logger-config/config/env/test.js#L21 inside the options object and above the stream option.

mleanos: @jloveland @lirantal I moved the comment, and re-worded it. Let me know if you'd like additional changes to it. Thanks.

mleanos:
@codydaig I have rebased, and squashed. If there's no other feedback, or requested changes, then I think this is ready.

config/lib/logger.js
```diff
@@ -0,0 +1,80 @@
...
+  var options = config.log && config.log.options ? _.clone(config.log.options, true) : {};
+
+  // check if the current environment config has the log stream option set
+  if (_.has(options, 'stream')) {
+
+    try {
+
+      // create the WriteStream to use for the logs
+      if (options.stream.length) {
+        options.stream = fs.createWriteStream(process.cwd() + '/' + config.log.options.stream, { flags: 'a' });
```
jloveland: @mleanos, great job on this!

I think it would be more robust to provide an ability to specify the full path to the directory which we would like logs to be stored. For example, I like my logs to go to `/var/log/meanjs`. Your code assumes the root directory if the cwd.

jloveland:
I think it would be nice add log rotation configs to the `config.env.production.js` and have that propagate to the `config.lib.logger.js`
```js
// create a rotating write stream
var accessLogStream = FileStreamRotator.getStream({
  filename: logDirectory + '/access-%DATE%.log',
  frequency: 'daily',
  verbose: false
})

// setup the logger
app.use(morgan('combined', {stream: accessLogStream}))
```

lirantal:
@jloveland sounds like a good diea
@mleanos can we add the logrotating functionality?

config/lib/logger.js
mleanos: @jloveland That's a good idea. I'll work on adding this. 

I think moving the definition of the log location to be in the env config would be the way to go. This way the the `logger` config won't need to *calculate* the location of the logs.

mleanos:
@jloveland @lirantal The log rotating is a good idea. 

I'll have to reconfigure the stream options, but I think it can be done to easily allow switching between both log rotating, and single file log (or even possibly allow for both options).

mleanos:
@lirantal @jloveland I started looking at how I could change the Stream options in the env configs. Here's my initial idea... Any thoughts on this approach?

```javascript
stream: {
  single: {
    active: true,
    directoryPath: process.cwd(),
    fileName: 'access.log'
  },
  rotating: {
    active: true,
    directoryPath: 'c:/windows/logs/meanjs',
    fileName: 'access-%DATE%.log',
    frequency: 'daily',
    verbose: false
  }
}
```
We could use these settings to build the options in the `logger` config that is passed along to Morgan.

mleanos:
I guess the above won't work, since I can't pass both streams into Morgan. I'll keep at it.

mleanos:
@jloveland @lirantal I have implemented the Rotating Logs functionality. This still needs a little bit of work, and I need to write tests for the Rotating logs. But I wanted to get this out for feedback.

I had to re-work the Stream options to account for the Rotating Logs. How is this new implementation looking?

I've also refactored the tests to be more maintainable. There's one test in particular that I'm having an issue with. I have disabled this test for now, but would love to know if anyone has any insight into this. I'm getting a ENOTEMPTY error from the `fs.rmdir` call.

codydaig:
@mleanos I think you have to empty the directory before you can remove the directory.

mleanos:
@codydaig I'm using `unlink` before the `fs.rmdir` call to get rid of the file I added. Shouldn't that be sufficient?

codydaig:
@mleanos I would assume so as well. Sorry, I didn't get a chance to look at the code before I made that comment, just going off the error message.

lirantal:
Quite a lot of logic to add for the log rotation. Is there no package for that on npm that we can rely on and complicate the logic less?

mleanos:
@lirantal What logic are you referring to?

Don't we want to offer the option of using a single file log, or a rotating log?

lirantal:
Yes that's the point. I was wondering if there's some package to handle that so we don't need to start messing around with deleting files, creating directories, etc.

package.json
```diff
@@ -34,6 +34,7 @@
     "crypto": "0.0.3",
     "express": "^4.13.1",
     "express-session": "^1.11.3",
+    "file-stream-rotator": "0.0.6",
```

mleanos:
@lirantal I agree. I'm looking into possible packages to use. One problem is the unit testing though. If we want to write Unit Tests against the logging, then we will need to remove the test logs after the tests are complete. I'm not seeing a solution to that particular issue, unless a package offers this out of the box.

jloveland:
I tested these features and they LGTM!

mleanos:
@lirantal I ended up implementing `mock-fs` into the tests here; using this we don't have to worry about cleaning up files/directories after the tests. 

Thank you for testing with me @jloveland It's much appreciated :) 

We still have to make sure the provided log directory is created, if it doesn't exist. There's no way around it, but I don't think it will be an issue.
https://github.com/meanjs/mean/pull/999/files#diff-c31b71d5f97621da6a2b4c696f198f49R63
https://github.com/meanjs/mean/pull/999/files#diff-c31b71d5f97621da6a2b4c696f198f49R84

@codydaig @ilanbiala @lirantal @jloveland Let me know what you think of these changes, and if we're satisfied I can squash.

codydaig:
LGTM

package.json
```diff
@@ -112,6 +113,7 @@
     "karma-ng-html2js-preprocessor": "^0.1.2",
     "karma-phantomjs-launcher": "~0.2.0",
     "load-grunt-tasks": "^3.2.0",
+    "mock-fs": "^3.4.0",
```

ilanbiala:
@mleanos pretty much LGTM, just looking at tests and I'm wondering if we can add a bit more coverage for those edge cases?

https://coveralls.io/builds/3890174/source?filename=..%2Fconfig%2Flib%2Flogger.js

codydaig:
@ilanbiala @mleanos Something along these lines should work. I hope I remember the exact syntax correctly. 

```javascript
var spy = sinon.spy(console, 'log');

spy.firstCall.calledWith();
spy.secondCall.calledWith(chalk.yellow('Warning: An invalid format was provided. The logger will use the default format of "' + format + '"'));
spy.ThirdCall.calledWith();

console.log.restore();
```

mleanos:
@codydaig Thank you for the suggestion. I'll try to implement this. We have a few other areas of the app where this would be useful.

mleanos:
@codydaig @ilanbiala I looked into using Sinon, and it was quite handy. It turns out that using Sinon's sandbox feature, it's possible to *mute* all console logs for the duration of a test/test-suite. I think this will really be useful in a few different areas of our application.

I'd like to look at incorporating the muting of the console logs into our test coverage, but more on a global test coverage scope. Let's not worry about adding the extra coverage for the lines that log to console in this PR. I think it would be better to address it separately.

@ilanbiala I addressed the `mock-fs` dependency comment you left.

I have rebased & squashed since this should be ready now.

ilanbiala:
Did another once over, LGTM.

codydaig:
LGTM

PR #1008, opened 2015-10-20
[bug] Dont use Cache

codydaig:
Having cache on the build is gibing us a false positive. As soon as I remove the cache, it will use the latest npm modules, and let us know when it's truly failing because of package dependencies.

I cloned the repo into a new directory and I'm getting the same mongo error locally as this PR is able to duplicate on the server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,3 @@ notifications:
     on_success: change  # options: [always|never|change] default: always
     on_failure: always  # options: [always|never|change] default: always
     on_start: never     # options: [always|never|change] default: always
-cache:
-  apt: true
-  directories:
-  - node_modules/
-  - public/lib/


codydaig:
This build will pass once #1009 gets merged in and I rebase.

codydaig:
@lirantal This actually removes the cache that you added to show us if a package updated and causes things to break in the project.

ilanbiala:
@codydaig do we still need this?

codydaig:
@ilanbiala I would still like this. It will allow us to see when updated dependencies are failing the build versus just using outdated dependences.

ilanbiala:
@codydaig when do we get outdated dependencies though? I haven't really seen it other than with Mongoose.

codydaig:
@ilanbiala Any time a package updates, it won't download the latest on Travis, it will use the cached version. This is just the first time that a package released a patch that broke our code.

ilanbiala:
@codydaig sent an email to Travis-CI, let's see if they have any plans to check against npm and see whether new versions are available before using the cache.

lirantal:
@codydaig  Ok, though this should stop 0.4.2 from being release so let's move it to 0.5.0 milestone?

codydaig:
@ilanbiala IMO, regardless of what Travis' plans are, this should get merged in and if Travis releases support to check against npm versions, then we can readdress and add cache support back in. 

@lirantal This is either a merge or don't merge. If we don't merge it then we should just close it out.

ilanbiala:
@codydaig right now Travis doesn't properly verify npm deps, so let's merge and I'll keep following up with them to see if they have plans to change it so we can cache.

PR #1009, opened 2015-10-20
[hotfix] Lock mongoose version to a working version

codydaig:
MAJOR BUG!!
(the app does not run until this is merged in)
The latest release of mongoose this morning broke changes in the repo. #1008 proves this by removing the cache in travis and forcing travis to use the latest version of mongoose.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "lodash": "^3.10.0",
     "method-override": "^2.3.3",
     "mocha": "~1.20.0",
-    "mongoose": "^4.0.6",
+    "mongoose": "4.1.11",
     "morgan": "^1.6.1",
     "multer": "0.1.8",
     "nodemailer": "^1.4.0",


mleanos:
LGTM.

mleanos:
We should consider locking down Mongoose to a specific version going forward. Since mongoose is such a critical part of the project. I've had a lot of issues with certain upgrades in the past.

package.json
```diff
@@ -44,7 +44,7 @@
     "lodash": "^3.10.0",
     "method-override": "^2.3.3",
     "mocha": "~1.20.0",
-    "mongoose": "^4.0.6",
+    "mongoose": "4.1.11",
```
ilanbiala: @codydaig which change exactly caused this issue. I don't see anything in their changelog that is marked as breaking or very relevant to the error. Do you know of a specific issue in the Mongoose repo?

codydaig: From what I was discovering, it's an issue with the mongodb package.

ilanbiala: Definitely, but looking into their commit history maybe this https://github.com/Automattic/mongoose/commit/9af0d5fb0c5ffd4f8382715a1837e3c110b4ed3b

codydaig: Could have been in there. Their commit history is funky to read with the merging of 3.8.x into master between the 4.1.11 and 4.1.12 releases. I'll play and see if I can break it.

mleanos: It's definitely an authentication issue. I think we should require the `user` and `pass` options to be passed along to the `connect` method.

After reviewing it further last night, I came to the conclusion that this isn't a bug. It appears to be a limitation of how we're handling authentication.

I'm don't have any issues with Mongoose `4.1.12` locally when running with the Development env config, because I use local.js to manage the `db` setting; I'm setting `user` and `pass` there. However, I did run into an issue with running in Test env. Because the `db` setting is not using `user` and `pass`. 

Thus, why I added #1011

codydaig:
@mleanos If you have a local instance of mongodb running that doesn't require authentication, it's broken on 4.1.12. It also breaks on travis (see the PR where I removed the cache #1008). It's working for you because your using mongolab that requires a user and pass.

PR #1012, opened 2015-10-20
[WIP] [feat] Mongos new Auth Method

codydaig:
@mleanos @ilanbiala This is what broke from 4.1.11 to 4.1.12. They want the auth method to be specified and the user and pass fields only work if you actually have a user and pass.

### DIFF
--- a/config/env/cloud-foundry.js
+++ b/config/env/cloud-foundry.js
@@ -15,8 +15,11 @@ module.exports = {
   db: {
     uri: cfMongoUrl,
     options: {
-      user: '',
-      pass: ''
+      auth: {
+        authMechanism: ''
+      },
+      // user: '',
+      // pass: ''
     }
   },
   log: {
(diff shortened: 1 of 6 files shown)


codydaig:
@ilanbiala Stand-bye. There's still some hashing to do on this PR. @mleanos Are discussing it.

ilanbiala:
@codydaig @mleanos any progress?

codydaig:
@ilanbiala I'm still investigating the best method for moving forward. Regardless, I want to hold this off until 0.5 since it will most likely implement a breaking change. 

In the meantime, mongoose released another patch that doesn't break this anymore so we could go back to `~` in package.json and it should work but I have yet to test and confirm this.

ilanbiala:
@codydaig Definitely would rather use `~`, so let's set up a PR with that and see if Travis passes?

mleanos:
Just to follow up here, even though I already stated this in #1020 

I've been testing Mongoose 4.2.3, and have confirmed that this issue has been resolved in that release. Our current implementation works just as it did before their breaking change was released. 

IMO, this isn't really relevant anymore. We may need to pay attention to the upcoming Mongoose 5.0 release, and make sure we stay on top of any changes there. I have seen some discussion on them introducing some ES6 features.

PR #1014, opened 2015-10-22
Changed favicon string to env variable

rodp82:


ilanbiala:
@lirantal @codydaig LGTM. Whoever gets to this first, just make sure it's fine and then feel free to merge.

config/lib/express.js
```diff
@@ -65,7 +65,7 @@ module.exports.initMiddleware = function (app) {
   }));
 
   // Initialize favicon middleware
-  app.use(favicon('./modules/core/client/img/brand/favicon.ico'));
+  app.use(favicon(app.locals.favicon));
```
codydaig: @ilanbiala `app.locals.favicon` is set to `modules/core/client/img/brand/favicon.ico`
So, does this still need to prepend `./`?

ilanbiala:
@codydaig yes possibly, though maybe a path.resolve or something is better? Also, maybe there should be a test for different favicons based on the environment?

codydaig:
@rodp82 Can you make the changes to use `path.resolve` and the correct path using `./` and I think this should be good to go.

config/env/default.js
```diff
@@ -28,5 +28,5 @@ module.exports = {
   sessionKey: 'sessionId',
   sessionCollection: 'sessions',
   logo: 'modules/core/client/img/brand/logo.png',
-  favicon: 'modules/core/client/img/brand/favicon.ico'
+  favicon: './modules/core/client/img/brand/favicon.ico'
```
codydaig: I think this will break the other areas this is used. @mleanos Am I correct on that.

I think the `./` will have to be added onto it.

mleanos: @codydaig I'll look into this over the weekend.

My initial thought, is that the `./` isn't needed since `path.resolve` is being used. I'll confirm this when I test.

mleanos: @codydaig @rodp82 I don't see where else this is used. Have you seen it used somewhere else in the application? 

As for the question of whether we need `./`, the answer is no. We don't need it. The `favicon` library doesn't require it, so we're probably better off not changing this in case others have used this setting in other areas of the the application.

After testing, I am questioning why we're using `path.resolve` here https://github.com/meanjs/mean/pull/1014/files#diff-16e0a4c137ae50cf68b1d2c3c914be85R40

This will set the URL to an absolute path, which would expose the directory structure of where this file is located. We probably don't want that, and resolving the path isn't necessary for these purposes.

IMO, this is the only change that is required in this PR https://github.com/meanjs/mean/pull/1014/files#diff-16e0a4c137ae50cf68b1d2c3c914be85R69

I have tested the favicon, with just this change. Everything is working as expected.

rodp82: @mleanos @codydaig So you happy for me to revert those changes back to the original PR?

codydaig: @mleanos Here's where else it is used: https://github.com/meanjs/mean/blob/master/modules/core/server/views/layout.server.view.html#L32

mleanos: @codydaig I should have clarified my response. That's the only place I see that actually leverages that locals variables. I didn't know if you saw it somewhere else.

@rodp82 I think that one change, is all we need. It will be much better to not have that hard-coded.

@codydaig WDYT?

ilanbiala: The ./ is probably unnecessary. I'm wondering whether path.resolve is necessary when we pass it into favicon()

mleanos: @ilanbiala I agree. See my above comment.

rodp82: So, just curiously, is it better to revert my changes back to the original PR (which reflects just that one change), or make the changes (remove the ./ and path.resolve) and commit again?

codydaig: @rodp82 It's up to you how you want to go about making those changes. But before it's merged in, we'll ask that it is squashed into 1 commit.

PR #1016, opened 2015-10-23
[docs] Changelog [closes #881]

### LINKED ISSUE #881: CHANGELOG.md
Let's get a CHANGELOG.md added for easy version comparison. Something along the lines of AngularJS's should be good.

codydaig:
@lirantal @ilanbiala @simison 

This is an attempt at a chengelog. All I did was run the tool that @lirantal suggested to get an initial generation to spark more discussion about what we want out of the changelog. 

Generated using changelog:
`changelog github.com/meanjs/mean all -m`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,428 @@
+2015-10-22
+==========
+
+  * Merge pull request [#999](https://github.com/meanjs/mean/issues/999) from mleanos/morgan-logger-config
+    [bug] Enable log options for Morgan that were missed in the 0.4.0 merge
+
+2015-10-19
+==========
+
+  * Enable log options for Morgan
+    Adds the log options, and format to the Morgan middleware in the Express
+    configuration.
+    These options are defined in the environment configurations.
+    The implementation derived from https://github.com/meanjs/mean/pull/254
+    by @lirantal, which somehow got overlooked when merging 0.4.0 into
+    master.
+    Added tests for the Logger configuration.
+    Added the log settings to the Test env config.
+    Added environment variables for the log settings in the Test &
+    Production env configs.
+    Moved the Morgan Express middleware outside of the NODE_ENV ===
+    'development' check. Morgan should be used in all environments, and use
+    the settings set in each env config.
+    Changed the wording of the Stream option comments in the env configs.
+    Added Rotating Logs functionality, and refactored the log Stream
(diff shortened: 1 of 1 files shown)


ilanbiala:
@codydaig seems a little verbose. When I was looking into this, I thought a changelog that had the commits between tags woulld be better, and then even better would be if we start labeling our commits like Angular does so we can parse them into categories already.

codydaig:
@ilanbiala like [bug] [feat] [docs] etc...?

ilanbiala:
Yeah.

lirantal:
@ilanbiala but this is something to begin with. Do we want to add this until we're able to come up with anything better, or would you like to try another setup?

@codydaig good work. I'm not a fan of the dates either. I think it would be best if the changelog can be created per milestones, that would be best IMO. And then we can start with what we have until we enforce any other rules like tags, etc.

ilanbiala:
@lirantal I'd rather not keep changing stuff if we can avoid it by trying something else. @lirantal @codydaig standardizing commit messages would help greatly and we should probably do that at some point to make generating a changelog easier.

ilanbiala:
https://github.com/rafinskipg/git-changelog or https://github.com/ajoslin/conventional-changelog, both look pretty good and format pretty nicely. @codydaig thoughts?

codydaig:
@ilanbiala 
Here's the results from git-changelog https://github.com/codydaig/mean/blob/docs/changelog2/CHANGELOG.md
(I didn't have time to mess with settings, but I think since we don't match the angular commit message guidelines, it won't work how we want.)

codydaig:
@ilanbiala @lirantal 
This is from https://github.com/ajoslin/conventional-changelog
https://github.com/codydaig/mean/blob/docs/changelog3/CHANGELOG.md

ilanbiala:
@lirantal @codydaig I think the changelog3 is better to start with. We'll probably implement a commit structure for 0.5.0 going forward so we can have a better changelog.

If you agree that the last one is best, let's switch the changelog in this file and require that package as a devDep so we have it in the repo.

codydaig:
@ilanbiala I agree. Let me strip it down to just the latest 0.4.1.release and add instructions for generating the next releases to the README and I'll open the new PR for it.

ilanbiala:
@codydaig awesome!

PR #1019, opened 2015-10-27
[hotfix] Logger Development config default

mleanos:
Reverts the default Logger setting to use the stdout by default, rather than the stream option for the Development env config.

I realized this might trip some users up, since when the Logger config was implemented (https://github.com/meanjs/mean/commit/8cd2291a6a8d14cc7a75320944c532ff8b41668f)  I had overlooked the fact that users probably wouldn't want the Stream option turned on by default.

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -19,16 +19,16 @@ module.exports = {
     options: {
       // Stream defaults to process.stdout
       // Uncomment/comment to toggle the logging to a log on the file system
-      stream: {
-        directoryPath: process.cwd(),
-        fileName: 'access.log',
-        rotatingLogs: { // for more info on rotating logs - https://github.com/holidayextras/file-stream-rotator#usage
-          active: false, // activate to use rotating logs 
-          fileName: 'access-%DATE%.log', // if rotating logs are active, this fileName setting will be used
-          frequency: 'daily',
-          verbose: false
-        }
-      }
+      //stream: {
+      //  directoryPath: process.cwd(),
+      //  fileName: 'access.log',
+      //  rotatingLogs: { // for more info on rotating logs - https://github.com/holidayextras/file-stream-rotator#usage
+      //    active: false, // activate to use rotating logs 
+      //    fileName: 'access-%DATE%.log', // if rotating logs are active, this fileName setting will be used
+      //    frequency: 'daily',
+      //    verbose: false
+      //  }
+      //}
     }
   },
   app: {


codydaig:
LGTM

ilanbiala:
@mleanos why would users not want to stream into a file?

codydaig:
@ilanbiala I use PM2 across my servers for node applications, and PM2 automatically takes the stdout and saves it to a rotating file that I can easily tail multiple processes at the same time with, so I personally like stdout being the primary.

ilanbiala:
@mleanos @codydaig so PM2 has support for it, but then we should we include it at all? Why not remove it and just use PM2 instead of Forever in a new PR?

mleanos:
@ilanbiala I think it makes more sense to use stdout for Development, since it was the default being used before the Logger was implemented. I was afraid it would throw some users off. For Production, it makes sense to use the Stream option by default.

ilanbiala:
@mleanos who wouldn't use PM2 or forever in production though? My point is why not just let them handle it completely?

mleanos:
@ilanbiala This doesn't really affect the use of PM2 or forever. Although, PM2 has a way to rotate logs, our current implementation is using Morgan for logging. This change only affects the Morgan express middleware that we're using to log. 

The Development env config will use stdout by default, but the setting can be changed to use the Stream option. As for Production, it seemed appropriate to set Morgan to save the logs to a file by default. Is that what you're questioning?

I don't see how PM2, or forever come into play here. I can see a loose relation to the former, but for our current implementation it's not really relevant.

ilanbiala:
@codydaig said that PM2 can use output from stdout and rotate it through logs, so we can just always have Morgan log through stdout, no?

mleanos:
@ilanbiala We're not implementing PM2 in this project. So the Morgan Stream option provides a way to log to a file. This allows out-of-the-box filesystem logging capabilities for our users.

ilanbiala:
@mleanos Forever seems to have lost much of its support, and @codydaig and @lirantal have advocated for PM2, so I'm suggesting we switch it and avoid us doing more work than we need to by supporting an extra feature when we can use a library that does that and still get better production tools.

mleanos:
@ilanbiala That seems like a great suggestion. Let's get something into the 0.5.0 pipeline for PM2 integration. 

However, I think we can merge this simple change in, to support our current implementation of Morgan.

ilanbiala:
@mleanos have you used PM2 before? @codydaig maybe you could open up a PR just so we have it on record for 0.5.0, even if it's just adding PM2 to the package.json?

mleanos:
@ilanbiala No I haven't used it. But I've been looking at it, and like what I see. 

Thanks for merging!

PR #1023, opened 2015-10-28
Adding bootstrap script for mac users

jloveland:
run `./scripts/boostrap-mac.sh` to install the Mean.js prerequisites on a mac.

### DIFF
--- a/scripts/bootstrap-mac.sh
+++ b/scripts/bootstrap-mac.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+echo "  Installing Mean.js prerequisites."
+echo ""
+
+# Check for Homebrew
+if [ -z `which brew` ];
+then
+  echo "  Installing Homebrew for you."
+    ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)" > /tmp/homebrew-install.log
+fi
+
+# Make sure we’re using the latest Homebrew
+brew update
+
+# Tap versions to install older packages
+brew tap homebrew/versions
+
+# Upgrade any already-installed formulae
+brew upgrade
+echo ""
+
+echo "  Installing Nodejs using NVM."
+brew install nvm
+source $(brew --prefix nvm)/nvm.sh
+nvm install 0.12.7
+echo ""
+echo "*************************************************************************"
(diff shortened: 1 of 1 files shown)


codydaig:
@jloveland I don't like the node aspect since it would install node v4 and we don't support that yet. Also, what happens if you already have some if those components installed?

ilanbiala:
Not necessary in the framework because I imagine most people already have these things installed, we could put it in the gh-pages or somewhere else though.

mleanos:
:+1: On this. I have the same concerns as @codydaig about the Node version & if they already have some of these components, but overall I think this is a useful script.

For me, the purpose of the */scripts* folder is to put whatever we think might be useful for our users. Anything that would make the lives of our users easier, is a good thing. 

Since this isn't integrated into the framework, I don't see any down-sides to adding this.

jloveland:
I was thinking along the lines of if we had some simple bootstrap scripts for mac, windows, and linux, we could quickly help other people debugging issues related to development environments..for instance, I have seen some people (@mleanos) with issues with file paths on windows. I am on a mac. If there was a `scripts/bootstrap-windows.bat` I could run on a vagrant box, I could help those developers on windows test issues.

jloveland:
@codydaig if the users already have these components installed, I would suggest that they run individual commands that they need. However, I have run this script over and over again on my machine and it just warns that the prerequisites are already installed. The output is below:
screen shot 2015-10-28 at 8 38 15 pm

mleanos:
LGTM. Even though I can't test since I'm on Windows :(

lirantal:
I personally feel that it's out of the scope of the project. The `scripts/` directory should be useful for scripts supporiting the framework, and not necessarily for general installation guides. If we make this a precedent then we will later find similar scripts for windows, linux, os/2, freebsd, dos 3.22, well, you get the picture :)

It would be better IMO to put all of that information in:
1. A gist you can always link to
2. Add an installation guide section in the documentation area - that's where people look for help.

ilanbiala:
@lirantal agreed.

mleanos:
I couldn't argue with the points @lirantal made. A gist and/or adding a guide seems like a good approach.

lirantal:
@jloveland looking forward to seeing a PR in the `gh-pages` branch :)

jloveland:
ok, no prob guys! I will add a gist

PR #1025, opened 2015-10-28
[hotfix] SeedDB missing in env config

mleanos:
Adds a check for the existence of the seedDB config setting, before
attempting to read config.seedDB.seed setting.

Solves the problem when the seedDB config setting is missing from a
environment config, that causes the application to throw an exception at
startup.

As pointed out by @ryanjbaxter, due to the `seedDB` setting missing from the Cloud-Foundry env config.

### DIFF
--- a/config/env/cloud-foundry.js
+++ b/config/env/cloud-foundry.js
@@ -66,5 +66,29 @@ module.exports = {
         pass: getCred('mean-mail', 'password') || 'MAILER_PASSWORD'
       }
     }
+  },
+  seedDB: {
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
@mleanos I just tested your branch and it worked for me so this should be good to commit

lirantal:
LGTM

lirantal:
Changing to 0.4.2 so we can merge this defect in quickly.

codydaig:
LGTM

ilanbiala:
LGTM

PR #1026, opened 2015-10-29
Fixing jshint issues with grunt, gulp, and reset-password script.

jloveland:
Fixing jshint issues with grunt, gulp, and reset-password script.


codydaig:
LGTM

PR #1031, opened 2015-10-29
doc: Added access.log to gitignore

codydaig:


lirantal:
LGTM

mleanos:
@codydaig Just so I understand.. The `access.log` file was being generated from the application, and added to the *gh-pages* branch? I guess I don't understand the correlation between what is added to the *gh-pages* from *master*.

codydaig:
@mleanos Any files that arn't checked into git in the master get left behind when you change branches. So when you change branches from `master` to `gh-pages` the newly created `access.log` gets transfered and since we don't want it committed to the `gh-pages` branch either, we have to add it to the `.gitignore` in the `gh-pages` branch. Does that make sense?

mleanos:
Yes, completely. I've never checked out the *gh-pages* branch, so I wasn't thinking of it in terms of switching branches back & forth.

Is there a way to make it easier to manage *.gitignore* between the two branches? It may not be a big deal since there aren't many cases where a file is created in *master* that won't be tracked by Git.

Also, you may want to add `*.log` & `logs/` to this gitignore. Other log files, and the *logs* directory may be generated as well.

codydaig:
@mleanos Does the meanjs app create those anywhere?

mleanos:
@codydaig No, but a user may when running the application locally. Depending on their Logger settings.

If you look here, https://github.com/meanjs/mean/blob/master/.gitignore#L30, you'll see that these settings are present; `*.log` & `logs/` have been in the gitignore for quite some time.

PR #1006, opened 2015-10-19
Added short Gulp Section to Intro

zeldar:
Issue #743 - Adding Gulp.  
The basics have been added, but I will continue to add onto this throughout the week for another commit in the short future.

PR #1037, opened 2015-10-30
#1006 Added Gulp Setup and Troubleshooting to docs/0.4.x

zeldar:
Gulp Setup and Troubleshooting sections added to the 0.4.x docs.

### DIFF
--- a/_includes/docs/0.4.x/gettingstarted.html
+++ b/_includes/docs/0.4.x/gettingstarted.html
@@ -30,6 +30,11 @@
       You're going to use the <a target="_blank" href="http://gruntjs.com/">Grunt Task Runner</a> to automate your development process, in order to install it make sure you've installed Node.js and npm, then install grunt globally using npm:
       <pre>$ npm install -g grunt-cli</pre>
     </dd>
+    <dt>Gulp (Optional)</dt>
+    <dd>
+      You may use the <a target="_blank" href="http://gulpjs.com/">Gulp Task Runner</a> instead of Grunt.  You may use Gulp for Live Reload, Linting, and SASS or LESS compilation.
+      <pre>$ npm install -g gulp</pre>
+    </dd>
   </dl>
   <p class="alert alert-info">
     <b>Note:</b> Your user might not have the permissions to install package globally, so use a super user or <b>sudo</b>.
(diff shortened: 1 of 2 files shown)


_includes/docs/0.4.x/cloud-foundry.html
```diff
@@ -40,6 +40,11 @@
         <li><pre>$ grunt build</pre></li>
       </ul>
     </li>
+    <li>Run the Gulp Build task to build the optimized JavaScript and CSS files (If you chose Gulp instead of Grunt)
```
codydaig: I'd prefer to leave this out of the deployment sections until we switch over to gulp completely. 

@lirantal Thoughts?

_includes/docs/0.4.x/gettingstarted.html
```diff
@@ -30,6 +30,11 @@
       You're going to use the <a target="_blank" href="http://gruntjs.com/">Grunt Task Runner</a> to automate your development process, in order to install it make sure you've installed Node.js and npm, then install grunt globally using npm:
       <pre>$ npm install -g grunt-cli</pre>
     </dd>
+    <dt>Gulp (Optional)</dt>
+    <dd>
+      You may use the <a target="_blank" href="http://gulpjs.com/">Gulp Task Runner</a> instead of Grunt.  You may use Gulp for Live Reload, Linting, and SASS or LESS compilation. Node must be installed first.
```
codydaig: Do we really need to specify that Node must be installed first?

zeldar: @codydaig I think it's pretty obvious that you need npm.  Should this be removed for both Grunt and Gulp?

codydaig: @zeldar I would. It just seems redundant too me since right above that we tell them that they need node and npm.

codydaig:
@zeldar Besides the couple line comments above, looks good. 

THANKS A BUNCH! :-D

_includes/docs/0.4.x/cloud-foundry.html
ilanbiala: @codydaig I'm actually going to try to put some work into Gulp this weekend, so let's hold off on removing it if I can make reach parity between Gulp and Grunt.

codydaig: @ilanbiala Here it should be one or the other. Since they are deploying to a service, it should be what we think is the most reliable. Currently that is grunt. Now, I'd love to eliminate grunt in 0.5 and have everything run with gulp, and in that case we can switch the documentation over to reflect the new gulp set-up. Having both side by side in a deployment walk through is slightly confusing.

ilanbiala: True that we can't remove Grunt until 0.5, so let's go with your suggestion.

zeldar: @codydaig I'm removing this line and sending a new commit now. Will wait for 0.5 to add.

zeldar:
@codydaig Did that squash work?

codydaig:
@zeldar Yes  the squash worked. Thanks!

Issue #389, opened 2015-02-03
MEAN.js 0.4.0 with token authentication?


ricardodovalle:
I would like to use MEAN.js with token authentication, but I am a  Node.js beginner and I am porting a Rails app to learn.

I created a small project and add some features, the `'api/signin'` route is returning the token (`loginToken`) encoded with json web token (`jwt`), but I need help.

How to connect and keep Angular session a live using token?
How to add `express-jwt` and to protect all `/api` route keeping the project organized?
What is the correct way to store the secret? `(users.authentication.server.controller.js and strategies/local-token.js)`?
What is the correct way to user `signout`?

*Working on: https://github.com/ricardodovalle/meanjs-token-auth*

Thank you very much.

sylvainlap:
Use jwt-simple to create your JSON Web Token.
Once your client get a token, you can save it client-side, thanks to the LocalStorage. And then, at each requests to the API, the client must send the token.
On the backend side, the token has to be check at each request to the API, thanks to an Express middleware.

You can read more about the workflow here : http://www.sitepoint.com/using-json-web-tokens-node-js/

Good luck ;)
Sylvain

ricardodovalle:
Hi @sylvainlap,
I used jsonwebtoken instead jwt-simple, because it looks like more updated.
The project is creating the token and sending inside a json field (https://github.com/ricardodovalle/meanjs-token-auth)

I don't know how to do "...the token has to be check at each request to the API, thanks to an Express middleware."

Thanks

sylvainlap:
Are you using Passport.js ? If yes, you should use this Strategy : https://github.com/jaredhanson/passport-http-bearer

"passport.authenticate" will be the Express Middleware that will check for the token.

ricardodovalle:
Thanks @sylvainlap :clap: 
I will work and try to use.

acoulon99:
Working on updating our repo for a meanjs token auth example. Hopefully it'll be useful to the community :)
https://github.com/castlewhitehall/meanjs-with-token-auth

denzelwamburu:
I found this great Token-based AngularJS Authentication https://github.com/sahat/satellizer and works with any app perfectly.I prefer integrating this with the boilerplate but the issue is i can't seem to get the auth files in Meanjs.Any way to get around this?

acoulon99:
I would be interested in that as well

sylvainlap:
This lib looks great, but redondant with Passport.
[...]

PR #1040, opened 2015-10-30
Add first cut at token based authentication

kokokenada:
Re #389, I went through the repository from  https://github.com/vko-online/meanjs-token-auth from @vko-online and migrated the changes. It also addresses a couple of other issues that popped up with Ionic (changing the REST end point URL and CORS).  The tests don't pass yet and I have not fixed up the chat sample.


mleanos:
@kokokenada Thank you for working on this. Can you explain what you have working, and any struggles you've had with this thus far? There seems to be a lot going on.

I have a couple initial questions..

1) Is it necessary to handle the base REST endpoint in this implementation? I feel it might be complicating the issue.

2) There are drastic changes in the Socket.IO implementation. What prompted that?

3) It looks like you added the `userId` param to the User's service, so that it would go through the middleware. Is that correct?

I actually started working on a new branch for Token Auth this morning. I guess we both saw the urgency to get this done :) I'll report back when I push my branch to my Fork; we can compare & possibly combine some of the implementation. Mine differs quite a bit in most ways, but is similar in a couple as well. I ended up going with a the *jsonwebtoken* library, rather than *jwt-simple*.

I think with our combined efforts, and interest in this, we can get this done for the 0.5.0 release. Thanks, again!

codydaig:
@kokokenada Thanks for the work! 
However, I would much prefer you break this PR into multiple PRs. Tokens should be one, CORS should be another, and Socket.io changes should be another. There is too many different changes here for one PR.

kokokenada:
Thanks for the comments.  It's great that other people are working on this.  It was a blocker for my project, so I had to get something working. 

The code as is allows me to login, logout, change profile, change password, seed, etc.  Tests fail. I am able to use an Ionic client (which was impossible until now).

Re the REST endpoint, I found it necessary to achieve my over all goal of using an Ionic mobile client that also reuses some of the services and controllers from the web client. 

BTW - Is there a better way to get the client to read settings from the config directory?  I'm planning on using CORE_CONST to also configure accessing other cloud services from the client and found no existing mechanism (maybe I missed it).  It seems like an important configuration capability.

Re the sockets approach, this was what was used by @vko-online in the companion repository he developed (https://github.com/vko-online/Men-ionic) I had some sort of integration issues (can't remember what). I'm not really in familiar territory here...

Re the UserId param, if I understand the question, it has always worked that way. I made no changes.

BTW - a couple of issues the the code as is.  Doing a browser refresh leads to the login page.  I think this is a problem with the order that things happen.  Also, when you login into the mobile client, it kills the authentication for the web session and vice versa.  Maybe we we'll need some sort of client ID to key into a sub document that stores the salts for each client. (???)

What do you guys suggest I do for next steps? (As you might have guessed I'm new to making contributions - and, like everyone else. have limited time).

ilanbiala:
@kokokenada minimum if all these changes are really necessary is to put them in different commits. Even if they are all necessary for token auth, they aren't really related/in the same scope, so client-side stuff should at least be a separate commit. That said, anything that can be removed should absolutely be removed.

vko-online:
Changes looks great, we must hop on that asap

mleanos:
Here's the branch that I started yesterday. The implementation differs greatly in most ways.
https://github.com/mleanos/mean/commit/52c7e59f7463624d9e00816f67ee003d8ad77a95

I may be doing some things wrong, and not taking certain aspects of this into consideration. However, this is still a work in progress. 

What I want to demonstrate with this branch, is a way to go about implementing token-auth with the least amount of change to the core framework, and existing modules. A more seamless integration.

There are still some issues I haven't addressed, but it sounds like I'm at the same stage as this PR. For instance, on page refresh the user's authentication is lost too. I'm going to try to address that today. 

I'm hoping that we can combine efforts to get this done for 0.5.0. I think we can with this new energy behind it.

mleanos:
I was able to fix the issue with the page refresh. https://github.com/mleanos/mean/commit/d1f4bb2a1b26a0e34745d6050dcd424cea43da1e

modules/core/client/app/init.js
```diff
@@ -12,7 +12,23 @@ angular.module(ApplicationConfiguration.applicationModuleName).config(['$locatio
   }
 ]);
 
-angular.module(ApplicationConfiguration.applicationModuleName).run(function ($rootScope, $state, Authentication) {
+angular.module(ApplicationConfiguration.applicationModuleName).run(function ($rootScope, $state, $http, Authentication, CORE_CONST) {
+
+  //set token header
+  var auth_token = localStorage.getItem('auth_token');
```
jloveland: @kokokenada, @mleanos, I was looking into how we should store the token and Stormpath recommends to put it in cookies rather than localStorage:
https://stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage/
>Stormpath recommends that you store your JWT in cookies for web applications, because of the additional security they provide, and the simplicity of protecting against CSRF with modern web frameworks. HTML5 Web Storage is vulnerable to XSS, has a larger attack surface area, and can impact all application users on a successful attack.

mleanos: @jloveland Thank you for bringing up that point. It's definitely something that we should discuss.

I did read, however, that since the token's are sent with the Authorization header it's not vulnerable to CSRF because CSRF doesn't include the header. And I may be misinterpreting this statement from this article; I loosely based my implementation off this btw 
http://code.tutsplus.com/tutorials/token-based-authentication-with-angularjs-nodejs--cms-22543
>No Cookie-Session (or No CSRF). CSRF is a major problem in modern web security because it doesn't check whether a request source is trusted or not. To solve this problem, a token pool is used for sending that token on every form post. In token-based authentication, a token is used in authorization headers, and CSRF does not include that information.

mleanos:
This PR has a lot of changes, and I think some of them may be unnecessary. I agree with @ilanbiala, that if we find these changes are indeed required then we should break it up into smaller commits.

With that said, I propose that start with a fresh PR based off of my branch since it has less changes, and I've started to break it up into small-ish commits.

We can collaborate on this through my branch, and we'll be able to delegate the contributions. For instance, I may be doing some of the client-side fundamentally wrong. I'm more of a server-side developer, and I think someone else might be better suited for the client-side implementation.

Here are my two commits thus far
https://github.com/mleanos/mean/commit/7e853df8ef92757c9f08b65f9ffd907f290a8c26
https://github.com/mleanos/mean/commit/d1f4bb2a1b26a0e34745d6050dcd424cea43da1e

What I have working:
Server-Side
1) The user can login and a Token will be generated for them
2) The user's data is updated with the generated token, and an expiration
3) Default env config stores the tokenAuth setting for Secret & expiresInSeconds
4) Added an express middleware that handles the authorization of each request. If the Authorization header is present, then it attempts to validate it. If not, then the request continues to the route, which will either allow/deny based on whether or not the route requires authentication.

Client-Side
1) User can login, and the token returned from the server and be stored in Local Storage
2) The Authentication service has been updated to retrieve & update the Token stored in Local Storage
3) Added a Request interceptor that will add the Token to the each request to the server, if the Local Storage has a Token.
4) Page refreshes will persist the login
5) The logout will remove the Token from the Local Storage

TODO:
1) When the user updates their profile (image, or profile details) the Token is not updated. This causes an issue when the page is refreshed. The user's data is reverted back to what is stored in the Token. This may be a flaw in how I'm handling this on the client-side.
2) Socket.IO updated to get the User info from the Authorization header.

I'm sure there are other things I'm missing, or haven't thought of. I'm happy that we have this moving now, and I'm confident we can get this out in 0.5.0. It will need a collaborative effort :)

vko-online:
> Added a Request interceptor that will add the Token to the each request to the server, if the Local Storage has a Token.

That will cause more overheat, token must be injected into $http service in the beginning of angular bootstrap, so that it will persist anyway
```
        $http.defaults.headers.common.Authentication = new_user.loginToken;
```
--
>When the user updates their profile (image, or profile details) the Token is not updated. This causes an issue when the page is refreshed. The user's data is reverted back to what is stored in the Token. This may be a flaw in how I'm handling this on the client-side.

Token is used only to validate user, when updating user-info, it should be temporary deleted, just as roles
```
        var user = req.user;
        var update_user = req.body;
        delete update_user.roles;
        delete update_user.loginToken;
        delete update_user.loginExpiress
        var new_user = _.extend(user, updated_user);
        new_user.save();
        res.send(new_user); //this has updated picture, but same role, token
```

modules/core/client/app/init.js
kokokenada: Session cookies cause a lot of problems for Cordova clients, if it is possible at all.  For me, this is the main reason for token based auth.

 https://cordovablogsblogs.wordpress.com/2015/03/08/handling-cookies-in-phonegapcordova/

kokokenada:
On my local implementation (which is still the original PR), I've noted that when I run the app in development mode, register a user, run in production mode, that UID/PWD combo fails.  The reverse is true as well (UID/PWD's created in production fail in dev).  I can't see what accounts for that behavior. Does the new branch have this behavior?

modules/core/client/app/init.js
codydaig: I've personally stored tokens in local storage for all my projects. That's mainly where I've seen it anywhere else as well.

jloveland: I see conflicting information out on a lot of blogs. Maybe this is due to wanting to deploy iin different environments (web app vs. cordova/phonegap/ionic app)... 

There are some references to not caching sensitive data in Phonegap/Cordova here:
http://docs.phonegap.com/en/3.5.0/guide_appdev_security_index.md.html

It seems to me that *encrypting* the token would be the best practice if we cache it in `localStorage`
> Encrypt the refresh token and store it encrypted in the localstorage. You can use CryptoJS (a JS library to encrypt/decrypt the data) to encrypt your token using AES (see this example) and it also has an angularjs module

http://security.stackexchange.com/questions/71878/cordova-phonegap-refreshtoken-in-localstorage

Stormpath uses cookies for storing the token and their existence depends on security best practices. However, I am not sure how that blog is considering deploying a Cordova client. Again, I will ask them about this.

jloveland:
I think it would be easier to setup multiple, smaller PRs for this so we can better review and offer support. WDYT?

mleanos:
@vko-online Can you explain what you see as the differences in the overhead of using an angular Request interceptor vs injecting into the $http service?

We definitely need to send the Token along with every request to the server, so I'm just trying to understand the differences. I actually prefer to use the request interceptor. The existing authInterceptor service seems like an appropriate place to handle this; but I first want to understand the implications of doing so.

vko-online:
@mleanos Consider 10 requests containing 3 script files, 3 images, 3 html files and 1 articles.list, when angular is already bootstraped, interceptor will read from localStorage, apply to $http, and send, for all 10 requests. But if you apply it in the beginning, it will still send auth token along with 10 requests, but will not read from localStorage and apply. It's just set in the beginning. So 1 set operation better that 10 read & set i believe

kokokenada:
@mleanos - did you look into any issues with oAuth/passport and tokens? I think the oauth implementation relies on sessions.  Should we generate tokens after a successful oauth? Should we do away with sessions? One issue is that enabling CORS broke the oauth - which I think is related to to its use of sessions.

kokokenada:
@mleanos - did you look into any issues with oAuth/passport and tokens? I think the oauth implementation relies on sessions. Should we generate tokens after a successful oauth? Should we do away with sessions? One issue is that enabling CORS broke the oauth - which I think is related to to its use of sessions.

PR #1052, opened 2015-11-09
Update auth.interceptor.client.service.js

0xNacho:
Added 400 and 404 error routes


mleanos:
@0xNacho What happened to this PR? It looks like it will solve #1504

PR #1057, opened 2015-11-12
bug(docs): Update change log command

codydaig:


ilanbiala:
@codydaig does `-r 1` only run the last version and add it to the CHANGELOG.md? Ideally that's what we want, so if we tidy up 0.4.0 and 0.4.1 a little, then we don't lose that every time.

codydaig:
@ilanbiala Exactly

ilanbiala:
LGTM if it works.

CONTRIBUTING.md
```diff
@@ -82,4 +82,4 @@ The code for the documentation and the website are located in the meanjs/mean re
 
 For those who will be generating releases:
 * Don't forget to update the changelog!
-  * `conventional-changelog -r 1`
+  * `conventional-changelog -r 1 | cat > CHANGELOG.md`
```
lirantal: If -r 1 is only adding latest version information then we might want to append to the CHANGELOG's top of the file, not overwrite it.

@ilanbiala I assume we want CHANGELOG to have all milestones and not just the latest?

ilanbiala: Yeah, but prepend right, not append?

lirantal: Yep.

lirantal: @codydaig since that command appends to the file it will cause later versions to appear at the bottom of the file, right?

If so, then you need to figure out how to prepend to the file, something like one of these: http://www.commandlinefu.com/commands/view/14354/prepend-text-to-a-file

lirantal: @codydaig can you update this PR to change the command there to the following for actually prepending to the file:
```bash
conventional-changelog -r 1 | cat - CHANGELOG.md
```

codydaig:
@lirantal Updated.

lirantal:
@codydaig sorry for not being clear. That command will simply prepend the text but will output it to STDOUT (the screen), and not the file. If you intend for the file to be overwritten we should use the `tee` command as follows:
```bash
conventional-changelog -r 1 | cat - CHANGELOG.md | tee CHANGELOG.md
```

codydaig:
@lirantal Sorry. I'm not in an environment to easily test that today. I've made your requested change and can verify it later tonight.

lirantal:
Ok so I think we're good here.
@codydaig go ahead and merge it so we can continue with 0.4.2

Issue #511, opened 2015-04-08
The provider data are not refreshed when the user log again (so no new accesstoken)


junit38:

ilanbiala:
Please highlight your code using GFM.

codydaig:
@junit38 Are you still having issues?

junit38:
Yes, the user provider's data are no refreshed on each signin.
The lines to change is now in the file: 
modules/users/server/controllers/users/users.authentication.server.controller.js

at line 175-177:

```
} else {
          return done(err, user);
}
```

ilanbiala:
@lirantal @junit38 any idea whether this is still an issue?

lirantal:
I didn't eventually get to review that one.
If anyone has any social login configured we can easily test by changing some profile detail after being logged-in once and then logout/login to see if that new value is updated on meanjs.

junit38:
Yes there is still and issue on the last version of the framework.
Just sign in using facebook strategy and then logout/login to see that the data did'nt not change.
As @lirantal suggested.

mleanos:
@junit38 Can you explain what you're experiencing a little more clearly?

How can I reproduce? What data is not getting updated?

This is how I've tried to reproduce, and couldn't:

1) Created a new local provider account
2) Added Facebook as a provider
3) Logged Out
4) Logged back in using the local provider user/pass method
5) Facebook provider was still linked to the account

OR

1) Created a new local provider account
2) Added Facebook as a provider
3) Updated Last Name
4) Logged Out
5) Logged in again, and Last Name reflected my change 

**Repeated above step, but logging in again as Facebook provider; same result**

OR

1) Created a new account using Facebook provider
2) Updated Last Name
3) Logged out
4) Logged back in, and Last Name reflected my update

junit38:
1) Create a Facebook account using your Facebook signup.
2) See the accessToken
3) Logout
3) Login again

The accesToken is not refreshed. (You can try to change your Facebook name to see but this is a trap because you can change it only every 60 days....)

This is all the ProviderData which are not refresh.

So you can't use the Facebook accessToken to do action on the user Facebook account by example (the access token expiration limit is above 2months but it's still an issue)

mleanos:
@junit38 Ok, so if you change some data on your social media (Facebook in this case) account, then your meanjs account isn't updated the next tine you log in with that provider?

[...]

PR #1062, opened 2015-11-16
Refresh user providerData when the user log again #511

junit38:
The PR is refreshing the user providerData when the user is loggin again with a social account to maintain the accuracy of the data.

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -173,7 +173,26 @@ exports.saveOAuthUserProfile = function (req, providerUserProfile, done) {
             });
           });
         } else {
-          return done(err, user);
+          // set the new data from the provider
+          if (user.providerData) {
+              user.providerData = providerUserProfile.providerData;
+              // Then tell mongoose that we've updated the providerData field
+              user.markModified('providerData');
+               // And save the user
+              user.save(function(err) {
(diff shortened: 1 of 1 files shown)


codydaig:
@junit38 Thanks for the PR! Couple things:
- Please read the contributing doc about the git commit message guidelines: https://github.com/meanjs/mean/blob/master/CONTRIBUTING.md#commit-message-guidelines
- Please update the description to explain what this PR is accomplishing. 
- Please fix the eslint issues outlined in the failed build
- Please ensure that all tests pass. 

Ping me once these are addressed so we can review this PR. Thanks! :-D

junit38:
Ok I fixed the issue but i badly spelled the commit. Is that a problem ?

mleanos:
@junit38 What are your thoughts on just updating the user's Token? 

My idea is that we should treat whatever is in their MEANJS profile, as their *master* profile data. Meaning their data shouldn't be automatically synced with their Provider data. 

In a separate feature PR, we could add a "Sync" button for pulling over their Provider data into their MEANJS profile. But that would require some discussion, and something we can try to add to 0.5.0 as a bonus feature.

rhutchison:
Duplicates #1070 - does this need to remain open?

codydaig:
@junit38 I'm closing this for now. It does seem like a duplicate. Feel free to reopen if it needs to be open. 

@rhutchison Thanks!

PR #1083, opened 2015-12-01
feat(config): Gulp Server Watch for Mocha

mleanos:
Adds a Gulp task that watches all server files (including server tests),
and upon any changes it will perform the Gulp *test:server* task.

Also, includes an **optional** argument for the task that can specify
that if the changed file is a test, then it will only run that test
file.

Example usage: gulp test:server:watch --onlyChanged

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,6 +6,7 @@
 var _ = require('lodash'),
   defaultAssets = require('./config/assets/default'),
   testAssets = require('./config/assets/test'),
+  glob = require('glob'),
   gulp = require('gulp'),
   gulpLoadPlugins = require('gulp-load-plugins'),
   runSequence = require('run-sequence'),
@@ -16,6 +17,7 @@ var _ = require('lodash'),
   }),
   path = require('path'),
   endOfLine = require('os').EOL,
+  argv = require('yargs').argv,
   protractor = require('gulp-protractor').protractor,
   webdriver_update = require('gulp-protractor').webdriver_update,
   webdriver_standalone = require('gulp-protractor').webdriver_standalone,
@@ -65,6 +67,35 @@ gulp.task('watch', function () {
   } else {
     gulp.watch(defaultAssets.server.gulpConfig, ['jshint']);
     gulp.watch(defaultAssets.client.views).on('change', plugins.livereload.changed);
+  }
+
+  if (process.env.NODE_ENV === 'test') {
+    // Add Server Test file rules
+    gulp.watch([testAssets.tests.server, defaultAssets.server.allJS], ['test:server']).on('change', function (file) {
+      var runOnlyChangedTestFile = argv.onlyChanged ? true : false;
+
+      // check if we should only run a changed test file
+      if (runOnlyChangedTestFile) {
+        var changedTestFiles = [];
+
+        // iterate through server test glob patterns
+        _.forEach(testAssets.tests.server, function (pattern) {
+          // determine if the changed (watched) file is a server test
+          _.forEach(glob.sync(pattern), function (f) {
+            var filePath = path.resolve(f);
+
+            if (filePath === path.resolve(file.path)) {
+              changedTestFiles.push(f);
+            }
+          });
+        });
+
+        // set task argument for tracking changed test files
+        argv.changedTestFiles = changedTestFiles;
+      }
+
+      plugins.livereload.changed();
+    });
   }
 });
 
@@ -181,13 +212,14 @@ gulp.task('templatecache', function () {
 gulp.task('mocha', function (done) {
   // Open mongoose connections
   var mongoose = require('./config/lib/mongoose.js');
+  var testSuites = Array.isArray(argv.changedTestFiles) && argv.changedTestFiles.length ? argv.changedTestFiles : testAssets.tests.server;
   var error;
 
   // Connect mongoose
   mongoose.connect(function () {
     mongoose.loadModels();
     // Run the tests
-    gulp.src(testAssets.tests.server)
+    gulp.src(testSuites)
       .pipe(plugins.mocha({
         reporter: 'spec',
         timeout: 10000
@@ -273,6 +305,14 @@ gulp.task('test', function (done) {
 
 gulp.task('test:server', function (done) {
   runSequence('env:test', 'lint', 'mocha', done);
+});
+
+// Watch all server files for changes & run server tests (test:server) task on changes
+// optional arguments: 
+//    --onlyChanged - optional argument for specifying that only the tests in a changed Server Test file will be run
+// example usage: gulp test:server:watch --onlyChanged
+gulp.task('test:server:watch', function (done) {
+  runSequence('test:server', 'watch', done);
 });
 
 gulp.task('test:client', function (done) {
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
     "mock-fs": "~3.4.0",
     "run-sequence": "^1.1.1",
     "should": "^7.0.1",
-    "supertest": "^1.0.1"
+    "supertest": "^1.0.1",
+    "yargs": "~3.30.0"
   }
 }


mleanos:
This is a pretty handy feature that I've been using with a project based off meanjs. It really speeds development & encourages proper TDD considerations.

**Normal test execution**:
gh-1083_1

**When the --changedOnly option is set & the changed file is a test**:
gh-1083_2

**Another**:
gh-1083_3

lirantal:
Sounds like a great addition. @ilanbiala is the gulp guy.

mleanos:
@ilanbiala There were a few things here that I was unsure of; quick review?

In particular, I don't have much experience with Glob. I didn't see a real straight forward way of getting an individual test file from the `testAssets.server.tests` array of glob patterns. So this is the best way I saw fit. I'll point out a couple things in line notes.

gulpfile.js
```diff
@@ -68,6 +70,39 @@ gulp.task('watch', function () {
...
+  plugins.livereload.listen();
+
+  // Add Server Test file rules
+  gulp.watch([testAssets.tests.server, defaultAssets.server.allJS], ['test:server']).on('change', function (file) {
+    // check if we should only run a changed test file
+    if (runOnlyChangedTestFile) {
+      var changedTestFiles = [];
+
+      // iterate through server test glob patterns
+      _.forEach(testAssets.tests.server, function (tests) {
```
mleanos: @ilanbiala What do you think of this section overall?

Here, I'm iterating through the glob patterns because, since it's an array, technically a project could have multiple locations/patterns for tests.

gulpfile.js
```diff
@@ -68,6 +70,39 @@ gulp.task('watch', function () {
...
+        // determine if the changed (watched) file is a server test
+        _.forEach(glob.sync(tests), function (f) {
```
mleanos: @ilanbiala Is there any other, more straightforward, way of checking if the changed file from the *watch* is part of `assets.tests.server` glob pattern array? Ideally, I'd just like to check if `file` matches any of the `assets.tests.server` patterns; I just couldn't see a way to do that. My lack of experience with Glob showing here.

ilanbiala: Not really sure if this is a glob thing, but why not just check if the intersection of the two lists (watched files and assets.tests.server) is nonzero?

ilanbiala:
@mleanos this shouldn't all be one gulp task in my opinion. Ideally gulp tasks are composed, just like sass, less, watch, etc. which all do one thing and are used in other places to make general utility tasks.

I think it would be better to specify this functionality in the general watch task and just pare down the functionality a bit. I see the potential time value of just running the affected file tests, but when other things depend on others, you may still have issues, so I would just run all the tests anyway.

mleanos:
@ilanbiala Have you taken this for a test drive? I suggest that you do, so you understand the benefits a bit more.

As for it being included in the other Watch task, this is a separate task that runs without any other dependencies; it's not linked to any other Gulp task.

The way that I use this, is to have it running in a terminal as I'm developing.. anytime I save a server file, the task invokes the existing `tests:server` task. If everything passes, I continue developing. With this flow, you can see that it should be an isolated task on it's own; there's no need to couple it with the server task, for instance.

mleanos:
@ilanbiala After looking at the `watch` task again, I see what you mean. I think this new *watch* configuration can be added there. We would still use the new `test:server:watch` task, in order to run the server tests initially, and set the NODE_ENV to *test*

```javascript
gulp.task('test:server:watch', function (done) {
  runSequence('test:server', 'watch', done);
});
```

This is what it would look like added to the existing `watch` task
```javascript
if (process.env.NODE_ENV === 'test') {
    // Add Server Test file rules
    gulp.watch([testAssets.tests.server, defaultAssets.server.allJS], ['test:server']).on('change', function (file) {
      var runOnlyChangedTestFile = argv.onlyChanged ? true : false;

      // check if we should only run a changed test file
      if (runOnlyChangedTestFile) {
        var changedTestFiles = [];

        // iterate through server test glob patterns
        _.forEach(testAssets.tests.server, function (tests) {
          // determine if the changed (watched) file is a server test
          _.forEach(glob.sync(tests), function (f) {
            var filePath = path.resolve(f);

            if (filePath === path.resolve(file.path)) {
              changedTestFiles.push(f);
            }
          });
        });

        // set task argument for tracking changed test files
        argv.changedTestFiles = changedTestFiles;
      }

      plugins.livereload.changed();
    });
  }
```

gulpfile.js
mleanos: We need to check if the currently changed file, that invoked the `watch.on('change')` event is part of the `assets.tests.server`; not so straight forward because the latter is an array of glob patterns.

ilanbiala:
Still making the watch a little too powerful in my opinion. Also, I don't think adding argv parsing to gulp is a great idea because it seems unnatural in gulp. Is there a better way you can think of? What would creating a separate task for watch changed and not watch changed look like?

mleanos:
@ilanbiala I don't understand what you mean by "too powerful".. What's the point of Gulp, if not to help manage mundane development tasks? Think of the development workflow that you have to follow to save a file, and then run the tests, and keep repeating that process.

And what is unnatural about adding `argv` parsing to a Gulp task? I'd like to understand what you mean, but I'm just not following.

ilanbiala:
Instead of argv parsing for different tasks, Gulp seems to prefer just creating a different task with a different name.

Tasks that are alone too powerful do too much in the function and don't compose the functionality from other tasks that should do one specific task well.

mleanos:
@ilanbiala I removed the new Gulp task, and moved the watch into the main Gulp watch task.

The *optional* argument that I'm using `argv` to track, is not specifying that a different task should be run. It's determining whether or not *all* server tests should be run, or just the modified file that triggered the `watch.on('change')` event. In other words, the argument is only affecting the tests to be ran with Mocha here https://github.com/meanjs/mean/pull/1083/files#diff-b9e12334e9eafd8341a6107dd98510c9R222

Having a separate task that runs just one server test file, and then having one that runs all server tests, would be duplicate code. I think the use of `argv` is the clean way of handling a requirement such as this.

As for your concern over the `watch` definition being "too powerful", I think the purpose of the `watch` callback is for tasks that need special handling. I believe this is such a case.

How would you approach this differently, given these requirements?
1) Start a Gulp task that initially runs all Server Tests
2) Now watch for any changed Server files
3) Upon the above watch being triggered, run all Server Tests; with an option that the user specifies to just run a single Server Test, if the changed file that triggers the `watch.on('change')` event is a Server Test file.

I'm open to suggestions. However, this feels like the most straight-forward, and clean, way of satisfying the above requirements.

ilanbiala:
@lirantal @codydaig any suggestions/feedback?

lirantal:
nothing special from me

codydaig:
@ilanbiala I haven't followed this too closely. But the concept of being able to watch test files and either run a single test from a passed in arg or an entire suite of tests is awesome.

mleanos:
@ilanbiala Any more feedback on this? 

I've been using this implementation for development, and it's immensely useful. It makes TDD so much faster. I think others will see it as a very useful tool for productivity.

ilanbiala:
I have finals until next week, so I can't really spend any time on other stuff until then. Ping me then.

mleanos:
SGTM. Thanks.

mleanos:
@ilanbiala Let's get this merged in, so we can keep going on enhancing the Gulp configuration.

package.json
```diff
@@ -116,6 +116,7 @@
     "mock-fs": "~3.4.0",
     "run-sequence": "^1.1.1",
     "should": "^7.0.1",
-    "supertest": "^1.0.1"
+    "supertest": "^1.0.1",
```

package.json
```diff
@@ -116,6 +116,7 @@
...
+    "yargs": "^3.30.0"
```

gulpfile.js
```diff
@@ -66,6 +68,35 @@ gulp.task('watch', function () {
...
+    // Add Server Test file rules
+    gulp.watch([testAssets.tests.server, defaultAssets.server.allJS], ['test:server']).on('change', function (file) {
+      var runOnlyChangedTestFile = argv.onlyChanged ? true : false;
+
+      // check if we should only run a changed test file
+      if (runOnlyChangedTestFile) {
+        var changedTestFiles = [];
+
+        // iterate through server test glob patterns
+        _.forEach(testAssets.tests.server, function (pattern) {
```
ilanbiala: why _ and not native Array.forEach?

ilanbiala:
Last couple fixes, and then I'll merge.

gulpfile.js
mleanos: I've found that lodash is more efficient with their looping, and we're already using lodash in the Gulp configuration.

mleanos:
@ilanbiala I made the change to the `yargs` dependency.

I left the use of lodash's `.forEach`, as it seems to be more favorable & efficient than the native `Array.forEach`.

package.json
ilanbiala: never mind, just comma change, so this is fine

ilanbiala:
LGTM. @lirantal @codydaig any comments before I merge?

mleanos:
@ilanbiala I noticed, after the fact, that you also pointed out the dependency of `supertest`. The diff makes it look like I added it, but it was already there. If we want to make a sweeping change to use `~` on all dev dependencies, then that can be a dedicated PR.

ilanbiala:
@mleanos yep, I commented on that before, so it's fine.

codydaig:
@mleanos This needs a blurb in the documentation, but that's a separate PR to a separate repo. 

LGTM

mleanos:
@ilanbiala Can we get this merged?

ilanbiala:
Sorry about the delay, just merged it in.

mleanos:
No worries. I jhst have some other work I wanna do with Gulp. 

Thanks!

Issue #1072, opened 2015-11-27
Using both ESLint and JSHint


feimosi:
Is it a good idea to keep both: ESLint and JSHint in the project? It'll print many duplicate linting errors to the console. Also ESLint haven't been attached to the gulp watchers. And lastly, as an enhancement, there's a great eslint-plugin-angular for linting Angular apps.
Referencing #990

lirantal:
@feimosi do you want to submit a PR to clean it up and add the angular eslint plugin?

feimosi:
@lirantal sure, no problem. Should get rid of JSHint completely?

lirantal:
If jshint is redundant then yes obviously remove it in favor of eslint.

codydaig:
:+1:

ilanbiala:
I think the specific duplicates are just some rules we have in both places. I'm not sure if JSHint is has all the features JSHint has and more though, so that will need to be checked if we want to remove JSHint.

feimosi:
That's the point, ESLint offers the same functionality and even more, so there's no point in using two linting tools. AFAIK it's closely compatible with JSHint, so most of the rules should be there. We just need to do the migration from `.jshintrc` to `.eslintrc`. Preferably also add new rules, including code style, but that may need an additional discussion. 
Also I think it wouldn't be a bad idea to use the 'extending config' feature from ESLint and use someone else's config as the base e.g. well-known Airbnb rules.
What do you say?

ilanbiala:
@feimosi If it doesn't throw errors, that sounds good to me. Otherwise, we'll need to just discuss any changes extending an existing config.

lirantal:
@feimosi we're waiting on that PR from you... :-)

Issue #943, opened 2015-09-29
Coding Styles and Code Formatting


lirantal:
# Topic
Coding Styles and Code Formatting across the MEAN.JS project code base.

# Conventions
* Named functions (no anonymous functions allowed)

# Tool
We chose ESLint as the upcoming tool to perform the job, if you would like to propose anything else please comment with the pros/cons (see reference for PR request: https://github.com/meanjs/mean/issues/763)

lirantal:
@ilanbiala @mleanos @codydaig @trainerbill @rhutchison we've got this place to discuss items now related to coding conventions. You're welcome to raise all the formatting and code style that were raised in other PRs here and we'll add them to the main issue description as this grows.

codydaig:
No nested functions! It should be a closure instead: ie. 

Incorrect:
```
function foo() {
  var user = 'cody';

  someFunctionReturningAPromise()
    .then(bar);

  function bar() {
    // uses the user variable
  }
}
```

Correct:
```
function foo() {
  var user = 'cody';

  someFunctionReturningAPromise()
    .then(bar(user));
}

function bar(user) {
  return function () {
    // Yay Closures!!
  }
}
```

ilanbiala:
It should actually just be `.then(bar)` to keep it simplest.

codydaig:
@ilanbiala so you prefer nested functions?

trainerbill:
This is bound to ruffle some feathers, but for the client side / angular, can we just put conform to john papa style guide?  Or are we creating a whole new style guide for this project?  I personally like the john papa style guide, and they already have it documented to the T.  No need to reinvent the wheel IMO.

ilanbiala:
@codydaig not nested functions. I'm not sure how you get that. Look at this example with bluebird (the concept holds for promise as well):

```
// orders.between
exports.between = function findOrdersBetween(d1, d2) {
  var p = Promise.defer();

  Order
		.find({})
		.where('created').gte(d1).lte(d2)
		.sort('created')
		.populate('items')
		.exec(function(err, orders) {
			if (err) {
				p.reject(err);
			} else {
				p.resolve(orders);
			}
		});

	return p.promise;
};

function createDay(orders) {
	var p = Promise.defer();

	var today;

	today = new Day() // options omitted

	today.save(function(err, today) {
		if (err) {
			p.reject(err);
			console.error(err);
		} else {
			p.resolve(today);
		}
	});

	return p.promise;
}

function emailCSV(csvContents) {
[...]

PR #1097, opened 2015-12-10
feat(config): Deprecate JSHint in favor of ESLint

feimosi:
This is the basic ESLint setup extending well-known Airbnb code style and dedicated Angular rules.
What also has changed:
- removed eslint task from `client.views` path in gulp watch, possibly we can think of using some html linter
- changed most of the globals by `environment` property, still I'd recommend not using / limit the use of global variables by dependency injection
- the previous contributor's rules are left in place, but we need to review them to make the migration seamless. For now it's not consistent with JSHint, so someone with longer experiences in the project could help convert all the rules - docs are quite well-written
- I haven't used Grunt much, so even though it was a basic change, someone could take a closer look :)

References #1072


mleanos:
One thing to keep in mind is that we've decided on using John Papa's Angular Style-guide. I haven't taken a close look at this PR, but does the ESLint configuration take this into consideration? Or are there conflicting styles?

https://github.com/meanjs/mean/issues/943

feimosi:
ESLint itself has only general purpose JS linting, but the plugin I've added has the ability to lint Angular apps. As they write [in the summary](https://github.com/Gillespie59/eslint-plugin-angular#summary): "some rules defined in John Papa's Guideline have been implemented". I just don't know how the defaults follow your conventions.

codydaig:
@feimosi Thanks for the PR!

I am in favor of ESLINT, however, it doesn't match our style guide currently. There are a lot of linting errors, and the handful that I picked out and looked at I do not agree with. 

Ex.) 
In creating an object, it should not have a trailing comma
Incorrect: `{ foo: 'bar', }`
Correct: `{ foo: 'bar',}`

Ex.) 
`Strict mode is not permitted.` 
We want to have 'use strict' enabled. 

Can you go through the errors and see if we can adjust the linter accordingly. If there are actual issues somewhere, fix those in a separate PR to make this PR pass the tests. 

Thanks!

feimosi:
@codydaig Ok, sure, I'll go through them when I find some time.

rhutchison:
@codydaig, I am confused by this - typo?

> Ex.) 
> In creating an object, it should not have a trailing comma
> Incorrect: { foo: 'bar', }
> Correct: { foo: 'bar',}

rhutchison:
@feimosi I am in favor of this PR, but have questions about maintaining a single eslintrc file for this project.

I noticed that you have added the reference to angular default config, but I am curious what you have noticed when running these rules against the server-side files.  Also, I have not seen much progress with adopting the style guide - are these rules failing?

Should nodejs lint rules be separated from angularjs rules?  Can they be the same and still be restrictive enough?

feimosi:
@rhutchison sorry, I wasn't available lately to take care of this PR. 
Regarding the Angular config, I haven't noticed any conflicts so far. It looks like those rules are only fired for Angular-specific code. ESLint allows you to specify separate configs for each folder, but unfortunately not for filename patterns, nonetheless it's not a problem for now.
So I think we can keep both rule sets together.

ilanbiala:
@feimosi still have lots of errors in the build. Can you take a look at them?

package.json
```diff
@@ -66,6 +66,9 @@
   },
   "devDependencies": {
     "coveralls": "^2.11.4",
+    "eslint-config-airbnb": "^2.0.0",
```
ilanbiala: These should be ~

ilanbiala:
@feimosi can you rebase and fix up some of the changes requested above so we can merge?

feimosi:
@ilanbiala I need some more time for this PR. Too busy these days. 
Regarding angular-plugin, eventually I decided to remove it as I've noticed, in same cases it indeed interferes with node's code.

trendzetter:
I had too disable eslint and rely only on jshint because it gave me too many warnings on code I migrated. The --fix option in the command line eslint is also broken so I couldn't fix the thousands of indentation errors which were cluttering my screen.
How do you guys correct styling errors automatically?

ilanbiala:
@trendzetter I don't think anyone has corrected errors automatically.

feimosi:
@trendzetter Last time I tried autocorrect it was working pretty badly, so I wouldn't rely on it.
@ilanbiala The config is ready. Now someone needs to review it. There are still ~150 errors, but I think they should be fixed and not muted by the rules. Also I had to turn off several rules following JSHint config, but I really think you should prepare a more strict style guide.

ilanbiala:
@feimosi there shouldn't be any errors. JSHint never had any, so we should be able to switch to ESLint with whatever properties there and still not have any errors.

feimosi:
@ilanbiala The linting rules are already very loose and there are many inconsistencies throughout the codebase. If I mute the given errors, the sense of using ESLint slowly fades out.

trendzetter:
I could reduce the indentation and other styling errors for code I imported from close to 2000 to 10 using **js-beautify**
Then I fixed the remaining remarks manually. I am now running eslint with gulp again.

I noticed that jshint is running again when nodemon reloads saved files. I appears eslint isn't. Will there still be error checking if save changes to your files after deprecating jshint?

lirantal:
It's quite obvious that this PR should bring with it actual fixes to all the errors and warnings so that eslint can actually go in.
@feimosi are you going to take care of it?

lirantal:
@ilanbiala we can figure out which rules trigger errors and discuss on whether we should 'drop the rule' to make it compatible with what we have today or adopt that rule and make the necessary changes.
What do you think of this plan?

ilanbiala:
Yeah, that sounds fine. First we need to rebase and go over each type of error we currently get. Though I'd rather rebase because some code will change and the errors may disappear/more may appear.

feimosi:
That was exactly my point. ESLint introduces much more new rules, so they should be discussed by the core team and decided whether to keep them or not.

ilanbiala:
@feimosi can you point us to the new ones/ones we don't use specifically? Also, can you rebase before we go any further?

feimosi:
@ilanbiala These rules are failing:
```
comma-dangle
comma-spacing
consistent-return
default-case
guard-for-in
key-spacing
keyword-spacing
new-cap
no-alert
no-spaced-func
no-throw-literal
no-trailing-spaces
no-undef
no-unneeded-ternary
no-unreachable
one-var
one-var-declaration-per-line
quote-props
spaced-comment
warnings)
wrap-iife
```
You can always check it with this command:
``` eslint modules/**/*.js | awk 'NF>1{print $NF}' | sort | uniq ```

**Comment by ilanbiala** [comment]
"comma-spacing": [2, {"before": false, "after": true}]
"key-spacing": [2, {"beforeColon": true, "afterColon": false}]
"keyword-spacing": [2, {before: true, after: true}]
"new-cap": [2, {"capIsNewExceptions": []}]
add guard-for-in, we should enforce that
add no-spaced-func
add no-throw-literal
add no-trailing-spaces
add no-undef
add no-unneeded ternary
add no-unreachable
no need for one-var
"one-var-declaration-per-line": [2, "always"]
no need for quote-props
"spaced-comment": [2, "always"]
"wrap-iife": [2, "outside"]

That should cover all of them I believe.

**Comment by feimosi** [comment]
@ilanbiala Are you sure about `"key-spacing": [2, {"beforeColon": true, "afterColon": false}]` ? It causes 1.5k extra errors and is not auto-fixable by ESLint

**Comment by ilanbiala** [comment]
Sorry, "key-spacing": [2, {"beforeColon": false, "afterColon": true}]

**Comment by feimosi** [comment]
Now everything should be fine.

File: gulpfile.js [hunk 2]
```diff
@@ -57,23 +57,24 @@ gulp.task('watch', function () {
...
   } else {
-    gulp.watch(defaultAssets.server.gulpConfig, ['jshint']);
+    gulp.watch(defaultAssets.server.gulpConfig, ['eslint']);
     gulp.watch(defaultAssets.client.views).on('change', plugins.livereload.changed);
   }
 
   if (process.env.NODE_ENV === 'test') {
     // Add Server Test file rules
     gulp.watch([testAssets.tests.server, defaultAssets.server.allJS], ['test:server']).on('change', function (file) {
+      /* eslint no-unneeded-ternary: 0 */
```
ilanbiala: What's this for?

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -35,6 +35,7 @@ exports.read = function (req, res) {
 
   // Add a custom field to the Article, for determining if the current User is the "owner".
   // NOTE: This field is NOT persisted to the database, since it doesn't exist in the Article model.
+  /* eslint no-unneeded-ternary: 0 */
```
ilanbiala: Why here?

modules/core/client/services/menus.client.service.js
```diff
@@ -155,9 +159,11 @@ angular.module('core').service('Menus', [
 
       // Search for menu item to remove
       for (var itemIndex in this.menus[menuId].items) {
-        for (var subitemIndex in this.menus[menuId].items[itemIndex].items) {
-          if (this.menus[menuId].items[itemIndex].items[subitemIndex].state === submenuItemState) {
-            this.menus[menuId].items[itemIndex].items.splice(subitemIndex, 1);
+        if (this.menus[menuId].items.hasOwnProperty(itemIndex)) {
+          for (var subitemIndex in this.menus[menuId].items[itemIndex].items) {
```
ilanbiala: hasOwnProperty inside this for?

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
```diff
@@ -7,7 +7,9 @@ angular.module('users').controller('SocialAccountsController', ['$scope', '$http
     // Check if there are additional accounts
     $scope.hasConnectedAdditionalSocialAccounts = function (provider) {
       for (var i in $scope.user.additionalProvidersData) {
-        return true;
```
ilanbiala: Why do we have a return in a for loop to begin with?

scripts/reset-password.js
```diff
@@ -81,6 +81,7 @@ mg.connect(function (db) {
       if (processedCount === 0) {
         console.log(chalk.yellow('No users were found.'));
       } else {
+        /* eslint  no-nested-ternary:0 */
```
ilanbiala: Why here?

ilanbiala:
@feimosi like the changes. Let me know about the couple comments I left.

gulpfile.js
feimosi: I wasn't sure whether `onlyChanged` is boolean. If it is we can remove this ternary operator.

modules/articles/server/controllers/articles.server.controller.js
feimosi: When I was trying to get rid of this ternary operator one of the tests was failing, so I've put an exception. Any ideas to solve it in a better way?

modules/core/client/services/menus.client.service.js
feimosi: You wanted me to enforce `guard-for-in`, so we can either check for `hasOwnProperty` or put an ESLint exception.

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
feimosi: I guess it's a way to check whether the object has at least one key. Probably not the cleanest way, but to satisfy `guard-for-in` I had to put `hasOwnProperty` check.

scripts/reset-password.js
feimosi: That was the easiest way, not to interfere with the code too much. If you can provide `if`'`s version of this, I'll replace it.

gulpfile.js
ilanbiala: @mleanos it is a boolean right?

modules/articles/server/controllers/articles.server.controller.js
ilanbiala: Why can't it just be `article.isCurrentUserOwner = req.user && article.user && article.user._id.toString() === req.user._id.toString();`

modules/core/client/services/menus.client.service.js
ilanbiala: No I'm saying I think there should be one inside the loop on L163.

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
ilanbiala: Can you try to fix this to be more logical instead of us just band-aiding it if possible?

ilanbiala: @mleanos do you know what we're doing here exactly?

scripts/reset-password.js
ilanbiala: @feimosi can you see if you can rework this to not need the nested ternary?

modules/articles/server/controllers/articles.server.controller.js
feimosi: Because then if `req.user` is `undefined` we get returned `undefined` instead of `false`. That was failing one test.

modules/core/client/services/menus.client.service.js
feimosi: Do you mean just one check of `hasOwnProperty`? We have to do it in both for loops, if we want to be sure it's the object's own property.

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
feimosi: If it's doing what it looks like doing - checking whether an object has at least one key - I have no idea how to write it in a cleaner way. It's a bigger code design flaw.

scripts/reset-password.js
feimosi: @ilanbiala done

modules/articles/server/controllers/articles.server.controller.js
ilanbiala: What about `article.isCurrentUserOwner = !!(req.user && article.user && article.user._id.toString() === req.user._id.toString());`?

modules/core/client/services/menus.client.service.js
ilanbiala: Yes, but it's missing in the inner one, no? I think we need to add 1 here is what I'm saying.

gulpfile.js
mleanos: Technically.. It will either be `undefined` when not provided, or `true` when it is.

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
mleanos: What this is checking for is whether or not the User has any data in the `additionalProvidersData` field. Unfortunately, it's not an array so we can't just check for length. This function should really be just one line:
`return Object.keys($scope.user.additionalProvidersData).length;`

mleanos: Also, I don't know why we have `provider` there as a parameter. It's not being used. We probably should remove it.

gulpfile.js
ilanbiala: we can just do `var runOnlyChangedTestFile = !!argv.onlyChanged;`

ilanbiala:
@feimosi can you take a look at the other few comments?

gulpfile.js
feimosi: Ok, done

modules/articles/server/controllers/articles.server.controller.js
feimosi: This should do the trick.

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
feimosi: Done

feimosi:
@ilanbiala I've fixed the last few issues

ilanbiala:
@feimosi LGTM. @mleanos @lirantal @codydaig :eyes:? Once someone else reviews, @feimosi can you rebase and squash and change the commit message to follow our commit format.

lirantal:
LGTM, I need to do a more thorough review through all the changes.

Without a doubt an amazing endeavor to make all these updates.
Good job @feimosi  and @ilanbiala for pushing this forward!

lirantal:
@ilanbiala @feimosi let's see this goes in so we can quick push for standards on the code base.

ilanbiala:
@feimosi can you rebase so I can merge?

feimosi:
@ilanbiala, sorry for delay. Please take a look at the commit message after rebasing. If it's okay you can proceed with the merge.

ilanbiala:
@lirantal @mleanos final check before I merge?

lirantal:
I'm all up for it.

lirantal:
Just to add though that there are warnings we still have to deal with, such as:
image

ilanbiala:
@lirantal I think that's fine that we have console statements and putting the comment above seems like clutter, no?

ilanbiala:
@feimosi can you rebase and squash?

lirantal:
@ilanbiala yeah not the end of the world.

feimosi:
Done, you'd better hurry with that merge. It starts to get boring fixing the merge conflicts all the time.

ilanbiala:
@feimosi there are some build errors, can you take a look?

package.json
```diff
@@ -68,6 +68,8 @@
   },
   "devDependencies": {
     "coveralls": "~2.11.6",
+    "eslint": "~2.2.0",
```
ilanbiala: Please delete JSHint dependencies too and remove .jshintrc.

feimosi:
@ilanbiala everything's done

mleanos:
I just restarted the build. Looks like a connectivity/time-out error.

ilanbiala:
LGTM, merging. @feimosi thanks for the hard work and following up on this every time!

mleanos:
@feimosi Yes, thank you for all the hard work! No more complex/unnecessary ternary's, among other styling issues you fixed here ;)

feimosi:
Thanks too, I'm glad I could help! :)

codydaig:
WooHoo!!

trendzetter:
I am not happy, has any of you written any code with this enabled? It's really unbearable.
Plus it contains at least 2 bugs (#1319)  probably not tested well. This is Style over Substance.

config/env/development.js
```diff
@@ -19,16 +19,16 @@ module.exports = {
...
       //  fileName: 'access.log',
       //  rotatingLogs: { // for more info on rotating logs - https://github.com/holidayextras/file-stream-rotator#usage
-      //    active: false, // activate to use rotating logs 
+      //    active: false, // activate to use rotating logs
       //    fileName: 'access-%DATE%.log', // if rotating logs are active, this fileName setting will be used
       //    frequency: 'daily',
       //    verbose: false
       //  }
-      //}
+      // }
```
simison: That extra space breaks indentation here, now space to the next level is 1 space.

trendzetter:
These lint errors are making me angry, sorry for the ranting. I found I can skip rules by setting them them 0 in .eslintrc rules. I am giving it another shot at cleaning things up with the best beautification tricks found so far.

trendzetter:
Just tried the eslint cli again and it has improved in recent times. It now is able to fix some issues automatically using the --fix flag. You would expect it to fix more issues automatically but at least its a beginning.

Issue #1098, opened 2015-12-11
Guest role added to State doesn't allow guest access


mleanos:
When adding the `guest` role to any client route's `data.roles` array, it doesn't have any effect on the resolving of that route; the route is inaccessible to the unauthenticated user.

To reproduce, just add the following to this route config [articles.client.routes](https://github.com/meanjs/mean/blob/master/modules/articles/client/config/articles.client.routes.js#L13)
```javascript
data: {
  roles: ['guest']
}
```

Currently, this route is accessible to unauthenticated user because of how the [core client app init $stateChangeStart](https://github.com/meanjs/mean/blob/master/modules/core/client/app/init.js#L20-L26) is working.

It doesn't take into account that when the `toState.data.roles` has length, but there's no authenticated user. It just continues processing and never checks for *guest* access.

I realize that if a client route has *guest* access, then the `data.roles` probably wouldn't have any other roles defined; and most likely, it would not have this setting on the route. However, this is still a bug because of the intended use of `data.roles` is to allow complete flexibility,& the developer should have complete control.

One should be able to define a client route with `data.roles = ['user', 'admin', 'guest']`, or any combination that suits their needs.

The simple fix is to add `|| role === 'guest'` to the *$stateChangeStart* code like so, and should satisfy this bug..
```javascript
var allowed = false;
  toState.data.roles.forEach(function (role) {
  if ((Authentication.user.roles !== undefined && Authentication.user.roles.indexOf(role) !== -1) || (role === 'guest')) {
    allowed = true;
    return true;
  }
});
```

PR #1100, opened 2015-12-11
fix(core): Client routes guest access bug

mleanos:
Adds a check for the existence of the "guest" role in the client route
state configuration of `data.roles` in the $stateChangeStart of core
client app init configuration.

If this role is present then it doesn't attempt to check authentication
on the route that is being transitioned to.

Fixes https://github.com/meanjs/mean/issues/1098

### DIFF
--- a/modules/core/client/app/init.js
+++ b/modules/core/client/app/init.js
@@ -19,7 +19,7 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
     if (toState.data && toState.data.roles && toState.data.roles.length > 0) {
       var allowed = false;
       toState.data.roles.forEach(function (role) {
-        if (Authentication.user.roles !== undefined && Authentication.user.roles.indexOf(role) !== -1) {
+        if ((role === 'guest') || (Authentication.user && Authentication.user.roles !== undefined && Authentication.user.roles.indexOf(role) !== -1)) {
           allowed = true;
           return true;
         }


modules/core/client/app/init.js
```diff
@@ -16,7 +16,7 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
 
   // Check authentication before changing state
   $rootScope.$on('$stateChangeStart', function (event, toState, toParams, fromState, fromParams) {
-    if (toState.data && toState.data.roles && toState.data.roles.length > 0) {
+    if (toState.data && toState.data.roles && toState.data.roles.length > 0 && toState.data.roles.indexOf('guest') === -1) {
```
codydaig: Can this be moved to inside the forEach loop instead of the if statement? 

I'm not a fan of iterating over an array to check if we need to iterate over an array.

mleanos: @codydaig Sure. I originally had this check in the forEach loop. Then thought it would be more appropriate here; since it doesn't really make sense to go any further, if the `state` allows guest access.

I'll move it back down though. Just wanted to point out my reasoning here. Thanks for the comment.

mleanos:
@codydaig I've addressed your line comment.

codydaig:
@mleanos Just to clarify: If the state it is transitioning too has the role 'guest' it doesn't check authentication anymore. Correct?

mleanos:
Well, it does check since the auth check is on the left-side of the `||` I suppose I should move the check for the *guest* role to the left-side. 

In its its current state, this doesn`t enforce the authentication.

mleanos:
@codydaig Did my last comment address what you were after?

mleanos:
@codydaig

codydaig:
LGTM

modules/core/client/app/init.js
```diff
@@ -19,7 +19,7 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
     if (toState.data && toState.data.roles && toState.data.roles.length > 0) {
       var allowed = false;
       toState.data.roles.forEach(function (role) {
-        if (Authentication.user.roles !== undefined && Authentication.user.roles.indexOf(role) !== -1) {
+        if ((Authentication.user && Authentication.user.roles !== undefined && Authentication.user.roles.indexOf(role) !== -1) || (role === 'guest')) {
```
rhutchison: if this needs to be added, I would put the `role === 'guest'` check first.

mleanos: I was waiting for clarification on that. It's probably more intuitive to have the *guest* role check first.

mleanos: What do you mean by "if this needs to be added"? Do you have another option, or suggestion?

rhutchison: > What do you mean by "if this needs to be added"? Do you have another option, or suggestion?

I haven't tested or know what issue you're fixing - just added my comment.

modules/core/client/app/init.js
```diff
@@ -19,7 +19,7 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
...
           allowed = true;
           return true;
```
rhutchison: I know you didn't add it, but while you're at it, what is the point of return true here?

modules/core/client/app/init.js
```diff
@@ -19,7 +19,7 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
     if (toState.data && toState.data.roles && toState.data.roles.length > 0) {
       var allowed = false;
       toState.data.roles.forEach(function (role) {
```
rhutchison: Probably not best to use Array.forEach if we want to interrupt execution to improve performance.

modules/core/client/app/init.js
mleanos: Ah ok. Just making sure ;) And yea, def more optimized. I'll go ahead amd get that changed.

modules/core/client/app/init.js
mleanos: I see.that too. I was tempted to remove it here, but opted for less changes since this is a bugfix. A few other aspects of this config seems off to me.

modules/core/client/app/init.js
mleanos: Lodash?

rhutchison: I don't think lodash is a dependency right now, and wouldn't want to introduce it for this case.  Can use Array.prototype.some() but not sure about browser support issues.

Probably best to just change to standard for loop, though it's not as elegant :)

mleanos: I'm leaving this for now. I don't think we need lodash on the client-side, just for this, either. 

I don't think there will be a huge performance increase, since we're talking about a small list of roles, that are being iterated. If we decide that we want refactor these loops, then I'd say we look at implementing something like ng-lodash; and do that as a sweeping change.

rhutchison:
LGTM

mleanos:
Thanks guys!

PR #1101, opened 2015-12-13
Adding host configuration feature

farajfarook:
Adding the functionality of configuring the host to start the server. 

By default set this to `127.0.0.1` as the best practice is to run the server locally and expose the port via a proxy like nginx.

### DIFF
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -7,6 +7,8 @@ module.exports = {
     certificate: './config/sslcerts/cert.pem'
   },
   port: process.env.PORT || 8443,
+  // Binding to 127.0.0.1 is safer in production.
+  host: process.env.HOST || '0.0.0.0',
   db: {
     uri: process.env.MONGOHQ_URL || process.env.MONGOLAB_URI || 'mongodb://' + (process.env.DB_1_PORT_27017_TCP_ADDR || 'localhost') + '/mean',
     options: {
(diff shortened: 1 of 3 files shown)


lirantal:
why not bind to 0.0.0.0 all ips?

codydaig:
@farajfarook I would agree with @lirantal that it should be bound to 0.0.0.0 by default. However, I'd also recommend putting this in the production env file, but commented out with a note saying binding it to 127.0.0.1 is much safer in production.

farajfarook:
@lirantal You are welcome, and thank you for the opportunity to contribute. I made the required changes to my commit according to the guidelines and as per @codydaig 's suggestion.

config/env/production.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
     certificate: './config/sslcerts/cert.pem'
   },
   port: process.env.PORT || 8443,
+  host: process.env.HOST || '127.0.0.1',
```
lirantal: @farajfarook we should be consistent and set the production config to 0.0.0.0 too.
IMO it's not *that* much of a security risk because users usually clone on their own machine and set it up later to run on production usually using build tools.

More than that, nowadays people use docker (or other types of VMs) and it is required to bind to a non-loopback host/ip to make the app accessible.

lirantal:
@farajfarook I added another comment, once we're done and decided upon this please squash your commits to just one and we'll merge.

Thanks again for contributing!

farajfarook:
@lirantal Did the changes as per your comment and squashed the commits. :+1: Please merge.

lirantal:
thanks I'll review and merge later on.

codydaig:
LGTM

PR #1107, opened 2015-12-16
fix(core): Remove duplicate angular interceptor

### LINKED ISSUE #1096: Angular interceptor possible dup
Hey, is there any particular reason to have two similar interceptors in [core/services](https://github.com/meanjs/mean/blob/master/modules/core/client/services/interceptors/auth.interceptor.client.service.js) and in [users/config](https://github.com/meanjs/mean/blob/master/modules/users/client/config/users.client.config.js).

IIRC the first one is the most recent and maybe people forgot to delete the latter or merge in the same file possible changes (only one I can think of is `Authentication.user = null;` which is missing in the first one).

pgrodrigues:
Remove the interceptor defined in the users module
Update the interceptor defined in the core module
Update the respective test

Fixes #1096


mleanos:
LGTM.

mleanos:
@codydaig @lirantal Do we want to leave the file `users.client.config`, as a placeholder, but have it empty? Like what we're doing with *articles.server.config*.

pgrodrigues:
I removed it since there was no js logic to be added to the user config, but I can bring the file back if needed.

codydaig:
@mleanos IMO, placeholder files should be in the generator, not laying around in this repo.

ilanbiala:
No strong feelings about it, so I'd say let's just keep the repo uncluttered.

codydaig:
LGTM

rhutchison:
LGTM

codydaig:
Thanks @pgrodrigues.

Good to see your comments again @rhutchison!

PR #1108, opened 2015-12-16
fix(articles): Orphaned User reference throws server error

### LINKED ISSUE #1082: Articles created by deleted users cause server errors
I create an article with user1, then I delete user1 from the system. I log in as user2 and try to view the article created by user1. This causes a server error:
screen shot 2015-11-30 at 10 19 18 pm

We could solve this by checking to make sure `article.user` exists in `articles.server.controller.js`
here:
https://github.com/meanjs/mean/blob/master/modules/articles/server/controllers/articles.server.controller.js#L107
```
    if (article.user) {
      req.article = article;
    }
```

mleanos:
Adds an additional check for the existence of a populated user reference, when determining if the current user has immediate access to the requested article.

Without this fix, the server will throw an error if the requested article doesn't have a populated user field.

Modified the article & articles list view's to check if the article has a populated user. If not, then it will display "Deleted User" in place of the missing user reference.

Fixes #1082

### DIFF
--- a/modules/articles/client/views/list-articles.client.view.html
+++ b/modules/articles/client/views/list-articles.client.view.html
@@ -8,7 +8,8 @@
         Posted on
         <span ng-bind="article.created | date:'mediumDate'"></span>
         by
-        <span ng-bind="article.user.displayName"></span>
+        <span ng-if="article.user" ng-bind="article.user.displayName"></span>
+        <span ng-if="!article.user">Deleted User</span>
       </small>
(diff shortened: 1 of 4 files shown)


lirantal:
LGTM.
@codydaig @ilanbiala ?

ilanbiala:
LGTM, just squash into 1 commit.

mleanos:
@lirantal @ilanbiala Squashed

lirantal:
Thanks @mleanos!

mleanos:
Sure, np! @lirantal

PR #1120, opened 2015-12-28
fix(tests) fixing assertion test for logger which failed

lirantal:
logger test failed on undefined variable


mleanos:
LGTM

mleanos:
Out of curiosity..

Under what conditions did this test fail? Was it when the logger options were not set in the environment config?

codydaig:
LGTM

lirantal:
@mleanos it failed locally, maybe I have a newer shouldjs library, not sure. It wasn't related to the actual logging options in the env config (although that could be a different issue).

It anyway insisted that the options.should object is undefined for some reason.

PR #874, opened 2015-09-02
Article Refactor

trainerbill:
Superceeds https://github.com/meanjs/mean/pull/800

@ilanbiala @rhutchison

PR #1122, opened 2015-12-30
feat(chat): Modify chat module to implement johnpapa styleguide.

rhutchison:
I know this will further stir the pot about which guidelines the project should follow, but the intention is not to delay progress, and instead push forward.

This is not up for debate - but I will take constructive feedback.

### DIFF
--- a/modules/chat/client/controllers/chat.client.controller.js
+++ b/modules/chat/client/controllers/chat.client.controller.js
@@ -1,43 +1,55 @@
-'use strict';
+(function () {
+  'use strict';
 
-// Create the 'chat' controller
-angular.module('chat').controller('ChatController', ['$scope', '$location', 'Authentication', 'Socket',
-  function ($scope, $location, Authentication, Socket) {
-    // Create a messages array
-    $scope.messages = [];
+  angular
+    .module('chat')
+    .controller('ChatController', ChatController);
 
-    // If user is not signed in then redirect back home
-    if (!Authentication.user) {
-      $location.path('/');
+  ChatController.$inject = ['$scope', '$state', 'Authentication', 'Socket'];
+
+  function ChatController($scope, $state, Authentication, Socket) {
+    var vm = this;
+
+    vm.messages = [];
+    vm.messageText = '';
+    vm.sendMessage = sendMessage;
+
+    init();
+
+    function init() {
+      // If user is not signed in then redirect back home
+      if (!Authentication.user) {
+        $state.go('home');
+      }
+
+      // Make sure the Socket is connected
+      if (!Socket.socket) {
+        Socket.connect();
+      }
+
+      // Add an event listener to the 'chatMessage' event
+      Socket.on('chatMessage', function (message) {
+        vm.messages.unshift(message);
+      });
+
+      // Remove the event listener when the controller instance is destroyed
+      $scope.$on('$destroy', function () {
+        Socket.removeListener('chatMessage');
+      });
     }
-
-    // Make sure the Socket is connected
-    if (!Socket.socket) {
-      Socket.connect();
-    }
-
-    // Add an event listener to the 'chatMessage' event
-    Socket.on('chatMessage', function (message) {
-      $scope.messages.unshift(message);
-    });
 
     // Create a controller method for sending messages
-    $scope.sendMessage = function () {
+    function sendMessage() {
       // Create a new message object
       var message = {
-        text: this.messageText
+        text: vm.messageText
       };
 
       // Emit a 'chatMessage' message event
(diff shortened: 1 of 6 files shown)


modules/chat/client/controllers/chat.client.controller.js
```diff
@@ -1,43 +1,55 @@
...
+        Socket.connect();
+      }
+
+      // Add an event listener to the 'chatMessage' event
+      Socket.on('chatMessage', function (message) {
+        vm.messages.unshift(message);
+      });
+
+      // Remove the event listener when the controller instance is destroyed
+      $scope.$on('$destroy', function () {
```
mleanos: can we abstract this out into a service?

rhutchison: That can be a separate PR?

modules/chat/client/controllers/chat.client.controller.js
```diff
@@ -1,43 +1,55 @@
...
-      $location.path('/');
-    }
+  ChatController.$inject = ['$scope', '$state', 'Authentication', 'Socket'];
 
-    // Make sure the Socket is connected
-    if (!Socket.socket) {
-      Socket.connect();
-    }
+  function ChatController($scope, $state, Authentication, Socket) {
+    var vm = this;
```
ilanbiala: Would we rather not do "chat", i.e. more appropriate, less generic names? I think it's useful in the HTML and reads better, but it's up to popular opinion as it won't really affect anything other than variable naming within the controllers.

ilanbiala:
LGTM mostly, @lirantal @codydaig :eyes:?

modules/chat/client/controllers/chat.client.controller.js
codydaig: If your referring to vm, I'd rather leave it at vm for consistency.

modules/chat/client/controllers/chat.client.controller.js
codydaig: Agreed. That should be a separate PR.

codydaig:
I would like to see scope changed to $scope in the tests. But otherwise, LGTM.

modules/chat/client/controllers/chat.client.controller.js
mleanos: Since this is a styling refactor, I thought it would be good to address this as well; if we're strictly adhering to John Papa's style-guide. Might be worthwhile to start the precedent now, with regards to SOC & moving things out into services that shouldn't be in controllers.

It's ok if you want to leave this here though. I don't feel super strong about it :)

modules/chat/client/controllers/chat.client.controller.js
mleanos: I think "vm" is best. If you start getting into more specific names (i.e. chat), then you'll end up dancing around that name, when naming other more ambiguous members/properties of a controller. To me it's very clear what "vm" is when looking at the code; it's very accurately named, IMO.

mleanos:
@codydaig What's the reasoning behind changing **scope** to **$scope**? Just to follow suit with the other services? If that's the case, then I'd say we should remove the "$" from the others.

trainerbill:
LGTM, if it helps.  excited to see progress on standards

codydaig:
@mleanos It's to match how they are actually used. And in the code, it's referred too as $scope, so I don't think we should be changing how we reference it in the tests.

ilanbiala:
Yeah definitely change to `$scope`

ilanbiala:
I think this PR is fine, and we can immediately add another PR for the service @mleanos referred to. @codydaig once the $scope change is made and the commit message is reworded, LGTM and you can merge.

codydaig:
@mleanos @ilanbiala @trainerbill  I think that's the last piece holding up #874 as well is the $ in front of variables and maybe one of the controller/service names.

mleanos:
@codydaig This PR is related to #874? Or you mean the changes we want here are similar to what we have left on #874? I'm eager to get that merged. 

Thanks for clarifying about the `$`. That seems right to me too.

ilanbiala:
Similarity of the $.

rhutchison:
@ilanbiala @codydaig good to merge?

ilanbiala:
@rhutchison LGTM, just fix the commit message so it follows these guidelines https://github.com/meanjs/mean/blob/master/CONTRIBUTING.md#commit-message-guidelines

mleanos:
LGTM. I took it for a test drive, and everything looks good, and is working as expected. 

Thanks @rhutchison

codydaig:
LGTM.

Thanks @rhutchison!!!

PR #1126, opened 2016-01-01
feat(articles): Modify articles module to implement style guidelines.

rhutchison:
Update the articles module to implement the style guidelines.

Much of this work is from @trainerbill - I sent him a PR so he can get credit for the work.  If he doesn't merge and fix up #874 soon, then this is available.

Closes #874
Closes #339


rhutchison:
#874 is ready to merge. closing

ilanbiala:
@trainerbill are we using this PR or #874?

modules/articles/client/services/models/article.client.model.js
```diff
@@ -0,0 +1,19 @@
+(function () {
```
mleanos: @rhutchison We don't need this. It's just a duplication of *article.client.service*

mleanos: Ok. I see you removed *articles.client.service*. However, why add it to a *models* folder & rename with *.model.js*?

Also, the name should `ArticleService` to follow John Papa's guide.

trainerbill:
Close 874.  It's messed up and I am out for a few days.

modules/articles/client/config/articles.client.routes.js
```diff
@@ -12,25 +16,56 @@ angular.module('articles').config(['$stateProvider',
...
+        controllerAs: 'vm'
       })
       .state('articles.create', {
         url: '/create',
-        templateUrl: 'modules/articles/client/views/create-article.client.view.html',
+        templateUrl: 'modules/articles/client/views/form-article.client.view.html',
+        controller: 'ArticlesController',
+        controllerAs: 'vm',
+        resolve: {
+          articleResolve: newArticle
```
ilanbiala: At first glance I thought these all resolved to the same thing. @trainerbill says it saves a little code, but it isn't clear and it makes more sense to just call this the 1-2x necessary in the controller. @mleanos @rhutchison @lirantal @codydaig any other suggestions? This saves a couple lines but it is easy mistaken for the same injection and it just moves the code if it is only used one or two times.

modules/articles/client/controllers/articles.client.controller.js
```diff
@@ -1,84 +1,52 @@
...
-        $scope.content = '';
-      }, function (errorResponse) {
-        $scope.error = errorResponse.data.message;
-      });
-    };
+    vm.article = article;
+    vm.authentication = Authentication;
+    vm.error = null;
+    vm.form = {};
+    vm.remove = remove;
```
ilanbiala: Why not just set the property = to the function definition?

modules/articles/client/controllers/articles.client.controller.js
```diff
@@ -1,84 +1,52 @@
...
-        // Clear form fields
-        $scope.title = '';
-        $scope.content = '';
-      }, function (errorResponse) {
-        $scope.error = errorResponse.data.message;
-      });
-    };
+    vm.article = article;
+    vm.authentication = Authentication;
+    vm.error = null;
```
ilanbiala: Doesn't it initialize error and form by default?

ilanbiala:
@rhutchison nice progress. Definitely some fixes made that aren't in #874. One thing that I'm not sure about, does it make sense to split the CRUD across multiple controllers? I feel like it segments the code a little.

modules/articles/client/config/articles.client.routes.js
mleanos: I don't mind how it is, now that I understand it. When I was looking at this a while back, it was really confusing to me. I don't think it's necessary, but it's probably ok. However, I know others will be confused as well; especially when just looking at the controller code.

It does feel a little weird to have logic, that seemingly belongs in the controller, in the route config when it's not really resolving any data; but just an empty new instance of an Article.

mleanos:
@rhutchison I've pulled this branch down & tested it. Everything seems to be working as expected. I think we're close on this. 

Why are some of the client tests disabled?

Lastly, I don't think we should go with the new *models* directory, and changing the name of the file to *.model.js*. I know this isn't truly a service, but rather a model. However, it's still functioning like a service at the moment; nothing has changed so why make this change now? I don't see the point even If it's only to prepare for the use of models on the client-side. As well, I don't see it making anything else any more clear to our users.

modules/articles/client/controllers/articles.client.controller.js
rhutchison: Yes, but you would only know that by reading through all of the views.  Listing it in the controller gives you a quick glance at what is available to you.

rhutchison:
> @rhutchison nice progress. Definitely some fixes made that aren't in #874. One thing that I'm not sure about, does it make sense to split the CRUD across multiple controllers? I feel like it segments the code a little.

@ilanbiala If you are referring to ListController vs Controller, I would say that it absolutely makes sense for lists vs model.  If you just look at the controllers, you will notice there is no overlap - a lot of extra bindings.  When the list controller gets more functionality later, you will continue to see they are unrelated (bulk functionality, sorting, filters, etc.)

rhutchison:
> Why are some of the client tests disabled?

@mleanos I fixed 2 of the 3.  The 3rd I am not exactly sure how to implement atm, so I'll leave it there as a reminder.

modules/articles/client/controllers/articles.client.controller.js
rhutchison: Ref: [Bindable Members Up Top](https://github.com/johnpapa/angular-styleguide#style-y033)

Setting anonymous functions in-line can be easy, but when those functions are more than 1 line of code they can reduce the readability. Defining the functions below the bindable members (the functions will be hoisted) moves the implementation details down, keeps the bindable members up top, and makes it easier to read.

modules/articles/client/config/articles.client.routes.js
rhutchison: I am not sure I understand the concern?  We are resolving a new instance of an article which is being injected to the controller.  Unless you want to have a separate controller for create vs edit/view, you need to resolve something.  One can argue that you can resolve null and then check for null in the controller and instantiate there, but what is the point when you can avoid it?  Adding another controller will just duplicate code.

modules/articles/client/services/models/article.client.model.js
rhutchison: Because it's a model, not a service.

modules/articles/client/controllers/articles.client.controller.js
mleanos: I a huge fan of this way of defining memebers. I use this in a couple of my projects. The code is so much easier to read, and manage.

modules/articles/client/services/models/article.client.model.js
mleanos: Sure, symantically speakijg, it's a model. However, we're treating it as a service; the $resource object is ambigious as to whether it's a model or service. And now this moves a file around, and adds a new directory to our structure. 

I just don't see the point in this change if we're not modifying the implementation here. Why even introduce it, just for the sake of making a point that its a model. 

I would rather wait, and address client-side models. We'll then be able to take part in a meaningful discussion, rather than tacking on a new idea that isn't relevant to the current PR. 

If everyone else is fine with it, then I'll support it. I wanted to at least point it out, and get feedback from others. 

@codydaig @ilanbiala @lirantal

modules/articles/client/controllers/articles.client.controller.js
mleanos: +1 on what Ryan is saying.

modules/articles/client/services/models/article.client.model.js
ilanbiala: I'd say if Angular calls it a service, so should we. Let's not change it just because it matches server-side wording or whatever else cosmetic reason we may have.

modules/articles/client/config/articles.client.routes.js
ilanbiala: @rhutchison in one place, you resolve newArticle -> articleResolve, and in another getArticle -> articleResolve, and they are both referenced as article in the controller...that's confusing because you have to know the state as well when you are writing code. I'd rather not do that.

rhutchison: They both return a model of the same type - one is instantiated, the other is retrieved from the server.

articleResolve is Article in both cases

modules/articles/client/services/models/article.client.model.js
rhutchison: In this case, angular calls it a factory, not service.  If we are going to have this rigid way of thinking, then menu's should be [menuRun](https://github.com/meanjs/mean/pull/1126/files#diff-7868497d6ee128c169faf0cb57b8f855R6)?

The style guide leaves this type of detail up to interpretation.

mleanos:
Tested, and this LGTM. Thank you Ryan!!

modules/articles/client/config/articles.client.routes.js
ilanbiala: Yes, but it's different and it depends on the state, and that isn't always crystal clear and people may confuse that.

codydaig: Can we add a comment here so people diving into the code knows what is going on?

codydaig:
Made a suggestion to add a comment where there seems to be a point of confusion for people. Overall, LGTM. Nice work!

modules/articles/client/config/articles.client.routes.js
ilanbiala: Honestly I think it makes sense just to do this in the controller. It's less confusing and self-documenting that way.

trainerbill: One additional benefit is to share the resolved value among nested views.  I use nested views all the time, so if I pass the Article in the resolve then populate it in one controller that value gets shared among my other views and controllers.  If you instantiate the model in the controller the nested views would not get that data.  I think we should start looking at nested views for much of the project and this is a good start.

mleanos: @trainerbill I think I like what you're saying. However, I have a couple questions.. 

1) Are you sharing the *resolved* data among multiple child states?
2) Do your nested views have their own states, and/or controllers?

Can you elaborate on the use case of your nested views, or give an example (or gist) of what this implementation looks like?

trainerbill: 1.  Yep
2.  Sometimes.  You can define a new controller otherwise it inherits the parents.  Also inherits all the resolves.

This is just off the top of my head:

https://gist.github.com/trainerbill/f51897d1a1e9ab5eaf26

So in my layout.html default server file i have <div ui-view="hero"> and ui-view="main"

I target main:

https://gist.github.com/trainerbill/f51897d1a1e9ab5eaf26#file-gistfile1-txt-L12

And put a layout file in it.  This is a 2-4 column grid with 4 rows all with new ui-views like fluid-col-1-row-1 that I can now target in my nested states.  Its tough to explain because articles is basic, but if we also wanted to save references with the article then we can have separate views that show the reference information and since it is using articleResolve it is shared so when the main article is saved so are the references.

Here is an abstract route that just resolves the article:
https://gist.github.com/trainerbill/f51897d1a1e9ab5eaf26#file-gistfile1-txt-L48

Now on the edit route my one view has the SaveArticleController which has the saving logic
https://gist.github.com/trainerbill/f51897d1a1e9ab5eaf26#file-gistfile1-txt-L84

and the references has no controller because it is using the ResolveArticleController, which just sets vm.article = articleResolve and so now both views have article resolve.  If I update anything in one the other gets the update.

I think we need to move to nested views across the whole project and need to bring it up for discussion.  It allows you to reuse views all over the place.  For instance if I have a author module then I can use the view article easily, by resolving the article in the route and copy and pasting the view.  done.

ilanbiala:
@lirantal LGTY?

mleanos:
@codydaig @ilanbiala Can we get this merged? It's been sitting for almost two weeks.

I don't think there's anything else up for debate here. Cody's suggestion for a comment for clarification isn't that vital; and I think users will be able to figure it out by just looking at the route configs, even though it's not really clear from just the controller.

I don't mean to be pushy, but we're so close to the finish line with the style-guide. Also, I've been putting off a PR while waiting for this.

codydaig:
@mleanos @lirantal @ilanbiala I'm cool with getting this merged. I think @ilanbiala is on board as well, just waiting final approval from @lirantal.

ilanbiala:
@codydaig let's just merge this and if @lirantal has suggestions, we'll do a new PR.

PR #1134, opened 2016-01-13
fix(users): Removing redundant variables

farajfarook:
Removing redundant variables defined and not used

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -22,11 +22,8 @@ exports.signup = function (req, res) {
   // For security measurement we remove the roles from the req.body object
   delete req.body.roles;
 
-  // Init Variables
+  // Init user and add missing fields
   var user = new User(req.body);
-  var message = null;
-
-  // Add missing user fields
   user.provider = 'local';
   user.displayName = user.firstName + ' ' + user.lastName;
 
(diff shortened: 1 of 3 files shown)


modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -24,7 +24,6 @@ exports.signup = function (req, res) {
 
   // Init Variables
```
ilanbiala: Just combine this block and the add missing user fields into 1 block

ilanbiala:
LGTM other than the one suggestion I made above.

modules/users/server/controllers/users/users.authentication.server.controller.js
farajfarook: I'm sorry I didn't get you.

codydaig:
After @ilanbiala's suggestions are made, LGTM.

modules/users/server/controllers/users/users.authentication.server.controller.js
ilanbiala:
```
// Init user and add missing fields
var user = new User(req.body);
user.provider = 'local';
user.displayName = user.firstName + ' ' + user.lastName;
```

farajfarook:
changes done as per @ilanbiala 's suggestion.

codydaig:
LGTM

lirantal:
LGTM

PR #1149, opened 2016-01-18
fix(articles): Article edit/delete validation

### LINKED ISSUE #1146: [bug] Article edit/remove buttons ng-show logic uses user._id
This line, in the [Article edit view](https://github.com/meanjs/mean/blob/master/modules/articles/client/views/view-article.client.view.html#L5), is attempting to validate if the current User is the *owner* of the Article being edited. Since we're not storing the User's `_id` field on the [client-side User object](https://github.com/meanjs/mean/blob/master/modules/core/server/controllers/core.server.controller.js#L12-L21) any longer, it cannot evaluate properly.

To reproduce this...
1) Login with a User that has created Article(s) (or create an Article after log in). 
2) Go to an article that the user *owns* and you'll see that the edit & remove buttons are present
3) Do a hard refresh of the browser (Ctrl+F5), and you'll see the button are now gone

I know the decision to remove the `_id` field from the client-side User object was intentional, for security reasons. I'm unclear as to the exact implications of having the `_id` field; can someone shed light on that for me?

A simple, but *hackish* way of fixing this would be to compare the User's `username` field, with the Articles `user.username` field; which means we'd have to populate the latter on the Article object when we retrieve it from the database. 

A better solution that comes to mind is to handle the validation in the controller, by retrieving the User's info from the back-end & then setting a property on the view model (`vm`) that specifies if they *own* the document. This option requires more changes than the simple solution I described above, but is probably better practice.
Something like this..
```javascript
function isArticleOwner() {
  // we need to load the user from the backend somehow
  // maybe, with a resolve User in the client route config
  if (user._id.toString() === article.user._id.toString()) {
    return true;
  } else {
    return false;
  }
}
```
Any ideas?

mleanos:
Adds validation of the owner of an Article to fix a bug with the current
validation logic, in the edit Article view.

Since we're not storing the User's `_id` field in the client-side User
object any longer, the validation of whether or not the current User is
the owner of an Article is broken.

These changes fix the issue, by adding a property to the vm in the
articles client controller; `vm.isOwner`. This property will determine
whether or not the edit & delete buttons are available to the currently
logged in User when viewing an Article.

Added a method to the Users client service, for retrieving the currently
logged in User data; `Users.me` performs a GET request to
*api/users/me*.

Fixes #1146

### DIFF
--- a/modules/articles/server/controllers/articles.server.controller.js
+++ b/modules/articles/server/controllers/articles.server.controller.js
@@ -30,7 +30,14 @@ exports.create = function (req, res) {
  * Show the current article
  */
 exports.read = function (req, res) {
-  res.json(req.article);
+  // convert mongoose document to JSON
+  var article = req.article ? req.article.toJSON() : {};
+
+  // Add a custom field to the Article, for determining if the current User is the "owner".
+  // NOTE: This field is NOT persisted to the database, since it doesn't exist in the Article model.
+  article.isCurrentUserOwner = req.user && article.user && article.user._id.toString() === req.user._id.toString() ? true : false;
+
+  res.json(article);
 };
 
 /**
--- a/modules/articles/client/views/view-article.client.view.html
+++ b/modules/articles/client/views/view-article.client.view.html
@@ -2,7 +2,7 @@
   <div class="page-header">
     <h1 ng-bind="vm.article.title"></h1>
   </div>
-  <div class="pull-right" ng-show="vm.authentication.user._id == vm.article.user._id">
+  <div class="pull-right" ng-show="vm.article.isCurrentUserOwner">
     <a class="btn btn-primary" ui-sref="articles.edit({ articleId: vm.article._id })">
       <i class="glyphicon glyphicon-edit"></i>
     </a>
--- a/modules/articles/tests/server/article.server.routes.tests.js
+++ b/modules/articles/tests/server/article.server.routes.tests.js
@@ -400,6 +400,163 @@ describe('Article CRUD tests', function () {
     });
   });
 
+  it('should be able to get a single article if signed in and verify the custom "isCurrentUserOwner" field is set to "true"', function (done) {
+    // Create new article model instance
+    article.user = user;
+    var articleObj = new Article(article);
+
+    // Save the article
+    articleObj.save(function () {
+      agent.post('/api/auth/signin')
+        .send(credentials)
+        .expect(200)
+        .end(function (signinErr, signinRes) {
+          // Handle signin error
+          if (signinErr) {
+            return done(signinErr);
+          }
+
+          // Get the userId
+          var userId = user.id;
+
+          // Save a new article
+          agent.post('/api/articles')
+            .send(article)
+            .expect(200)
+            .end(function (articleSaveErr, articleSaveRes) {
+              // Handle article save error
+              if (articleSaveErr) {
+                return done(articleSaveErr);
+              }
+
+              // Get the article
+              agent.get('/api/articles/' + articleSaveRes.body._id)
+                .expect(200)
+                .end(function (articleInfoErr, articleInfoRes) {
+                  // Handle article error
+                  if (articleInfoErr) {
+                    return done(articleInfoErr);
+                  }
+
+                  // Set assertions
+                  (articleInfoRes.body._id).should.equal(articleSaveRes.body._id);
+                  (articleInfoRes.body.title).should.equal(article.title);
+
+                  // Assert that the "isCurrentUserOwner" field is set to true since the current User created it
+                  (articleInfoRes.body.isCurrentUserOwner).should.equal(true);
+
+                  // Call the assertion callback
+                  done();
+                });
+            });
+        });
+    });
+  });
+
+  it('should be able to get a single article if not signed in and verify the custom "isCurrentUserOwner" field is set to "false"', function (done) {
+    // Create new article model instance
+    var articleObj = new Article(article);
+
+    // Save the article
+    articleObj.save(function () {
+      request(app).get('/api/articles/' + articleObj._id)
+        .end(function (req, res) {
+          // Set assertion
+          res.body.should.be.instanceof(Object).and.have.property('title', article.title);
+          // Assert the custom field "isCurrentUserOwner" is set to false for the un-authenticated User
+          res.body.should.be.instanceof(Object).and.have.property('isCurrentUserOwner', false);
+          // Call the assertion callback
+          done();
+        });
+    });
+  });
+
+  it('should be able to get single article, that a different user created, if logged in & verify the "isCurrentUserOwner" field is set to "false"', function (done) {
+    // Create temporary user creds
+    var _creds = {
+      username: 'temp',
+      password: 'M3@n.jsI$Aw3$0m3'
+    };
+
+    // Create temporary user
+    var _user = new User({
+      firstName: 'Full',
+      lastName: 'Name',
+      displayName: 'Full Name',
+      email: 'temp@test.com',
+      username: _creds.username,
+      password: _creds.password,
+      provider: 'local'
+    });
+
+    _user.save(function (err, _user) {
+      // Handle save error
+      if (err) {
+        return done(err);
+      }
+
+      // Sign in with the user that will create the Article
+      agent.post('/api/auth/signin')
+        .send(credentials)
+        .expect(200)
+        .end(function (signinErr, signinRes) {
+          // Handle signin error
+          if (signinErr) {
+            return done(signinErr);
+          }
+
+          // Get the userId
+          var userId = user._id;
+
+          // Save a new article
+          agent.post('/api/articles')
+            .send(article)
+            .expect(200)
+            .end(function (articleSaveErr, articleSaveRes) {
+              // Handle article save error
+              if (articleSaveErr) {
+                return done(articleSaveErr);
+              }
+
+              // Set assertions on new article
+              (articleSaveRes.body.title).should.equal(article.title);
+              should.exist(articleSaveRes.body.user);
+              should.equal(articleSaveRes.body.user._id, userId);
+
+              // now signin with the temporary user
+              agent.post('/api/auth/signin')
+                .send(_creds)
+                .expect(200)
+                .end(function (err, res) {
+                  // Handle signin error
+                  if (err) {
+                    return done(err);
+                  }
+
+                  // Get the article
+                  agent.get('/api/articles/' + articleSaveRes.body._id)
+                    .expect(200)
+                    .end(function (articleInfoErr, articleInfoRes) {
+                      // Handle article error
+                      if (articleInfoErr) {
+                        return done(articleInfoErr);
+                      }
+
+                      // Set assertions
+                      (articleInfoRes.body._id).should.equal(articleSaveRes.body._id);
+                      (articleInfoRes.body.title).should.equal(article.title);
+                      // Assert that the custom field "isCurrentUserOwner" is set to false since the current User didn't create it
+                      (articleInfoRes.body.isCurrentUserOwner).should.equal(false);
+
+                      // Call the assertion callback
+                      done();
+                    });
+                });
+            });
+        });
+    });
+  });
+
   afterEach(function (done) {
     User.remove().exec(function () {
       Article.remove().exec(done);


mleanos:
I know the tests are failing. I'm working on them right now, but I wanted to get feedback on the implementation of this fix. I'm open to alternative ways of handling this issue.

Also, I'm going to add new tests for the added `vm.isOwner` property.

ilanbiala:
So now edit/delete makes 2 requests? What was the issue with having the _id on the frontend security-wise?

mleanos:
@ilanbiala See this issue https://github.com/meanjs/mean/issues/1106

I don't feel like it's a problem to have multiple requests when viewing an Article. It's a small amount of data, and it's not hitting the database.

I'm still working on the tests. They're a bit complicated since this change uses a combination of the `init` method & Promises (returned from the $resource). Any help/guidance on this would be much appreciated; along with any of the actual implementation here.

rhutchison:
I don't see why this is needed?

mleanos:
@rhutchison Are you referring to the fact that in #807, we're talking about removing the Edit/Create capabilities for user's other than Admin's (or whatever Role we land on)?

If so, I see your point. However, this is fixing a bug. I don't want a bug to be sitting there waiting on a feature enhancement that may never get merged in. 

If that's not what you meant, could you clarify?

vaucouleur:
@mleanos 

1) I'm wondering whether most of the code in validateArticleOwner could be moved into a function of a Factory/Service (for example Users), so that much of this code is factored out and not duplicated across client-controller, something like Users.validateOwner(vm, article) ? The client-side controllers are starting to get quite large; and as it is, this code will be duplicated again and again across controllers. 

2) Here is how I work out permissions; it might be interesting for those that face the same issues. I don't claim that this the right thing to do for everyone, but it works out for me. 

- Basically, I do not want to duplicate the code of the server-side policy file in the client-side code. Said differently, I do not want to have any "vm.authentication.user._id == vm.article.user._id" in the client-side. You can imagine that in many real-life situation, the code is much bigger that this "one-liner". Also sometimes, it requires information that is simply not available on the client side.

- So, what I do, is that I add a route "/api/articles/:articleId/permissions". When one does a GET on this route, he gets back an array that contains the permissions of the current user for the given article, for example ['get', 'post']. Now the interesting thing is that this "array of permissions" can be calculated using the same existing acl code/logic on the server-side (policy file). 

- The key here is that the client-side *needs* to know whether a user has 'post' permission on a given article. But the client-side *does not* need to know how to work-out (calculate) those permissions. Code duplication is not necessary here. Finally, when the article is posted back to the server, the check is done again (as it is done currently, that part remains the same).

mleanos:
@vaucouleur I'm not opposed to the type of implementation you're describing. I had a similar type of permissions route on a resource, in my app; similar to: `/api/books/:id/permissions`. I think we can figure out a better solution than what we have now. It's just a bit more involved.

The question is whether or not we want to go for a change like what you're describing with this PR, or get a fix out for the broken functionality. After all, this PR is attempting to fix a broken feature.

I think we should take this conversation to a new issue for easier tracking, and to not complicate the discussion here.

@ilanbiala Do you have any more feedback here?
@rhutchison Can you clarify your comment?

mleanos:
@vaucouleur Regarding your idea of moving the `validateArticleOwner` to a service.. I had that thought originally, but just went with the controller implementation. 

I think the best place for this method would be in the ArticleService with a method named `$isCurrentUserOwner()`; the $ signifying that it's an instance method. Implementing this in the controller would be quite simple, as we'd just have to call `article.$isCurrentUserOwner()` (which would return a promise), when we have a single article loaded.

By tightly coupling this method with the resource that we're concerned with, we can choose which resources to implement this method on. We always know if we have an authenticated User, and we have the User in context on the server with every request. This makes it easy to implement a feature such as this.

codydaig:
LGTM

modules/articles/client/controllers/articles.client.controller.js
```diff
@@ -48,5 +59,29 @@
...
       }
     }
+
+    // Check if the current user is the owner of the specified Article
+    function validateArticleOwner(article) {
+      var isOwner = false;
+
+      // Get the current user, and handle the resolved promise
+      Users.me()
+      .$promise
```
ilanbiala: indent

ilanbiala:
@mleanos do we need to add a couple of tests testing that a user who isn't the owner can't delete and a user who is can?

Otherwise, LGTM after the indent change.

mleanos:
@ilanbiala I've been planning on adding tests, but I wanted to get feedback first. I didn't want to write tests first, in case the implementation changed.

@ilanbiala @codydaig Any thoughts on moving the validateArticleOwner method into the Articles service, and renaming it to something like "isCurrentUserOwner"? @vaucouleur suggested something similar, and I think it could be a better place to have this check.

I'll write up some tests, and sync this PR.

ilanbiala:
@mleanos I'm not sure if it would make sense to make that move because then the article service depends on the user service, which isn't so ideal. @codydaig I don't think it's necessary because this also seems to be business logic, but unless there is a compelling reason, I think it's fine to leave it in the controller. Let me know if you agree or disagree.

codydaig:
I would agree with @ilanbiala. I don't think we should be injecting the User service in to the Article Service.

vaucouleur:
The Users service is already pulled by the Article controller. If the logic is moved there, there would be no new dependency, and the code would be factored-out instead of being duplicated across generated controllers.

codydaig:
@vaucouleur To me it's best practice to have those dependencies in the controllers rather than the service. Also, this is the only controller that needs it, so why refactor it out?

vaucouleur:
@codydaig That's the only controller that needs it from the point of view of the *generator*. From the point of view of the *generated code*, there will be N instances (one per type of business object: article, book, restaurant, etc.). Not a big deal really, it's not that much duplicated code, but in my opinion, it's better to avoid it.

With respect to "best practice" here is a recommendation from the angular style guide recommending to defer controller logic to services : https://github.com/johnpapa/angular-styleguide#style-y035 .

Again, this proposal is not adding any new dependency, it's the PR in its current form that adds a new dependency. This proposal is just using this **existing** dependency to factor-out code.

mleanos:
I agree with @vaucouleur that we want to put as much of the application logic in services as possible; if it makes sense. However, I'm also torn in how I feel about adding the Users service as a dependency to the Articles service (I don't see it as a huge problem if we did though) as it adds a dependency that won't be used very often; it isn't vital to Articles. @codydaig Could you give a little more context of why this might be a bad choice?

With that said, I'm actually indifferent on whether to add this particular logic to the Articles service, or keep this PR the way it is. As a way of exploration, in how this type of implementation would look, I made some changes locally.

This is what it would look like
```javascript
// Articles service changes
function ArticlesService($resource, Users) {
    var Article = $resource('api/articles/:articleId', {
      articleId: '@_id'
    }, {
      update: {
        method: 'PUT'
      }      
    });

    Article.prototype.isCurrentUserOwner = function () {
      var that = this;

        // Get the current user, and handle the resolved promise
        Users.me()
        .$promise
          .then(successCallback)
          .catch(errorCallback);

        function successCallback(user) {
          if (user) {
            resolve(that.user && that.user._id === user._id);
          }
        }

        function errorCallback(error) {
          reject(error);
        }
      });
    };

    return Article;
  }
```
The Article controller would look like this
```javascript
// Initialize controller
    init();

    function init() {
      // Check if the current user is the owner of the article
      article.isCurrentUserOwner()
        .then(function (isOwner) {
          vm.isOwner = isOwner;
          $scope.$digest();
        });
    }
```
I'm going to be working on the tests tonight, and just wanted to leave this comment to get feedback. I'm fine with whichever way everyone thinks is best.

codydaig:
@vaucouleur @mleanos @ilanbiala 
I can get onboard with the refactor to the service. It's something I've never done in the past, but it does follow the John Papa style guide and after seeing it drawn out, I like it.

mleanos:
Ok.. I'm liking it more now too; not just because you do, i promise :) there's a pun i think..

I'm thinking it's going to be more "testable" in a service method; and the tests aren't very straight forward with this one.

trainerbill:
I am late to the party.

1.  Why are we not storing _id in the user object anymore? The JWT PR still has the _id
2.  The JWT PR hits the user/me route to get the user so why would the article service?  Seems very inefficient
3.  This will break the JWT PR so someone needs to prioritize.

codydaig:
@trainerbill This PR is fixing a bug. Does the JWT PR handle this bug?

mleanos:
How will this break the JWT PR?

The '_id' was removed for a reason. @trainerbill, can you confirm that the same xss vunerabilities are addressed with the JWT?

The way I see it, is that we don't know when/if the JWT PR will be merged in. It's better to have this bug fixed here in a small focused change, and any other PR that comes after will adjust accordingly; even if that means reverting the changes here.

ilanbiala:
@mleanos clarification regarding the implementation, why is it on the prototype? Also, is there no way to avoid the `that = this`?

mleanos:
@ilanbiala I don't think we can avoid using `this`, since we need the Article instance in scope. Also, using the prototype is a way of extending the $resource object for custom methods. 

I think this might be a better approach.
http://sauceio.com/index.php/2014/07/adding-custom-methods-to-data-models-with-angular-resource/

I think we can even get rid of the controller changes using this strategy. 

Additionally, to make it clearer as to what's going on, we can change the `that` line to `var article = this`, since it really is just an Article instance of the $resource. 

WDYT?

ilanbiala:
@mleanos I'm fine with the renaming, but I'm not sure the extend would be better. Can you put together a gist or small snippet to show what it would look like instead of prototype?

mleanos:
I'm travelling right now, and don't have access to my dev machine atm. I'll continue to research right now though. 

Do you have any ideas? I see the issue you started on Angular's repo on this subject. What was your resolutuon?

ilanbiala:
@mleanos I haven't really done too much Angular work as of late :(, but I think that was just a nicety I wanted and I just didn't end up using it.

rhutchison:
Why are you jumping through hoops?  This is so hacked up and not necessary.

Introduce a property on the article that is returned from the server 'IsEditable: true'.  If article.IsEditable then do whatever you need to do.  When you post to the server, you will have validation and if they don't have permissions then you throw an exception.

vaucouleur:
@rhutchison   Precisely ! This is very close to what I described in my point 2).. all the way up in this thread. In my case, I generalized it a bit more by adding a route /api/articles/:articleId/permissions that returns an array of permissions, but a simple property would do just fine.

The key point here is to avoid duplication of the permission checking logic on the client-side. We need the **result** of the permission checking, not the logic itself (not the the "x == y").

mleanos:
@rhutchison @vaucouleur Let's discuss this on Gitter. I'd like to understand both of your perspectives. I'm not quite sure that you're both advocating for the same thing here.

The way I look at it is this..
1) The server permission logic is for determining what *actions* a User can perform on a particular API route. The policies config is setup to handle this; and it's quite effective.
2) The client needs some simple logic/rules that are set in place to determine how the UI should behave with particular data models. This should be completely independent of what the server-side policies are. Any *one* client application may implement a different set of UI controls, while using the same API as other client applications; IMO, this follows simple SOC.

@vaucouleur this is probably directed more at your points, because I don't quite follow what Ryan is saying..
To me, there doesn't seem to be any sort of duplication of "permission logic" going on here. The server sits there waiting for requests, and when one comes in it either allows or denies. The server doesn't care what the client (that's sending the request) allows in terms of what requests it can actually send.

Having clear SOC between the client-side & server-side functionality is going the provide the most flexibility for all the possible implementations that come from a framework like this. A simple example of how we're accomplishing this (right now) is the way we've setup the client-side routing based on the User's `roles`.

I know this may sound repeating.. 
**The server should not care, nor dictate, what UI controls a client-application provides to it's User's. The client-application should merely provide an interface of how it wants it's User's to interact with any API; this can greatly differ from one to the next**

mleanos:
@rhutchison Trying to understand your proposal, but I'm not sure still.. one interpretation I've come to is that you'd like to see something like this..
1) Add an API route: `/api/articles/:articleId/is-current-user-owner`
2) We just add a new instance method on the Articles $resource:
```javascript
isCurrentUserOwner: {
        method: 'GET',
        isArray: false,
        url: 'api/articles/:articleId/is-current-user-owner'
      }
```

vaucouleur:
@mleanos I'm happy to discuss on Gitter, but this is clearly a design discussion directly related to this PR. 

In my view, in my point 2), the server does not dictate anything to the client. 

- The server just says "I am the Server and I *claim* that the current user has this set of permissions". 
- This is completely **declarative**. 
- The client may choose to do something stupid if he wants to (like *always* show the edit button).
- The check is done again on post.

Again: 

* the permission set does **not** say "show the button". 
* It says, "the current user had edit permission on this article". 

This is very different.

Finally, in the current implementation of the PR, the logic is obviously duplicated since we've got equality-checking of the owner ("x == y"), both on the server-side (in the acl code), and on the client-side ! This does not look like much, because this is a one-liner; but in a real-life scenario, the logic is likely to be much more complex (and in some scenarios, the client might not have all the information needed to compute the permission set).

https://github.com/meanjs/mean/blob/d2b2dfd606362547211d5f05b2a4d8013a3417c2/modules/articles/server/policies/articles.server.policy.js#L52

trainerbill:
@mleanos I am guessing this is the commit you are talking about:

https://github.com/meanjs/mean/commit/b9e3fd1625b0e766331b9a673bf2c800f3fed1c1

Why was user id removed.  What was the XSS vulnerability?  The JWT PR stores the user object(-salt -password) in the AuthenticationService and the _id is still in it.  I really don't know whey we are protecting a Mongoose ID, but I am probably missing something.  So I need to know why this is bad, or if it is acceptable practice, then this PR isn't needed as we can go back to checking if user._id = article.user

mleanos:
@vaucouleur I understand that both the client and server check that the User is the owner, with the same equality check. However, this is how we've defined the rules on either side; they happen to be the same. Take my #807 for example. It completely changes who can create/edit/delete articles. There's only one server-side change (besides the tests) to achieve this, that removes an action that one role can perform; see [here.](https://github.com/meanjs/mean/pull/807/files#diff-34394d118c31dcf4a90e8e5094a12204R28) This clearly demonstrates the SOC design consideration I'm advocating for.

In a real-life scenario, the client-side could define what it means to have "edit" rights much differently; and so could the server. IMO, that's why it becomes a SOC consideration in that the client-side **should be able to** define what it means to have "edit" rights, completely independent of what the server-side considers to be "edit" rights. 

What happens when an application comes along, that uses this backend, and wants the edit/delete buttons to show up for not only the User that created the Article? Maybe the User must have a specific Role & no Articles can be edited/deleted on Tuesdays (weird but you never know). The server will always enforce whatever restrictions that are in place, on each of the API routes; the ACL's acts as a gateway between the application & database. Furthermore, an API might be setup to allow any manipulation from any User. Now that same client-application, in my example, still wants **it's** rules to be enforced from it's UI. It can't help if the user makes a request directly to the API with some other UI/tool. But from it's UI, it has defined a set of Rules that show/hide controls.

We might be at an impasse here. However, my main concern is that we must allow for the most flexibility & have proper SOC in place. That way any number of possibilities can be achieved using this framework; no matter how crazy (or similar) the business rules may be.

trainerbill:
@mleanos Can you address my question?  I need to know if having _id on the Authentication service is bad...  If we can leave it in there in the JWT PR then this PR isn't needed as you can check user._id === article.user.

mleanos:
@trainerbill I don't know the full implications of the xss vulnerabilities either. @lirantal might be someone to ask, or follow this [issue's](https://github.com/meanjs/mean/issues/1106) discussion to understand it a bit more.

If it's decided that it's ok to add the `_id` field back into the client-side storage for the User object, whether in a JWT or the $window object that's used now (doesn't matter to me), then yes you're right this PR would be a moot point & not needed.

mleanos:
Ok, so after discussing this with both @rhutchison & @vaucouleur on Gitter, this is what I've gathered..

1) @rhutchison Is advocating for a new Article model field, that isn't mapped to the database but is set right before the Article is sent back to the client. `article.IsCurrentUserOwner = ..(logic)`
&nbsp;&nbsp;a) I feel like this is just a patch & seems like it has a very narrow scope of use. Although, this would work & fairly simple to implement.

2) @vaucouleur Would like to see the ACL's enhanced to include things like "edit", "delete", etc on the back-end routes. These "permissions" would be sent back to the client in a separate API route; something like `/api/articles/:articleId/permissions`.
&nbsp;&nbsp;a) This seems like a good idea in general for an advanced feature set that needs additional security on the API routes. Other than what we have in place right now.
&nbsp;&nbsp;b) Again, this seems like merely a back-end consideration. I don't see why the client application should be concerned with what the ACL's allow.

3) And now, it's up for debate if this PR is needed. It's hinging on whether or not we store the User's `_id` field on the client-side again. 

IMO, if we want to get a fix for this bug out then I say we go with my proposal of adding a new method to the Articles service to get this info. Is it really hacky, as Ryan said? I don't see it as a hack. In fact, I think it could be setting a precedent of how to extend our $resource objects to include domain level logic for the client-side of the application. It's nothing fancy like Angular Model Factory, but it's a step in the right direction; I think using a client-side model library could be a good idea, but that's a different discussion.

What do you all think is the best option? I'll do whatever the consensus is. For now, I'm going to update this PR with my latest proposal so everyone can see what it looks like.

Thanks @ilanbiala @rhutchison @vaucouleur for challenging me on this & starting a great discussion. To me, that's what this collaboration thing is all about :)

trainerbill:
I would like to know @lirantal thoughts on why _id was removed in the first place.  Its not sensitive data, and the backend will deny any request with an _id that isn't yours.  If we can keep _id in the JWT PR then this PR isn't necessary.  I also don't like Users being a dependency of Articles, especially since Authentication already is.

vaucouleur:
@mleanos Thanks for the summary. 

A minor correction: no permissions have to be added to the ACL. My simple point is that the permissions (for the current user and a given article) should be returned to the client side, either via a route or via a property on article. Those are the same permissions that are *already* declared in the ACL. "you just asked for an article, here is what you can do with it: ['get','post']. That's it. Really simple.

Thanks @mleanos for tackling this issue and looking at the big picture. Always nice to have good technical discussions.

mleanos:
@codydaig @ilanbiala @rhutchison @vaucouleur 

I've modified this PR to implement a fix, based on the suggestion & feedback from everyone that has helped out here. These changes are very low impact, and should suffice to fix the issue.

Thanks everyone for the feedback, and for taking the time to discuss all the possible considerations with this issue.

I'll be adding some tests here shortly.

mleanos:
Added 3 server-side route tests. Didn't see any relevant client-side tests, which should be enough since the main changes are on the server side.

ilanbiala:
LGTM, @codydaig anything before I merge?

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -30,7 +30,14 @@ exports.create = function (req, res) {
  * Show the current article
  */
 exports.read = function (req, res) {
-  res.json(req.article);
+  // convert mongoose document to JSON
+  var article = req.article ? req.article.toJSON() : {};
+
+  // Add a custom field to the Article, for determining if the current User is the "owner".
+  // NOTE: This field is NOT persisted to the database, since it doesn't exist in the Article model.
+  article.isCurrentUserOwner = req.user && article.user && article.user._id.toString() === req.user._id.toString() ? true : false;
```
rhutchison: I prefer something that is more future-proof, like isEditable - but glad you cleaned this up.

vaucouleur: I guess the goal was to avoid undefined.. but IMHO.. x ? true : false.. is not super elegant..  what about !!x ?

mleanos: @rhutchison I think `isCurrentUserOwner` is more accurate based on the logic of the condition. It just happens to be that we're using it to show the edit/delete controls. However, that may change & we could still have a need for the `isCurrentUserOwner` field.

mleanos: @vaucouleur I think the way it's written out is more clear. Especially since we need to verify we have a `req.user` object as well. But I agree it's not the most elegant way to do it.

codydaig:
LGTM.

PR #1155, opened 2016-01-20
(fix AuthenticationService) User /me server route instead of $window

trainerbill:
Initial commit.  Seems to work for me so far.  Need someone else to take it for a spin

### DIFF
--- a/modules/users/client/services/authentication.client.service.js
+++ b/modules/users/client/services/authentication.client.service.js
@@ -1,12 +1,35 @@
-'use strict';
+(function() {
+  'use strict';
 
-// Authentication service for user variables
-angular.module('users').factory('Authentication', ['$window',
-  function ($window) {
-    var auth = {
-      user: $window.user
+  angular
+    .module('users')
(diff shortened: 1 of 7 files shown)


trainerbill:
@mleanos checking in here instead of the issue.  I am good with moving to jwt.  I will check out your PR next week.  On vacation starting today.

modules/users/client/services/authentication.client.service.js
```diff
@@ -1,12 +1,35 @@
...
+    var service = {
+      ready: init(),
+      user: null,
+      refresh: init
     };
 
-    return auth;
+    function init() {
+      return $q(function(resolve, reject) {
+        $resource('api/users/me')
```
mleanos: @trainerbill Can we add this api route to the Users service?

I'd rather see this service more focused with methods that pertain specifically to Authentication; logout, signin, and eventually any logic we need for Token-based authentication (JWT).

trainerbill:
Closing in favor of JWT

https://github.com/meanjs/mean/pull/1163

PR #1156, opened 2016-01-22
(fix password-validator) Simple Mistype

Meistercoach83:


modules/users/client/services/password-validator.client.service.js
```diff
@@ -11,7 +11,7 @@ angular.module('users').factory('PasswordValidator', ['$window',
         return result;
       },
       getPopoverMsg: function () {
-        var popoverMsg = 'Please enter a passphrase or password with greater than 10 characters, numbers, lowercase, upppercase, and special characters.';
+        var popoverMsg = 'Please enter a passphrase or password with more than 10 characters, numbers, lowercase, uppercase, and special characters.';
```
ilanbiala: @lirantal @codydaig Why do this when result tells us the errors? I think the popoverMsg should be result.requiredTestErrors if possible.

https://www.npmjs.com/package/owasp-password-strength-test

mleanos: @ilanbiala I think the point is to give the User a heads up on what the requirements are for the password. Rather than displaying the message on the page, I think this is a good way to convey this info to the User.

jloveland: Thanks for catching this @Meistercoach83! @mleanos is right, the popover was intended to give the user instructions for the password requirements when they click into the form field.

ilanbiala:
Fix the commit message, and I'll merge it in.

ilanbiala:
@Meistercoach83 the actual commit message needs to be changed, not the title of the PR. See https://github.com/meanjs/mean/blob/master/CONTRIBUTING.md for what it should look like.

codydaig:
@Meistercoach83 Can you change the commit message?

PR #1160, opened 2016-01-25
fix(gulp): Fixing/updating templatecache gulp task

tfernandez:
The transformUrl replace function has been removed from templatecache gulp task.

The client view routes include the string "/client" in the url from mean.js 0.4.2 and the browser view request is not matching with the angular template cache url, so all the templates cache is being ignored at startup.

The task has been updated.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -204,18 +204,13 @@ gulp.task('imagemin', function () {
 
 // Angular template cache task
 gulp.task('templatecache', function () {
-  var re = new RegExp('\\' + path.sep + 'client\\' + path.sep, 'g');
-
   return gulp.src(defaultAssets.client.views)
     .pipe(plugins.templateCache('templates.js', {
       root: 'modules/',
       module: 'core',
       templateHeader: '(function () {' + endOfLine + '	\'use strict\';' + endOfLine + endOfLine + '	angular' + endOfLine + '		.module(\'<%= module %>\'<%= standalone %>)' + endOfLine + '		.run(templates);' + endOfLine + endOfLine + '	templates.$inject = [\'$templateCache\'];' + endOfLine + endOfLine + '	function templates($templateCache) {' + endOfLine,
       templateBody: '		$templateCache.put(\'<%= url %>\', \'<%= contents %>\');',
-      templateFooter: '	}' + endOfLine + '})();' + endOfLine,
-      transformUrl: function (url) {
-        return url.replace(re, path.sep);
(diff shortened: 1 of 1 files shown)


ilanbiala:
@tfernandez have you tested and confirmed it's working now?

tfernandez:
@ilanbiala, yes, is tested and working now, from meanjs 0.4.2 the client view route matches with the real path, the transformUrl function is generating a wrong url now.

To reproduce the error just execute "gulp prod" and check the "build/templates.js" file, the urls generated does not match with the html loaded by the browser causing the browser loads all the templates on demand and ignoring the cache. 

Probably @rhutchison can confirm the problem.

ilanbiala:
@tfernandez any reason the regex might have existed in the first place? Maybe for cross-OS support? What platform(s) have you tested on?

tfernandez:
@ilanbiala the regex was used inside the "transformUrl" function to replace the "/client/" string from the template realpath with "/" and generate a new valid html view url.

Original/real path:
/modules/core/client/views/home.client.view.html

Generated url:
/modules/core/views/home.client.view.html

However this replacement must not be done from meanjs 0.4.2

rhutchison:
@tfernandez this needs to be tested on Windows.

ilanbiala:
@tfernandez have you tested on windows?

tfernandez:
@ilanbiala tested this afternoon on Windows 10 in cmd and cygwin

The issue exists in the same way and is solved with the same solution.

ilanbiala:
@tfernandez solved with what solution?

tfernandez:
@ilanbiala the problem is the template cache gulp task is removing "/client" from the templatecache uris before saving it, and the solution is remove the regex and the transformUrl method from the task (the target of this pull request)

ilanbiala:
@tfernandez what OSes has this been tested on?

tfernandez:
@ilanbiala has been tested on CentOs 6/7, OSX El Capitan and Windows 10

ilanbiala:
@tfernandez SGTM, just fix the commit message to follow our guidelines in contributing.md, and I'll merge.

codydaig:
Update commit message, then LGTM.

tfernandez:
Changed the commit message

codydaig:
LGTM

rhutchison:
LGTM - merge.

It wasn't very clear on why the change was needed.  I did some testing/digging.

Regression from #758

Issue #1086, opened 2015-12-03
User logged out, but still appears to be logged in


wansco:
I'm not sure if I'm doing something wrong, so please bear with me.

I'm running in dev mode and every so often, I get sent to the /forbidden page on what should have been a valid request. On further investigation, I am actually logged out on the server, but the client UI shows the menu bar and user as if I am still logged in.

I've tracked it down to this:
The server session expires somehow (although I haven't figured out why, exactly).

When transitioning to another state, the server policy isAllowed() fires back a 403, which is reasonable given the unauthenticated user.

On the client, auth.interceptor.client.service.js catches the 403 and sends the user to /forbidden, which is also sort of reasonable, given the unauthenticated user. The problem is that the client still thinks they are authenticated because auth.interceptor.client.service.js doesn't clear out the client-side authentication in that case (on a 403).

I'm not sure if the solution would be to change the 403 to 401 in all the policy isAllowed()'s or to attach a message to the 403 that would clue auth.interceptor.client.service.js into the user's server login state (which would then transition to signin rather than forbidden). Changing to a 401 seems the cleanest solution, but I dont know for sure.

To reproduce this: Login, and then flush out the sessions collection on the backend DB. Also, the stock install of the articles module allows guest access in articles.server.policy.js... you should change line 34 to "user", or go to a page that requires an authenticated user to view.

Incidentally, some of the user management pages do not exhibit this behavior, which might be a bit of a security issue. In this unauthenticated state, I am still able to "view" the Edit Profile page, although I cannot make any changes. That throws back a 400 error, which would probably be better served with a 401 or 403. Either way, its sort of inconsistent as to what happens on the client when the server has logged the user out.

Also, a hard refresh solves the issue, but transitioning between states does not reload the client side authentication.

wansco:
I think the best approach would be to check if the user is set in isAllowed() (on the server policy), and return 403 if the user is set (which then would mean they are both logged in and forbidden), or 401 if the user is not set (which only says they are not logged in, we cant know know if they are forbidden).

The idea there would be a logged in user is forbidden (403), an un-logged in user would ultimately be bounced to the signin page (401, which auth.interceptor would handle, sending the browser to the signin page).

I gave that a shot (401 on an unauthenticated user on the server, in isAllowed()) and now the following line in auth.interceptor.client.service.js is giving me problems.
              $injector.get('$state').transitionTo('authentication.signin');

Its loading "/signin" instead of "/authentication/signin"

[...]

PR #1163, opened 2016-01-26
feat(core): JWT Authentication

trainerbill:
I took a stab at redoing the Authentication in favor of JWT.  I am using passport-jwt to leverage the existing authentication structure.  The Authentication service now holds a user and token and uses localStorage to store the token.  I am only storing the token.  If you refresh the page the AuthService will look for a token in local storage and if it exists hits the users/me route to get the user.  The auth service also sets the $http header for Authorization if the token exists so you don't have to add it to other services.  Currently working:

1.  Login
2.  Sign Up
3.  Persistent on Refresh and users/me route
4.  Edit Profile

I want to get buy in from contributors before going any further.  This PR will touch a lot of files and I don't want to get all the way done and someone has a problem and this ends up in PR hell.  If I can get some feedback and +1's I will have this done by Friday this week.  Also would like a volunteer to help fix the tests when this is done and ready.  Thanks to @mleanos for the server side kick start.

@mleanos @lirantal @ilanbiala @codydaig 

Fixes https://github.com/meanjs/mean/issues/389
Fixes https://github.com/meanjs/mean/issues/1086


config/env/default.js
```diff
@@ -37,5 +37,9 @@ module.exports = {
         fileSize: 1*1024*1024 // Max file size in bytes (1 MB)
       }
     }
+  },
+  jwt: {
+    secret: process.env.TOKEN_AUTH_SECRET || 'M3@N_R0CK5!',
+    expiresInSeconds: process.env.TOKEN_AUTH_EXPIRES_SECONDS || 200 * 60 * 60 // 200 hours
```
ilanbiala: Why 200 hours?

config/lib/authorization.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+// export the token auth service
+module.exports = tokenAuth;
+
+// Sign a token
+function signAuthToken (user) {
+  return new Promise(function (resolve, reject) {
+    try {
+
```
ilanbiala: extra newline

ilanbiala:
Did a quick read through, didn't comment on everything. @trainerbill can you give the pros and cons of JWT and any limitations of this implementation?

mleanos:
@trainerbill This looks pretty good. I'm gonna pull it down and play around with it.

The other day, I started looking at using the [passport-http-bearer](https://github.com/jaredhanson/passport-http-bearer) module. I think the module you're using is probably better, but what are the advantages of this one over the bearer?

One thing I started playing around with, mostly in thought, was to think of MEAN Token-auth as just another Provider. If we use a passport module for authenticating the JWT Token, can we merely treat it as a new Provider, and have it behave the same (just as Google, Twitter, GitHub, etc)?

Think of having a MEAN.JS logo mixed in with the other providers at time of sign in or sign up. The project user could decide if they want to allow either authentication method, or both.

It think it could allow the project to be easily switched from and from JWT & Session based authentication. Any thoughts on that approach?

trainerbill:
@ilanbiala Thanks for the feedback.  This is more of a rough draft than a code review so don't worry about new lines etc.  I am just looking for an overall discussion on the topic.  My goal is to get away from the current AuthService that reads from window.user and JWT is kinda the norm right now.  JWT is becoming the defacto standard for authenticating REST services.  The biggest pro is that it will work with mobile.  Session based auth will not.  A mobile app could be coded to hit the auth/sigin URL store the token and then make additional calls to the MEAN backend.  Most applications are going away from Session based authentications as they are not mobile friendly.  I am sure there are other benefits but those are the 2 I can think of off hand.  With JWT the client holds the keys instead of the server.  As for limitations, I am not aware of any.  Perfomance should be the same as it is the same amount of DB calls, just an added token decode which isn't much.

@mleanos The module you linked looks more like an oAuth strategy.  I like the passport-jwt strategy as it works with our current auth procedure and you can make it route based.  So if I have a public endpoint I don't do the passport.authenticate.  How I implemented it really is just another provider.

https://github.com/trainerbill/mean/blob/53302087a0fc1d2d38a867ca342d707786ca2bd2/modules/users/server/config/strategies/jwt.js

As far as switching between session and JWT, I don't see the point.  If you look at how passport and express currently work, only the _id is stored and there is a DB call to get the rest of the info:

https://github.com/meanjs/mean/blob/master/modules/users/server/config/users.server.config.js#L21

So the JWT approach is the same DB calls, as you decode the token, get the ID and find the user.  Plus it has the added benefit of being client independent.  Mobile devices, and any other client would be able to authenticate and use the REST interfaces.

For your question on if we can treat it as another provider, the answer would probably be no.  You still need a user/pass combo to authenticate a user and the token is merely a reflection of that authentication.  The "local" strategy now just checks the user/pass combo and returns the user info and the token.  This will be how providers work when I get to them.  The same steps as far as auth, but when it comes back we send a token to the AuthService and everything works as normal.

Like the discussion, lets keep it up.  Want everything ironed out before doing more work!

codydaig:
I'm totally onboard with this PR. I like token auth vs. session auth. It makes scaling the app to more than 1 server much easier.

trainerbill:
@codydaig awesome, I will continue to work on this then.  I think it will solve a lot of the other open issues as well

What do you all think of changing up the routes a bit:

https://github.com/trainerbill/mean/blob/JWTAuthorization/modules/users/server/routes/users.server.routes.js

Pros:
1.  Can do router.all() and hit all of the routes on the specific router.  Like the JWT auth
2.  Currently using app.param like we are doing can cause collisions if a submodule wants to use say :userId
3.  Keeps the route logic contained in the module instead of the app.

Cons:
1.  Its different
2.  have to require express and create a new router.

I personally like it, but want some feedback.

codydaig:
I personally am not usually a fan of the express router because it usually adds unnecessary complexity. However, I think it could be of benefit here to help keep things more modular. I think that is a huge undertaking though because I would like to keep it consistent throughout the app.

trainerbill:
@codydaig Yeah I know... but this whole PR is gonna be a huge undertaking...  I know in one of my apps the app.param collision screwed me.

codydaig:
@trainerbill Doing things in smaller PRs is what I usually like to go for. It makes it easier to review, discuss, and get merged. My vote would be to have the router refactor be its own PR.

trainerbill:
@codydaig True.  However adding passport.authenticate(jwt) to every route is going to be kind of a pain.

ilanbiala:
@trainerbill if possible, @codydaig's suggestion of keeping changes in smaller PRs will definitely speed up the whole conversion process. @mleanos has a PR (#1121) that seems to also implement JWT instead of sessions and his is slightly smaller in impact. @trainerbill can you and @mleanos can combine PRs to create a better one that has less impact and incorporates the changes of both?

trainerbill:
@ilanbiala I understand that smaller PR's are easier and will try to keep it as small as possible, but this is a large change.  I have spoken with @mleanos and offered for him to assist.  His current PR isn't close to being complete, and will need 90% of the changes in this PR.

On the plus side I am almost done.  I need some help on the chat module.  Anyone familiar enough with the module and socket.io to make the auth changes?  I found this article but I am not at all familiar with this module or socket.io

http://wmyers.github.io/technical/nodejs/Simple-JWT-auth-for-SocketIO/

mleanos:
@trainerbill I can help out with the Socket.IO changes

My intentions with my PR was to keep the server-side implementation independent from the client-side. Really those two things are separate. It looks like most of your server-side changes are the same that I implemented in my PR. My suggestion from the beginning of working on this, was to work on a shared branch; with client-side & server-side changes made independent of one another.

So shall I submit PR's to your branch?

trainerbill:
@mleanos Absolutely.  I don't think you can really keep client/server in independent PR's.  In order to make sure that it is working you have to remove the session authentication.  Otherwise the app will continue to use session and JWT so you really don't know if the changes you are making are working as expected with JWT.

I will be AWK till a little later tonight and will check back in.  Can some of you pull the branch and take it for a spin?  It should be in a semi workable state.  Whats currently working:

1.  Authentication.  Signin/Signup/Signout
2.  User Routes.  Edit Profile, Password, Picture
3.  Article Module.  Add/Create/Update/List

Whats not working:
1.  Chat Module
2.  Possibly user admin.  Not sure on this if someone wants to check and see
3.  Signout bug.  When on a user profile and sign out it should be forbidden.  Possibly just redirect to home on signout?
4.  Social Accounts.  I need some client/secrets if anyone has them.
5.  Tests.  REally need help here.

@mleanos If you are working on specific items let me know and I will stay away from them.

Thanks Everyone!

codydaig:
@trainerbill `socketio-jwt` is what I've used in the past.

jameslegue:
+1 on this. Have been waiting for token based auth for awhile now (main use case is sharing of REST APIs with mobile).

trainerbill:
Can someone look into CSRF with JWT?  I don't think we need the CSRF PR's after this goes in.

trainerbill:
Anyone have any thoughts on the following file:

config/lib/jwtAuthentication.js

I think it should go in the users module since that is where the authentication is taking place.   Just not sure where to put it, controllers?, and what to name it.

config/lib/jwtAuthentication.js
```diff
@@ -0,0 +1,36 @@
...
+      };
+
+      var options = {
+        expiresIn: config.jwt.expiresInSeconds
+      };
+
+      var token = jwt.sign(payload, config.jwt.secret, options);
+
+      resolve(token);
+
```
ilanbiala: extra newline

config/lib/jwtAuthentication.js
```diff
@@ -0,0 +1,36 @@
...
+var config = require('../config'),
+  jwt = require('jsonwebtoken');
+
+// build token auth service
+var tokenAuth = {
+  signToken: signAuthToken
+};
+
+// export the token auth service
+module.exports = tokenAuth;
```
ilanbiala: I'd rather have this at the bottom because that's where we have most of the  exports. Also, you can just assign the object rather than create one.

trainerbill: will do.  stole this from @mleanos PR.  Do you have an opinion on the file naming and location?  I am going to work on the tests and need to require the file so I want it to be settled on name and location.  See my comment on the main conversation

ilanbiala: - in the names instead of camelcase. I don't see any good reason for it to be in either. @codydaig any suggestions on file location?

trainerbill:
All.  I have this PR about ready.  I will do the Password Reset this afternoon:

1.  @codydaig I made a frame of the auth service test in modules/users/tests/client/authentication.client.service.tests.js.  The describes and its are there but I don't know enough about the client side tests to finish this.  Can you do it or find someone who can?
2.  @mleanos How is the chat module coming along?
3.  @ilanbiala @rhutchison  You can do a first round code review now if you want to.

The coverall tests decreased because it looks like there is a duplicate in the user tests

config/env/development.js
```diff
@@ -60,8 +60,8 @@ module.exports = {
     callbackURL: '/api/auth/github/callback'
   },
   paypal: {
-    clientID: process.env.PAYPAL_ID || 'CLIENT_ID',
-    clientSecret: process.env.PAYPAL_SECRET || 'CLIENT_SECRET',
+    clientID: process.env.PAYPAL_ID || 'AVWxUS4uQVfXAzKdCQKrIkczdPxixgKQ-SA7Ewodn9UmLOjHSCkftB6D9Wj_lVjRZpF9VCanp7thMCeG',
```
ilanbiala: These probably shouldn't be here

modules/articles/server/policies/articles.server.policy.js
```diff
@@ -52,7 +52,7 @@ exports.isAllowed = function (req, res, next) {
   if (req.article && req.user && req.article.user && req.article.user.id === req.user.id) {
     return next();
   }
-
+
```
ilanbiala: spaces

modules/articles/server/routes/articles.server.routes.js
```diff
@@ -4,20 +4,27 @@
  * Module dependencies
  */
 var articlesPolicy = require('../policies/articles.server.policy'),
-  articles = require('../controllers/articles.server.controller');
+  articles = require('../controllers/articles.server.controller'),
+  passport = require('passport'),
```
ilanbiala: modules above files

modules/core/client/app/config.js
```diff
@@ -4,7 +4,7 @@
 var ApplicationConfiguration = (function () {
   // Init module configuration options
   var applicationModuleName = 'mean';
-  var applicationModuleVendorDependencies = ['ngResource', 'ngAnimate', 'ngMessages', 'ui.router', 'ui.bootstrap', 'ui.utils', 'angularFileUpload'];
+  var applicationModuleVendorDependencies = ['ngResource', 'ngAnimate', 'ngMessages', 'ui.router', 'ui.bootstrap', 'ui.utils', 'angularFileUpload', 'LocalStorageModule'];
```
ilanbiala: Is local storage the best place?

modules/users/server/policies/admin.server.policy.js
```diff
@@ -29,7 +29,7 @@ exports.invokeRolesPolicies = function () {
  */
 exports.isAllowed = function (req, res, next) {
   var roles = (req.user) ? req.user.roles : ['guest'];
-
+
```
ilanbiala: space

modules/users/client/services/authentication.client.service.js
```diff
@@ -1,12 +1,81 @@
...
+      readyPromise.resolve(service);
+    }
+
+    function setUser(user) {
+      service.user = user;
+    }
+
+    function setToken(token) {
+      service.token = token;
+      localStorageService.set('token', token);
```
ilanbiala: What does this do differently than just using the browser API?

trainerbill: Well for one it fallsback to cookies if localstorage is not supported.  After writing the tests, I had the same thoughts.  Especially for CSRF you don't want to store in cookies.  Would be willing to remove the localstorage module and just use the browser API if we don't want to support older browsers.  Let me know what you think.

modules/core/client/app/config.js
trainerbill: See below.  We can remove the module, but I am not sure where else we would want to store the token.  The articles I read all said local storage over cookies.

config/env/development.js
trainerbill: lol.  good thing that wasn't production...

modules/users/client/services/authentication.client.service.js
ilanbiala: I think localstorage is just fine, but what about session storage?

trainerbill: We could do that, but the token itself has an expiration as well so why not just keep it around in localStorage for as long as its valid?  We could add a check box for keep me logged in that sets the token expiration for like a year and place it in localStorage, otherwise do X(hour or so) amount of time and place it in sessionStorage.  But I am not sure how you would handle provider auths with that method.  Thoughts?

ilanbiala: The browser API is probably a better choice

modules/core/client/app/config.js
mleanos: +1 on using Angular Local Storage. I believe local storage is the preferred (and more secure) storage container. Most articles I've read have encourage this as well.

ilanbiala: @mleanos no need for a plugin when we can use the browser API which is fine imo, unless there's something that the angular plugin does that we need?

trainerbill: @ilanbiala The only thing it really provided was a fallback to cookie if localStorage was not supported in the browser.  My most recent commit removed the module and is now using browser API

trainerbill:
All.  For the password reset functionality would you mind just signing a JWT token and emailing it to the email address?  currently we are creating a new token and checking it, but with JWT you can now pass in ?token= for the social auth providers.  So essentially we could sign a token and put a link to the password reset page with ?token= and it would log the user in and display the password change form.  It should cut out a lot of code and work as expected.  Just wondered if this would be good with everyone.

jloveland:
@trainerbill awesome work!! My thought below may be orthogonal to your question about sending the token in email for password reset, but wanted to put it out there..

I am not sure emailing a signed token is the best approach. Since email is generally not secure. Although, the current method uses email. Would the token change once they reset their password? Possibly a different feature, but maybe combining temporary link in email with a couple security questions would be best? References:

- http://www.troyhunt.com/2012/05/everything-you-ever-wanted-to-know.html

- https://fishnetsecurity.com/6labs/resource-library/white-paper/best-practices-secure-forgot-password-feature

trainerbill:
@jloveland Thanks!  The current method uses email so that is what I am going off of.  The JWT token has an expiration so for this feature we could just make it valid for like 1-5 minutes.  I don't think adding security questions is necessary.  Would it be nice, yes, but it adds a lot of complexity that is generally handled by an IDMS

codydaig:
@trainerbill @jloveland I'd like to keep complexities like security questions out of the framework right now. In general, if someone gets into your email they pretty much own your life. I don't think we need to protect against that in the framework. I don't like the idea of the token expiring in 1-5 minutes either. Depending on your mail server, it may take longer than that to get the token. How about 12 hours? Can we make it a one time use token?

trainerbill:
@codydaig I will make the jwtSign function be able to accept an object so you can set an expiration.  We could code singleUse: true into the token, but how would I know if the token has been used?  We would need to store it somewhere or have a flag on the user object saying a password reset is in progress or something like that.  Seems a bit much to me.

ilanbiala:
I like single use, but I'm also fine with a short expiration, even something like 30 minutes.

trainerbill:
@ilanbiala For this PR, I am going to do the expiration of 30 minutes.  If we want a single use in the future it can be a separate PR.

codydaig:
:+1:

mleanos:
Why not keep the same Password Reset token functionality that exists now? To me that's separate than the Authentication token. And it wouldn't create any more work to be done in this PR. I think the main goal of this PR should be to implement a new Authentication method. Not refactoring existing, working, code.

mleanos:
Let's keep things simple here, and then after the JWT is merged we can look at enhancing an existing feature to take advantage of it.

trainerbill:
@mleanos I am just not familiar with the current reset process.  This obviously will need to change:

https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js#L147

But since we are changing auth schemes I am not sure what else would need to change for this process.  To me, generating a JWT is easier and waayy less code.  I am unsure how to get the client auth service injected with the new token as I can't test this process.  Does anyone have some SMTP creds or know how to get this working with Gmail?  For another application I had to redo the emailing portion and use nodemailer-smtp-transport to get it to work.  Anyone have success sending emails in the current master using Gmail?

mleanos:
I think you just need to add your credentials to the config with your Gmail account. I believe this worked for me, but I have since implemented AWS SES..
```javascript
mailer: {
    from: process.env.MAILER_FROM || 'youremail@gmail.com',
    options: {
      service: process.env.MAILER_SERVICE_PROVIDER || 'gmail',
      auth: {
        user: process.env.MAILER_EMAIL_ID || 'youremail@gmail.com',
        pass: process.env.MAILER_PASSWORD || 'yourpassword'
      }
    }
  }
```

trainerbill:
@mleanos Yeah I tried that.  Can you see if you can get it working with a clean copy of master?

mleanos:
Worked fine for me, on the latest master.

trainerbill:
Thanks @mleanos.  Got it working.  Reset process now works, though we should change to JWT in the future.

mleanos:
@trainerbill I started working on the SocketIO changes. It might take me a few days to get it sorted out. 

While I was testing, I was able to sign up fine. However, if I logged out, I could not sign in again. I was seeing a 400 response coming back from my signin post, but I didn't look into why.

trainerbill:
@mleanos I am not able to reproduce this.  I can login and log off as many times as I need to.

All.  Tests are fixed.  All I really need is the ChatModule and the new auth service test.

trainerbill:
Everything is now working properly.  All tests pass and coverage increased.  I am hoping @codydaig or @vaucouleur can finish the new auth service client test.  @ilanbiala @lirantal This PR is complete and ready for review/merge.

config/lib/jwtAuthentication.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
mleanos: I think we can rename this file to `authorization`; makes more sense to me. If we wanted to support having multiple Authorization methods we could always create a folder named *authorization* & put individual auth strategies there; then naming this `jwt` would make sense.

modules/users/server/routes/admin.server.routes.js
```diff
@@ -4,22 +4,28 @@
...
+  router.route('/:userId')
     .get(adminPolicy.isAllowed, admin.read)
     .put(adminPolicy.isAllowed, admin.update)
     .delete(adminPolicy.isAllowed, admin.delete);
 
   // Finish by binding the user middleware
-  app.param('userId', admin.userByID);
+  router.param('userId', admin.userByID);
+
+  app.use('/api/admin/users', router);
```
mleanos: @trainerbill Why do we have to add */admin/* here? 

I don't think we need to create a new express Router for each route configuration. We can just use the existing Express `app` and add the `passport.authentication` middleware there. If we keep the functionality that exists in *master* right now, then we won't have to redefine these routes here.

We shouldn't have to explicitly make the request using `/admin/`. What happens when we update the ACL policies to allow a new Role access to these routes? IMO, the routes should always be going through `/api/users` & the ACL's handle the rest.

trainerbill: @mleanos Its a better implementation.  Keeps the routes confined and off the main application to reduce collisions.  For instance if you wanted to do .param(:userid anywhere else it would no longer conflict like it does now.  For instance the admin routes would want to return a full user, while if you are not admin you may want to only return certain fields.  Also allows things like:

router.all('*', passport.authenticate('jwt', { session: false }));

to secure all the routes instead of adding them all individually.  There was a specific reason I had to do this, but I can't remember exactly why.  But its a better implementation regardless.  If you need a new role you just add it to the correct policy....

trendzetter:
I tried testing this but I am probably doing it wrong. 
`git clone -b JWTAuthorization https://github.com/trainerbill/mean.git meanJWT`
But after npm install I cannot run it using gulp:
`meanJWT$ gulp
[23:34:29] Local gulp not found in ~/Bureaublad/mean/meanjs/meanJWT
[23:34:29] Try running: npm install gulp
 `

trainerbill:
@trendzetter just clone the repo then checkout the branch and do npm install

modules/users/server/routes/admin.server.routes.js
mleanos: This is a good discussion on why not to use `/api/admin/users` https://github.com/meanjs/mean/pull/713

After a few long discussions, @rhutchison made me see the light & I feel like we shouldn't separate the User routes between Admin & normal User access. 

If the `/admin/` is removed from the API route & then removed from the client-side service, then Admin feature breaks. Perhaps, that's a hint on the reason for having to add the Express Router implementation here.

I think it adds an extra layer of complexity that doesn't add any value; and it actually forces us to create an API route (*/api/admin/users* that goes against convention.

Can we find a way to avoid these changes? I'm not opposed to a refactoring of the routes, but maybe not in this PR, or finding a way here to keep the same API for both Admins & Users.

trainerbill: Possibly but this is confusing to me:
```js
'use strict';

var passport = require('passport'),
  express = require('express'),
  adminPolicy = require('../policies/admin.server.policy'),
  admin = require('../controllers/admin.server.controller'),
  users = require('../controllers/users.server.controller');

module.exports = function (app) {
  // Create Router
  var router = express.Router();

  //Set JWT Auth for all user Routes
  router.all('*', passport.authenticate('jwt', { session: false }));

  // Setting up the users profile api
  router.route('/me').get(users.me);
  router.route('/')
    .get(adminPolicy.isAllowed, admin.list);
    .put(users.update);
  router.route('/accounts').delete(users.removeOAuthProvider);
  router.route('/password').post(users.changePassword);
  router.route('/picture').post(users.changeProfilePicture);

  //Admin Routes
  router.route('/:userId')
    .get(adminPolicy.isAllowed, admin.read)
    .put(adminPolicy.isAllowed, admin.update)
    .delete(adminPolicy.isAllowed, admin.delete);

  // Finish by binding the user middleware
  router.param('userId', admin.userByID);

  app.use('/api/users', router);
};
```

so api/users PUT a user can access but GET only and admin can.  To me it is bad to mix admin and user routes.  Seems to me like someone could confuse this, or even worse change something that allows non admins to access user data.

trainerbill: Also I really didn't like this:

https://github.com/meanjs/mean/blob/master/modules/users/server/routes/admin.server.routes.js#L11

mleanos:
@trainerbill When logging out, I'm not redirected back home. When on a state that is forbidden to non-authenticated user's, I get sent to the *Forbidden resource* state. Also, when logging out from an Admin view, I'm not redirected & I don't receive the Forbidden Access state, if I had previously done a "hard-refresh" (Ctrl+F5), on any view in the app.

My hunch is that when logging in initially, the User isn't added to the request properly **and/or** the User isn't getting cleared from the request object on logout. Can you investigate?

modules/users/server/routes/admin.server.routes.js
mleanos: I don't understand the first two sentences of your last paragraph above. Can you re-explain?

As for mixing the Admin & User routes, I think it's more of a true RESTful API design. The only restrictions on the routes should be through ACL's, which uses roles. So it's User/Roles & route that determine access; **only** through the ACL policies.

As @rhutchison proposed, I'm more of a fan to add routes for the user's to something like `api/users/account`

trainerbill:
@mleanos I changed signout to go home which fixes both issues.  Auth was clearing but $state.reload() was not working for some reason.  May be a ui router bug.  eitherway its fixed.

modules/users/server/routes/admin.server.routes.js
trainerbill: Idk.  I really like having them on separate routers which will require a separate base route.  For instance to protect all the routes on the admin user router all I have to do is:

router.all('*', adminPolicy.isAllowed);

Now ANY route added to the router is protected and you don't have to worry about messing up.

trainerbill: Here is another use case.  People have been asking about logging.  With this implementation you can log all admin actions easily by doing router.all('*', myloggingfunction); in the user admin router. This would not be available if it was all on the same router.  Separating them onto their own router makes a lot of additional features possible.  Adding an /admin route in my mind is worth it.

trainerbill: @mleanos I think I found the best of both worlds.  If i regex the admin :userId route it will work on the same base route since things like /me and /password won't get caught:

router.route('/:userId([0-9a-fA-F]{24})')

Need a contributor to weigh in.  @codydaig @ilanbiala would you rather see a new base path /api/admin/users or a regex on the route?

zeldar:
I have been integrating this as the commits came out.  Seems to be working for me.  

I'm using Mongoose Discriminators on my User Schema and the authentication is working as expected for the discriminators. 

Great job @trainerbill

0xNacho:
A really nice post explaining the possible vulnerabilities of JWT. 

https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/

Personally, I have no experience with this type of auth method, but I hope the vulnerabilties explained above have been already fixed in passport library.

mleanos:
@0xNacho I've read this post too. It seems to have some valid points.

@trainerbill The main concern I have is the algorithm method. I think `jsonwebtoken` has an option for specifying which algorithm methods are allowed. Does it makes sense to add that to the env config? Also, how would that affect passports JWT strategy?

trainerbill:
@0xNacho @mleanos Things like this will happen... no way around it.  That vulnerability was released in March and fixed that same day by bumping the version.  JWT has a huge following and they fix things faster than SSL.

As for the JWT Alogrith, all the jsonwebtoken options are exposed to the config.  All these options are available:

https://www.npmjs.com/package/jsonwebtoken

defaults to H256

https://github.com/auth0/node-jsonwebtoken/blob/master/index.js#L70

There is no none anymore.  FYI I don't like setting the algorithm, because the library should handle it.  If there is a vulnerability in H256 they will change the default in jsonwebtoken so I will update and think I am safe but unfortuately I'm not because the algo is hardcoded.

modules/users/server/routes/admin.server.routes.js
trainerbill: @ilanbiala @codydaig can you weigh in here and settle this so we can get it merged?

aralroca:
@trainerbill good contribution :+1: . Well, I implemented a similar JWT authentication sending the register token via email to validate this email with Nodemailer for my project. And I have an issue that maybe you know how to fix!  heh The problem is to check the URL from the email (with the token), is necessary that the session was opened. This is an issue when you are check-in the email from another device like a mobile... 

I have problems to clone your branch to check it, maybe I don't have permissions to access. Is this happening or you solved? Thank you.

modules/users/server/routes/admin.server.routes.js
ilanbiala: Why do you need a regex, why does a plain param not work?

trainerbill: Because the admin and user would bind to /api/users.  If you look at the current code base it requires the users first to prevent it as a hack.  Otherwise the admin feature of /:userId matches /api/users/me and /api/users/password.  Regex prevents it.  Otherwise we have separate admin routes /api/admin/users.

codydaig:
@trainerbill This PR is getting huge. It's becoming harder and harder for everyone to track what changes your throwing into here. The biggest thing for me is the router. Changing over to express.Router should be a separate PR. That PR can happen before this one for the reasons that you've already stated above. 

Also, the smaller you can keep the router PR, the faster it will get merged in. Especially if you're able to split that PR into multiple PRs for each module.

zeldar:
@trainerbill  @codydaig I think a good way to move forward would be to identify a target list of what we would like to change. Here is what I came up with: 

1. JWT Authentication
2. Chat  Socket.io + JWT
3. Angular Services that fit JohnPapas Style Guide 
  a. Core
  b. Users
  c. Articles
4. Angular Controllers (View-Models) that fit Style Guide
  a. Core
  b. Users
  c. Articles
5. Express Router
6. Admin Routes

Missing anything? Too much?

lirantal:
@trainerbill great work.
As @codydaig this PR is becoming increasingly huge that any code reviews or tests are bound to be of lower quality.

I suggest we stick to whatever is in this PR and try to merge it.
Can you attend for everyone's questions and let's push for merging it as soon as we're good on quality and comments?

ilanbiala:
@trainerbill this is going to work a lot better if we separate the router PR out and get that merged in first as @codydaig said. That way the two features are in separate commits (good), and the commits are smaller and easier to track (also good). This is not an easy PR to follow..especially when the changes in 33 files aren't just adding in semicolons or some syntax/style that is separate from the actual logic.

trainerbill:
@ilanbiala @codydaig The router changes need to be made.  I agree they should probably be on a separate PR, but splitting it out creating a PR that passes tests with the router fixes and then redoing the JWT pr to pass tests would be a large undertaking when this PR currently does both.  I have to side with @lirantal and say merge it.  That being said if you want to split this PR into 2 separate ones, by all means do it.

ilanbiala:
@trainerbill I cannot review a PR this big. If the other contributors want to, that's fine, but this PR is huge and is bound to have issues that will come up in smaller PRs and issues, so I'd rather do it properly. Often you submit very large PRs, which shows how willing you are to contribute, but we cannot review 40 changed files with serious changes all at once especially when it can be separated properly.

zeldar:
I understand @ilanbiala position, but I have to agree with @trainerbill.  I've merged this into my project already.  I've also started merging the Core Refactor at the same time.  

Unfortunately, I don't use the tests like I should, but I haven't ran into problems so far.  I've already refactored my modules to use the new router.  

I would like to see this merged in so that other PRs can be ready to merge with it.  Please let me know how I can help out to get this PR Merged.  I'm willing to offer assistance in the other PRs coming in to to help ensure they follow this style and the PA Style guide.  

I believe this project needs an E2E style guide for controlling permissions to help reduce confusion.  Certainly a separate PR, but maybe a good companion to this one.

codydaig:
This PR is too big to review. I can do my best, but We're going to have issues down the road because of how clustered it is. In all fairness, I mentioned that this should be separate PRs before the Router refactor started (<24 hours of the PR coming in). I could tell this was going to get big and that it could be broken down into smaller PRs early on.

trainerbill:
I understand that.  So like I said feel free to use this PR and separate it out.  Honestly, I am moving to material design so I am running a fork of MEAN with this PR already in.

lirantal:
@ilanbiala , @codydaig  JWT is very important to move forward especially with addressing mobile applications for MEAN.JS so I don't want to see this PR get dropped.
Can we identify problematic areas that we can perform manual testing with the help of @trainerbill and @zeldar (and ourselves) to make sure everything is covered and then merge this PR?

zeldar:
@codydaig I believe this PR starts to conflict with:
#1202 - Core Style Guide Rewrite
#1208 - Users Style Guide Rewrite 

It seems easier to do all of these together because they essentially rewrite the front end completely.

zeldar:
1. Moved auth-interceptor.client.service.js to Users Module.
2. Changed that file's responseError function with :

```JavaScript
function responseError(rejection) {
      if (!rejection.config.ignoreAuthModule) {
        switch (rejection.status) {
          case 401:
            var auth = $injector.get('Authentication');
            auth.signout();
            $injector.get('$state').transitionTo('authentication.signin');
            break;
          case 403:
            $injector.get('$state').transitionTo('forbidden');
            break;
        }
      }
```
Fixed an issue I was having with circular dependencies when integrating #1202 with this PR.

mleanos:
@zeldar Was it necessary to move the `auth-interceptor` service to the User's in order for your fix to work? Or was just injecting the Authentication service good enough?

zeldar:
@mleanos I would certainly think the service injection was what I was missing. I still think the right place to put the file is in the users module. That doesn't mean this project has to do the same thing, but thank you for pointing out the clarification to others reading.

mleanos:
@trainerbill I just pulled this branch down again, and the Social Account's aren't functioning properly. 

&nbsp; &nbsp; 1) Google, GitHub & Twitter works fine to register a new User, and logs in.
&nbsp; &nbsp; 2) Facebook creates the new User, but doesn't log in.
&nbsp; &nbsp; 3) Each social account tries to create a new User when already logged in. Subsequently, if the User is created, it is logged in & the previously logged in User is logged out.
&nbsp; &nbsp; 4) When creating a local provider account (with username & pwd), if I then try to signup/login/add-provider using a Provider that uses the same email as my local account, it creates a new User with the same email. Any clue on why the `unique` attribute of that field wasn't enforced?

Another issue I noticed, when logged in, is that when doing a *hard-refresh* (CTRL+F5) the User is logged out.

@trainerbill Can you rebase this branch & see if you're experiencing the same issues? I didn't test using LinkedIn or Paypal.

mleanos:
@lirantal I agree that JWT is of the utmost importance. I've been wanting this for quite some time. There was a lot of hard work put into this PR, so I don't want to see it dropped either. However, it still needs some work.

I'm concerned that this PR has gotten too big, and has changes that are unnecessary which is only adding to the amount of effort that is required to review this. Debugging issues are difficult when there are so many differences between it & *master*.

What I don't think is necessary..
&nbsp; &nbsp; 1. The move to using Express.Router in each server-side route config. It's a great feature that I would like to see in the framework. However, the implementation needs some work. I can go into more detail on this later.
&nbsp; &nbsp; 2. The implementation of the [Authentication server here](https://github.com/meanjs/mean/pull/1163/files#diff-90593c1684669b4a0cc185ec424f0a24R16) seems weird. I'm not a fan of how the Authentication client service is implemented now. I mentioned this in the PR that originally introduced these changes; now they're added here.

I feel like this PR was rushed, without proper consideration for the review process. I appreciate all the hard work that was put into it, and how fast the changes were made. However, I feel like the quality of the implementation suffered as a result. 

Can we find a way to trim down this PR, and get a bit more focus? If not, then I'm all for going for a full review (including testing) with others that are interested. It will slow the process down, but still doable.

lirantal:
@mleanos I appreciate the honest words and commitment to quality, and would like to further enhance that we aren't in any rush to press the merge button for any PR what-so-ever. My comment was actually about ensuring quality maintains high by performing manual tests on-top of all of our automated testing.

@trainerbill would you like to re-think submitting a smaller PR to just address the core JWT Auth feature first so we can review that?

mleanos:
@lirantal I didn't take your comment as you were rushing this. It was more in general from the beginning of this PR since it's such a big change,

vaucouleur:
@trainerbill Some JWT services tests there:
https://github.com/vaucouleur/mean/blob/JWTAuthorizationTest/modules/users/tests/client/authentication.client.service.tests.js

It's missing the "Token on URL" tests. I need to discuss those with you when you are available.

trendzetter:
@trainerbill I managed to run the branch. As soon as nodemon has started got this error:
```
/meanTB/node_modules/mongoose/node_modules/mongodb/lib/utils.js:98
    process.nextTick(function() { throw err; });
                                        ^
TypeError: undefined is not a function
    at allFailed (/meanTB/node_modules/passport/lib/middleware/authenticate.js:154:11)
    at attempt (/meanTB/node_modules/passport/lib/middleware/authenticate.js:160:28)
    at JwtStrategy.strategy.fail (/meanTB/node_modules/passport/lib/middleware/authenticate.js:277:9)
    at JwtStrategy.authenticate (/meanTB/node_modules/passport-jwt/lib/strategy.js:114:21)
    at attempt (/meanTB/node_modules/passport/lib/middleware/authenticate.js:341:16)
    at authenticate (/meanTB/node_modules/passport/lib/middleware/authenticate.js:342:7)
    at /meanTB/config/lib/socket.io.js:88:59
..
```
Am I doing it wrong or did I find an issue to resolve?

trainerbill:
@trendzetter Tests still pass so this should not be an issue.  MY guess is that you have a dependency issue.  Pull the branch and make sure you NPM Install

trendzetter:
You are correct. I forgot in all the rush to get this going..

trendzetter:
I tested this and hope to see this merged. It would improve greatly on security and fixes the bug hiding the edit/delete buttons. 
There is proposed to break this pr up in pieces. @lirantal Maybe you could do a suggestion? I would propose to spit the social network accounts into a separate pr because that part still seems to have some bugs while the rest of the functionality is working fully now.

lirantal:
@trainerbill ok, a lot has been going on in this PR so I'd need to free up some time to get it resolved in several PRs.

config/env/default.js
```diff
@@ -37,5 +37,12 @@ module.exports = {
         fileSize: 1*1024*1024 // Max file size in bytes (1 MB)
       }
     }
+  },
```
zargold:
```jwt: {
    secret: process.env.TOKEN_AUTH_SECRET || 'M3@N_R0CK5!',
    options: {  //Anything From https://www.npmjs.com/package/jsonwebtoken
      expiresIn: process.env.TOKEN_EXPIRES || '1d'
    }
  }``` reformatted.

**Comment by zargold** [comment]
screen shot 2016-02-22 at 12 31 11 pm
Seems like even after logging out (read: JWT token is gone) session storage is still there. Not sure if I messed up his copy though so maybe someone should test more thoroughly.

**Comment by trendzetter** [comment]
I am getting forbidden when trying to access articles/create when signed out.

**Comment by trendzetter** [comment]
If I use 2 tabs and log out in one I can continue my session with my token in the other tab as long as I do not refresh the page.

File: config/lib/socket.io.js [hunk 14]
```diff
@@ -85,7 +85,7 @@ module.exports = function (app, db) {
```
zargold: The stuff above here talks about sessionSecret sessionId not sure if that's need/was causing my problem.

trainerbill:
FYI one issue is that signout has to kill the token in the $http header.

0xNacho:
@trainerbill I've just read your comment about CSRF protection. If meanjs no longer use cookies, you don't need to worry about cross requests. In other words, with JWT you don't need CSRF protection.

config/lib/socket.io.js
zargold: I removed the above code (everything about Session and cookies and everything worked just fine!

jameslegue:
@trainerbill Saw your comment above you are working on a fork with JWTAuth and Material design. Any chance you're willing to share that branch? It's exactly what I'm after.

trainerbill:
@James league I do.  Hit me up on gitter

ilanbiala:
Let's rebase this and finalize everything. @mleanos @trainerbill can we work out the rest of the kinks?

avingochea:
Hey guys, I think I've found an issue. req.user is undefined in articles.server.policy.js so it always defaults to 'guest' role.

avingochea:
Fixed it by adding this to the articles.server.routs.js file just before the routes are declared

`  // Set JWT Auth for all article routes
  router.all('*', passport.authenticate('jwt', { session: false }));`

avingochea:
Just realized that the get route intentionally doesn't have authentication. What do you guys think about authenticating if there's a token, but continuing if there isn't. This would leave the policy to decide whether the route should be accessible to guests

mleanos:
@avingochea IMO, we should attempt to authenticate on each request. Thus, the responsibility of API access will solely reside in the server policies.

trainerbill:
That adds to much logic to the policy.  Just attached the passport Middleware to the route that you want to secure like it's designed to do.

lirantal:
@trainerbill this by far adds so much code that I'm incapable of code reviewing that.
We can take a look at this in a new PR if there's an easier approach?

PR #1165, opened 2016-01-27
fix(Travis build): PhantomJS and SocketIO dependencies

trainerbill:
changed PhantomJS and SocketIO dependencies to ~ to stop breaking changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,9 +58,9 @@
     "passport-local": "^1.0.0",
     "passport-paypal-openidconnect": "^0.1.1",
     "passport-twitter": "^1.0.2",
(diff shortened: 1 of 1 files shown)


codydaig:
LGTM

ilanbiala:
LGTM, not sure why those were not using `~` before.

codydaig:
@ilanbiala There seems to be a bunch that were/are like that. These in particular were breaking things so we fixed rapidly.

trainerbill:
Yeah, someone should probably do a dependency audit.  I can't think of any reason why they wouldnt all be ~.  Any breaking changes, ie minor/major upgrades, should be tested before.

PR #1175, opened 2016-01-30
deprecate(grunt): Remove Grunt

codydaig:
This PR serves as the PR to remove grunt, but also as the discussion of what needs to be accomplished before this can get merged in. Is there any functionality still in grunt that is not in gulp? I will add todo items left in comments to this description.

Tasks still to be added to Gulp:
- [x] `grunt coverage` #1466 
- [x] `copy:localConfig` #1191
- [x] `mkdir:upload`
- [ ] `ngAnnotate`
- [x] `mongoose` (unnecessary)
- [x] `node-inspector`


trendzetter:
Isn't it a pity to remove perfectly working code (as far as I used it) while it's replacements have still issues producing a production build?
#1160 

Wouldn't it be sufficient for the time being to add a notice that this code is no longer maintained?

codydaig:
@trendzetter It prevents the framework from getting bloated. Templatecache is not in grunt, it's only in gulp. We've been adding things to gulp lately in favor or grunt. Grunt is still in the 0.4 releases.

mg1075:
Agree with @trendzetter; I would vote for continued grunt support. 

I don't see where gulp covers the `copy:localConfig` task.

https://github.com/meanjs/mean/blob/e14088039748ba515e7de3f5a44ec20965337537/gruntfile.js#L218-L226

mleanos:
There's a bit more work to do to get Gulp in line with the tasks that exist in our Grunt implementation. This PR is a great starting point to get that conversation going, as Cody said. 

Look at all the dependencies we get to remove, when we get rid of Grunt. The more dependencies this framework has, the more likely we'll run into issues with environmental compatibilities. Ive run into quite a bit of issues with Grunt. Not really at all with Gulp. 

Again, like Cody said, the 0.4 releases will have Grunt. Even if someone were to install, or upgrade to 0.5.x, they would be able to add Grunt very easily. It would be a matter of a few copy & pastes. 

@codydaig I'll help out with getting Gulp ready for this to be merged. I know the tasks pretty well, between the two.

ilanbiala:
@codydaig @mleanos I'll help out with the review.

trendzetter:
@codydaig I am very much in favor of lean and mean. I also value continuity and stability as a user of the framework. It appeared to me that when tracking the upstream I could run into issues with producing occasional production builds.

ilanbiala:
@trendzetter what issues have you had with Gulp?

trendzetter:
I am just referring to #1160 
The production build produced by gulp is not usable if I understand correctly. A solution is in the making but not finished.
So far I have been using the grunt scripts to run my project, very recently I started running gulp for development. I can give the production build a try tonight to establish if it breaks currently working functionality or if it only affects "new things".

ilanbiala:
@trendzetter that is something that doesn't even exist in our Grunt setup, so that is not an issue really.

trendzetter:
I not sure what you mean by "that doesn't even exist in our Grunt setup". 
If I run "grunt build" I get a set of minimized files which I use for setting up a demoserver (in preparations of the real production when the application goes live).
I admit I have no clue about buildtools. I will look into what you are saying tonight.

ilanbiala:
@trendzetter nothing in Grunt creates a template cache for Angular in the same way that Gulp does. What you are referring to is minification, which both build tools already do.

trendzetter:
@ilanbiala Sorry for my ignorance. I just tried the minification build and they both run the same at first sight although the grunt dist files are 1kb smaller.

trainerbill:
Thoughts on using:

https://www.npmjs.com/package/gulp-strip-debug

Would be nice to have some debug statements in the code base.  Like on controller instatiation do

console.log('ArticlesController::Init', vm);

Then let gulp strip them on production build.

Thoughts?

ilanbiala:
@trainerbill I don't think we have that in Grunt, so that should be in a different issue.

mleanos:
@codydaig The following tasks can be added to the list

1) [mkdir:upload](https://github.com/meanjs/mean/blob/master/gruntfile.js#L245)
2) [ngAnnotate](https://github.com/meanjs/mean/blob/master/gruntfile.js#L117) - The Build task uses this

Not sure if we need these. Does anyone have any insight?
3) [mongoose](https://github.com/meanjs/mean/blob/master/gruntfile.js#L255)
4) [node-inspector](https://github.com/meanjs/mean/blob/master/gruntfile.js#L165)

codydaig:
thanks @mleanos

rhutchison:
@mleanos where are we using ngAnnotate?

lirantal:
@codydaig can you update the branch from master and also the description of the PR so we can see the progress of what's left to finish this grunt to gulp migration?

mleanos:
@rhutchison I'm not exactly sure to what extent we're relying on ngAnnotate. 

In Grunt, we're using it in the Build task; which is probably supporting the Uglify task. However, in Gulp we're using the ngAnnotate plugin directly in the Uglify task.

My guess, is that we can leave the `ngAnnotate` task out of Gulp. WDYT?

rhutchison:
@mleanos I don't mind adding it, but my point is there are no annotations in our code, where it would be needed.

mleanos:
@rhutchison It doesn't look like we're using it anywhere. But I think the idea is that user's might want to use it for their production/staging builds. So it's nice that the build task already includes it. I don't know a whole lot about the use cases with ngAnnotate, so it's really up to what others think is best.

mleanos:
@codydaig I was looking at the diff here, and it is a bit of a shame that all the Grunt code will go away. I know it could still be recovered by finding it in the older releases. 

However, I'm wondering if could keep the file (renaming it to something like `example.gruntfile.js`) with the whole file commented out. We could just have a note at the top of the file explaining that we've deprecated Grunt, and also provide a list of the dependencies. There may be some user's that have added to their Grunt configuration, or others that might just prefer Grunt. I think it could be a nice convenience to our user's. Just an idea.. WDYT?

ilanbiala:
We can just refer users to 0.4.2, which is tagger as a release so it shouldn't be hard to find.

codydaig:
I agree with @ilanbiala. 

Having the example there will cause confusion, and people will still want it supported. So, referring them back to 0.4 where it was working and support seems like the better option to me.

mleanos:
SGTM

ilanbiala:
I don't think we need to add ngAnnotate, I think we should only have gulp tasks for things that we actually use in the framework rather than trying to build in tasks in anticipation of a user's needs that will just bloat the gulpfile.

rhutchison:
@codydaig you may want to take a look at the other files where grunt is referenced (Dockerfile, README.md, etc).

mg1075:
@rhutchison - Looking at a 0.4.2 project, that might include:
- .travis.yml
- Dockerfile
- README.md
- .bluemix/pipeline.yml
- config/assets/default.js
- config/env/test.js  (in comments)
- package.json

ilanbiala:
@codydaig wanna rebase? I'll work on adding gulp coverage today.

itelo:
ngAnnotate isn't already in uglify task?
```js
// JS minifying task
gulp.task('uglify', function () {
  var assets = _.union(
    defaultAssets.client.js,
    defaultAssets.client.templates
  );

  return gulp.src(assets)
    .pipe(plugins.ngAnnotate())
    .pipe(plugins.uglify({
      mangle: false
    }))
    .pipe(plugins.concat('application.min.js'))
    .pipe(gulp.dest('public/dist'));
});
```

codydaig:
@ilanbiala Sorry about the delay. It has been rebased.

package.json
```diff
@@ -16,7 +16,7 @@
   },
   "scripts": {
     "start": "grunt",
```
ilanbiala: This should be gulp?

zargold:
Another thing is we could remove all the grunt package.json packages.

Wuntenn:
I quite like the idea of having a variety of ways to compile meanjs.

Having the ability to swich between npm, docker, docker-compose, gulp, and grunt, was is especially helpful/welcoming for me coming into this project. I can imagine that someone else who has already invested time in any of these build tools may also feel the same way.

I was able to trouble shoot my stack because I could switch from one to the other (with code to get me started) and compare what was going on. Although they're quite similar (gulp may be slightly better) we can't assume everyone uses Gulp only.

People can always delete the files they don't want/use.

trendzetter:
@Wuntenn I couldn't bring up a good example of why we still need grunt but I too feel like this is a bit overzealous to have it deleted from the trunk already. They say gulp can do the same and has better performance. I think the performance advantage might not stay with gulp forever and I found it more difficult to temporary disable tasks. This project is guided by people preferring code over configuration while I have a hard time getting into that perspective. I think configuration is mostly preferable over code, especially since there is little to no logic involved in defining build tasks, only configuration. More people will be able to use a configuration based approach, such as server admins.
Anyway, i don't see a blocking change for my project so I leave it to the people more 'into it' to do the strong worded discussions.

lirantal:
I am the first person to say (and have said it :))  that we don't have enough developer capacity to maintain both gulp, and grunt, and one of them should be sufficient. For the end user this should also doesn't matter, it's just another package to install that deals with the build system.

With that said, I wouldn't be quick on the trigger of removing `grunt` which already works and serves the community members of mean.js but rather give it a longer 'fade out' period.

mleanos:
I feel similar to what @lirantal expressed. I don't mind taking the strategy of a more gradual deprecation of Grunt. For me, the biggest benefit of removing Grunt will be to get rid of all the dependencies; there's quite a bit of them that's bloating the framework. 

However, the idea that Grunt is working (and our users are taking advantage of the tool-set it provides) is a good reason to **not** remove it at this time. Especially that we are just now getting Gulp fully in-line with what we already have working in Grunt.

Once we're happy with how Gulp is working, then we should probably remove Grunt. However, moving forward I say we don't put any effort into new Grunt development, and hopefully the maintenance will be minor.

mleanos:
@itelo You are right. We're using the ngAnnotate plugin directly in the Uglify task in Gulp. I think the debate was more, or less, if we should add a Gulp task for it.

@ilanbiala I agree we shouldn't add the task. However, should we leave the plugin where it is in the existing Gulp configuration? My vote would be **yes**, since it's already there & apparently working :)

ilanbiala:
@mleanos it works, but we don't use any of it, so I don't feel like it is useful to keep it just because it's dead code in our Gulpfile.

mleanos:
@ilanbiala I don't see it as adding too much overhead, having the package as a dependency & including it in the Uglify task. There are other convenience features of our framework that we don't actively *use*. As is, it's used in a task that's useful for deployment purposes.

With that said, I don't feel strongly about it either way. I've only used ngAnnotate once with a MEANJS project, and it was nice to have it already implemented in the project; I think it was necessary for something I was doing when deploying to Heroku (but I can't really remember why). Even if the plugin wasn't present, it wouldn't have been hard for me to add it. So I'm fine with removing it from the Gulp file. Unless someone else has strong feelings about keeping it, we should add it to the list here in this PR, to remove it.

lirantal:
@meanjs/contributors what say you on the Gulp readyness for ditching out good old Grunt?
cc @trendzetter @itelo @trainerbill

hyperreality:
@lirantal I'm currently adding coverage tests to gulp to tick off the last item on codydaig's list.

package.json
```diff
@@ -16,7 +16,7 @@
   },
   "scripts": {
```
simison: Scripts were already taken care of: https://github.com/meanjs/mean/commit/5a91d7d3a0e7204cd8453129821da58a45d9bd60

lirantal: Yeah I'm gonna close this PR in favor of a new one. Some things have changed.

mleanos:
I'm fine with removing Grunt officially. I had reservations on supporting users that rely on Grunt already. However, as someone pointed out earlier, anyone can look back at previous versions of the framework to see how to integrate Grunt.

mleanos:
I think we can merge this in, after the conflicts are resolved (re-base). Everything but the `ngAnnotate` has been added to Gulp. Although, we do include the Gulp plugin for `ngAnnotate` since it's being used in the minification of the js files. But we don't need a Gulp task dedicated to `ngAnnotate`.

mleanos:
@codydaig Would you like to wrap this one up? It looks like this is the last key issue before we're ready to release 0.5.0.

codydaig:
@mleanos Rebased.

codydaig:
Despite the one build chain failing, the other passes. It also passes when it runs on my fork of mean. 

@mleanos I believe this is ready to be merged in.

lirantal:
Yay, finally! :)

sujeethk:
@codydaig qq -Is there a reason why Grunt and Grunt-Cli are left out in the dependencies?

codydaig:
@sujeethk See #1543. I must have missed that in the rebase. Thanks for pointing it out!

PR #1176, opened 2016-01-31
Added a comment system to the mean starter template

tutley:
Created https://github.com/tutley/mean-replies (a replies module) and integrated it with the mean  Articles module. I named it replies because of a potential conflict with the word "comment". This provides a nested comment system of up to 9 levels. I'm sure this can be greatly improved but I figured this would help.


lirantal:
@tutley thanks for offering your work on extending the framework but we actually want to focus on the internal core framework and improve that, instead of adding more functional modules and having to maintain them.

Would love it if you could help us with the issue/PR queue!

PR #1183, opened 2016-02-02
[WIP] feat (core) UI Router Nested View

trainerbill:
@ilanbiala  @mleanos 

The goal here is to make it easier for frameworks to inject their views.  So there are 3 main views you can use, hero/main/subnav.  Each one is simply a container as the submodule itself will include the classes.  For instance you would need class="container" or container-fluid on the submodule view as we should allow the full page.  This initial PR just does the core.  I will work on articles and users if feedback is positive.  

Edit:  Also it may seem trivial in the core but once users and articles are done you will see the benefit.

Edit2:  forgot to mention one major advantage is that the views become smaller and reusable.


modules/core/client/config/core.client.routes.js
```diff
@@ -15,28 +19,52 @@ angular.module('core').config(['$stateProvider', '$urlRouterProvider',
     $stateProvider
     .state('home', {
       url: '/',
-      templateUrl: 'modules/core/client/views/home.client.view.html'
+      views: {
+        'hero': {
+          templateUrl: 'modules/core/client/views/hero.client.view.html'
+        },
+        'main': {
+          controller: 'HomeController',
```
ilanbiala: no controllerAs?

trainerbill: totally want to, but didn't want to introduce it in this PR...  If I go that route I would need to do it for all the routes which means I have to rewrite almost the entire user client side I think.  Maybe we should introduce that in another PR to get it up to date with our standards?

trainerbill:
@ilanbiala Can I get a SGTM, so I can continue into articles and users?

modules/core/client/config/core.client.routes.js
ilanbiala: @trainerbill I'm pretty sure @rhutchison started that with other PRs, so I don't see a reason why it would be so big in this one.

ilanbiala:
I'd rather not have 4 PRs open which will probably have conflicts and make more work, when we could nail down 1 at a time, but I'm pretty much fine with this. @codydaig @mleanos @lirantal wanna take a look?

modules/core/client/config/core.client.routes.js
mleanos: @trainerbill We're already using controllerAs in the Articles module. https://github.com/meanjs/mean/blob/master/modules/articles/client/config/articles.client.routes.js#L21

Since the merging of the Styleguide implementation, I believe we want to include the styleguide in all PR's going forward, if it makes sense; i.e. the client-side routing/templates are getting modified.

trainerbill: Yeah that makes sense and I will make that change for anything that currently has controllerAs.  I don't want to do it for anything that does not currently have controllerAs otherwise I will need to stylize the entire controller in this PR as well.  Right?

mleanos:
@trainerbill I agree with @ilanbiala .. Let's not rush things too much. We have a couple big PR's already that are making significant changes. 

I'm generally fine with these changes as well, but I haven't tested it or thoroughly reviewed the impact.

trainerbill:
I would love to nail down one PR at a time but everything is so slow.

Do we have any work on stylizing the core and user modules?  

I don't see any reason that this PR could not go in and stylize later.  I will just use controller on things that have not been stylized and controllerAs if they have been.  When we stylize the rest of the modules, the routes file will already be done and all we have to do is add controllerAs:'vm'

ilanbiala:
@trainerbill some modules have already been changed, why would making the style guide changes in the same PR be an issue?

trainerbill:
@ilanbiala If I conform to all of the style changes I will have to rewrite the non stylized controllers in this PR as well.  I don't think I can do controllerAs vm without updating the controllers.  I think it would make sense to do a nested views PR and a stylize PR to make them smaller.  All someone would need to do in the routes folder in the stylize PR is to add the controllerAs syntax.  Or I can do it all in one PR... up to you.  I would actually prefer @rhutchison to do the style PR as he is the master, but I could pull it off as well.  Let me know.

modules/users/client/controllers/authentication.client.controller.js
```diff
@@ -42,6 +43,8 @@ angular.module('users').controller('AuthenticationController', ['$scope', '$stat
         return false;
       }
 
+      console.log($scope);
```
ilanbiala: ?

modules/users/client/controllers/authentication.client.controller.js
```diff
@@ -3,6 +3,7 @@
 angular.module('users').controller('AuthenticationController', ['$scope', '$state', '$http', '$location', '$window', 'Authentication', 'PasswordValidator',
   function ($scope, $state, $http, $location, $window, Authentication, PasswordValidator) {
     $scope.authentication = Authentication;
+    $scope.credentials = {};
```
ilanbiala: ?

modules/users/client/config/users.authentication.client.routes.js
```diff
@@ -0,0 +1,45 @@
...
+          'main@': {
+            templateUrl: 'modules/core/client/views/layouts/fixed.one.column.client.view.html',
+            controller: 'AuthenticationController'
+          }
+        }
+      })
+      .state('authentication.signup', {
+        url: '/signup',
+        views: {
+          'row-1': {
```
ilanbiala: why row-1 and row-2? What are these?

modules/users/client/controllers/authentication.client.controller.js
trainerbill: Had to add it.  otherwise it wasn't binding correctly having split views

modules/users/client/config/users.authentication.client.routes.js
trainerbill: Making layouts.  Look at modules/core/client/views/layouts/fixed.one.column.client.view.html

These are reusable layouts that submodules can target.  So you can target the ui-view from the layout file.  So since ththe signup route has 2 views technically, one for the login form, and one for social auths they are injected into a layout file.

ilanbiala: @trainerbill I took a look at that HTML file, but this seems a little to specific and maybe shouldn't have 4 rows or something like that..

trainerbill: @ilanbiala can you give me an example of what you are talking about?  The abstract routes inject layouts into the main ui-view, then the nested routes inject into the layouts views.  I added 4 rows to make it extensive.  More than likely a submodule would use its own layouts, if they want like 6 columns 4 rows or whatever, but providing some basic layouts shouldn't be a huge deal.  Its just HTML that will probably never need updating.

You really cant have only one view since there are 2.  One for the social providers and one for the login form.  I could send the social providers to the hero view but it wouldn't work for the user module as that requires two columns.

Take a look at the most recent commit with the user module.  You have to inject a new layout since the nav has to be beside the actual forms.  So in this case on the abstract routes I put a layout file in the main view and then inject into the layouts file views.  I like doing it this way because it demonstrates that you can do really whatever you want with the views in the submodules.  The way I envision it is we just support a few basic layout HTML files to use in the core and demonstrate how to use them and then the devs can do whatever they want later on in their own modules.

trainerbill:
All.  This PR should be in a semi workable state.  Can you pull it down and give it a spin?  I want to get some feedback before I start fixing tests.

modules/users/client/config/users.authentication.client.routes.js
ilanbiala: @codydaig @mleanos maybe you guys can think of a better way to do this, but I don't think this is the best way.

lirantal:
This is a big one.
@trainerbill @ilanbiala any important changes here we still want to pursue?

Closing it in the meanwhile.

ilanbiala:
There was still a lot of WIP code and this was touching on a lot of stuff, so I don't think we would want to pursue any of it right away.

lirantal:
Cool thanks Ilan.

PR #1184, opened 2016-02-03
feat(travis): Test on Mac and Linux

codydaig:


.travis.yml
```diff
@@ -4,6 +4,10 @@ node_js:
   - 0.12
   - 4
   - 5
+os:
+  - osx
+  - linux
```
lirantal: any idea what linux varies from centos?

lirantal:
Very cool @codydaig.
I re-ran the builds again as they failed only on Mac, maybe due to networking issues, let's see.

.travis.yml
codydaig: Nope. The online example shows osx and linux. So I just tried centos to see if that worked. I know we've been having a couple issues with centos on the generator side of things so I specifically want centos. I may swap out linux for ubuntu and see how that works. I want to play with these settings for a bit before we merge. Once we merge this in here, I'll also add them to the generator project.

codydaig:
@lirantal Since when did we add CircleCI to the integration tests?

lirantal:
@codydaig I never added Circle CI nor codeship.
Maybe @ilanbiala ?

codydaig:
@lirantal Put it to just linux and centos. As long as these tests pass, and it looks good to you, I'll squash and merge. Then I'll work adding windows and mac. (Windows and mac both failed). I'll investigate what it will take to add those in. 

Codeship is the code coverage and has been there for a while, but CircleCI is new.

lirantal:
Not so sure on Codeship. I enabled code coverage based on coveralls.io.
I think someone had some updates there.

ilanbiala:
The reason it failed on Mac is because nvm isn't installed. I think Travis suggests installing using `brew`, but I'm not sure how we can do OS-specific stuff for pre build.

ilanbiala:
@lirantal @codydaig I think the other CIs was a glitch, but can you add:

```
  - windows
  - macosx
```

ilanbiala:
LGTM. @codydaig merge?

lirantal:
Very cool guys.

PR #1186, opened 2016-02-04
feat (title): Dynamic Title: Improve SEO

itelo:


zeldar:
I think Dynamic Title Tags are a great idea.  I don't see any immediate security concerns.  Someone with more knowledge on Angular's $root will know.

trainerbill:
Did not like it at first, but coming around.  title: needs to go in the data: on the routes as I don't think it is a ui-router property.  Take out the $rootScope and everything in the article controller as well.  If someone wants to do that on their own so be it but shouldn't be part of the framework.  Or you may be able to pull off the title change in the route file getArticle function.  I think this works in that function so once the article is resolved then this.data.title += article.title.  I think that would be better than injecting $rootScope into the controller.

Not sure on having $root in the server layout.  I would probably like the whole thing better as a directive:

http://stackoverflow.com/questions/23813599/set-page-title-using-ui-router

Ultimately I have no say.

modules/articles/client/views/view-article.client.view.html
```diff
@@ -1,4 +1,4 @@
-<section>
+<section ng-init="addArticleTitle()">
```
mleanos: I'd rather see a `init()` method added to the controller, and then have it called initially in the controller. I think we should avoid `ng-init` as some views might be re-used.

modules/core/client/app/init.js
```diff
@@ -36,11 +36,13 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
         }
       }
     }
+
   });
 
   // Record previous state
   $rootScope.$on('$stateChangeSuccess', function (event, toState, toParams, fromState, fromParams) {
     storePreviousState(fromState, fromParams);
+    $rootScope.$state = $state;
```
mleanos: Why do we need this? I'm not saying it's wrong, just would like to have some context.

modules/articles/client/controllers/articles.client.controller.js
```diff
@@ -8,6 +8,11 @@
   ArticlesController.$inject = ['$scope', '$state', 'articleResolve', 'Authentication'];
 
   function ArticlesController($scope, $state, article, Authentication) {
+
+    $scope.addArticleTitle = function () {
```
mleanos: To go inline with our implementation of John Papa's styleguide, can we do something like the following, after the `vm.save` definition?
```javascript
vm.addArticleTitle = addArticleTitle;
```
and after the Article specific methods, add
```javascript
function addArticleTitle()....
```

modules/core/client/app/init.js
itelo: without this,
```html
<title ng-bind="$state.current.data.pageTitle" ><title>
```
ng-bind="$state.current.data.pageTitle" will not work anymore

zeldar:
+1 for using the Style Guide moving forward for consistency.

mleanos:
I like this feature, and think it will add value to the framework. However, I'd rather see an implementation more along the lines of what @trainerbill suggested (along with the SO answer). 

Also, +1 on adding this to the Article's route config, rather than littering the controller with this.

itelo:
I already finished, but now i'm having some problems with jshint.
I do the same as johnpapa do [here](https://github.com/johnpapa/angular-styleguide#style-y034)
```js
ArticleTitle();

function ArticleTitle() {
  if (article.title) {
    $state.current.data.pageTitle += article.title;
  }
}
```
and gulp waring this
```bash
ArticleTitle();
^ Missing 'new' prefix when invoking a constructor.
```

I can't see anything wrong...

mleanos:
@itelo Where is this code? In the route config? I think since the function is title cased. The jshint is probably considering that a constructor.. just rename your function to `articleTitle`

itelo:
@mleanos thanks, now everything is running (:

mleanos:
What about trying this? http://stackoverflow.com/a/23814161/2228227

modules/core/client/app/init.js
mleanos: http://stackoverflow.com/a/23814161/2228227 as @trainerbill suggested would be better, IMO

itelo:
I've tried this before, but I got two problems:

1. the $stateChangeSuccess runs before the articleService receive the article, so i cant put the article title in title
2. the directive always get undifiened

itelo:
@mleanos now, is everything in a directive (:

modules/articles/client/config/articles.client.routes.js
```diff
@@ -38,10 +46,13 @@
         controller: 'ArticlesController',
         controllerAs: 'vm',
         resolve: {
-          articleResolve: getArticle
+          articleResolve: getArticle,
+          pageTitle: ['articleResolve', function (articleResolve) {
+            return 'MEAN.JS - Edit Article ' + articleResolve.title;
```
mleanos: Does this have to have it's own resolve? Can the `pageTitle` be updated when the `getArticle` promise is resolved?

I think Andrew mentioned that in his comment.. something like..

```javascript
function getArticle($stateParams, ArticleService) {
  var articlePromise = ArticleService.get({
  }).$promise;

  articlePromise.then(function (article) {
    this.data.pageTitle += article.title;
  });

  return articlePromise;
}
```

@trainerbill is that along the lines of what you were thinking? And would that work?

modules/articles/client/controllers/articles.client.controller.js
```diff
@@ -8,6 +8,7 @@
   ArticlesController.$inject = ['$scope', '$state', 'articleResolve', 'Authentication'];
 
   function ArticlesController($scope, $state, article, Authentication) {
+
```
mleanos: Remove empty line

modules/articles/client/controllers/articles.client.controller.js
```diff
@@ -48,5 +49,6 @@
         vm.error = res.data.message;
       }
     }
+
```
mleanos: Remove this empty line too, so there's no diff in this file.

modules/core/client/app/init.js
mleanos: Is this still needed? Doesn't the `getWorkingState` function in your directive take care of getting the current state?

modules/core/client/config/core-admin.client.routes.js
```diff
@@ -8,8 +8,13 @@ angular.module('core.admin.routes').config(['$stateProvider',
...
         url: '/admin',
         template: '<ui-view/>',
+        resolve: {
+          pageTitle: function () {
+            return 'MEAN.JS';
+          }
+        },
         data: {
-          roles: ['admin']
+          roles: ['admin'],
```
mleanos: Remove comma. There's a few other places that have unnecessary comma's after the `data.roles` definition.

modules/core/client/directives/title.client.directives.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+angular.module('core').directive('updateTitle', ['$rootScope', '$timeout', '$interpolate', '$state',
```
mleanos: Can you come up with a better name for the directive? Also, whatever name you land on, match the name of the file with it.

modules/articles/client/config/articles.client.routes.js
mleanos: You'll have to put the state into it's own var, since `this` will change..
```javascript
function getArticle($stateParams, ArticleService) {
  var state = this;
  ......
    state.data.pageTitle += article.title;
  .....
```

mleanos:
@itelo This looks much better. Thanks! I left a few comments.

@trainerbill Review?

trainerbill:
I don't like pageTitle in resolve.  Is there a reason it can't be in data?  @mleanos your comment above with the promise is along the lines of what I was thinking.

itelo:
@mleanos @zeldar @trainerbill thank you for the feedback, now i think this looks much better (:

modules/core/client/app/init.js
```diff
@@ -36,6 +36,7 @@ angular.module(ApplicationConfiguration.applicationModuleName).run(function ($ro
         }
       }
     }
+
```
mleanos: Remove empty line

modules/core/client/directives/title.client.directives.js
```diff
@@ -0,0 +1,29 @@
...
+
+          var title = (getWorkingState($state.$current));
+
+          $timeout(function () {
+            element.text(title);
+          }, 0, false);
+
+        };
+
+        function getWorkingState(currentState) {
```
mleanos: Probably should change the name of this function since it returns the `title` now.

modules/users/client/config/users.client.routes.js
```diff
@@ -15,32 +15,50 @@ angular.module('users').config(['$stateProvider',
...
+        templateUrl: 'modules/users/client/views/settings/change-profile-picture.client.view.html',
+        data: {
+          pageTitle: 'MEAN.js - Settings picture'
+        }
       })
       .state('authentication', {
         abstract: true,
         url: '/authentication',
-        templateUrl: 'modules/users/client/views/authentication/authentication.client.view.html'
+        templateUrl: 'modules/users/client/views/authentication/authentication.client.view.html',
```
mleanos: Unnecessary comma

mleanos:
@itelo This looks really good now. Thanks! I pulled this branch down and tested as well, and everything is working beautifully. 

I left a few minor comments. Overall, I like how this turned out.

mleanos:
LGTM! Thanks, again @itelo!

itelo:
@mleanos Done (:

codydaig:
@itelo Overall looks great. If you can update the commit message to match the [commit message guidelines](https://github.com/meanjs/mean/blob/master/CONTRIBUTING.md#commit-message-guidelines) we can get this merged in.

itelo:
@codydaig everything looks great now? (:

codydaig:
Not quite. The actual commit message needs to be changed, not just the title of the PR.

itelo:
@codydaig now?

codydaig:
:+1: Awesome! Thanks! Going to give this a day or two to ensure no objections from the rest of the team, but I don't see an issue with merging.

trainerbill:
@codydaig @itelo 
modules/core/client/directives/title.client.directives.js
Does not meet style requirements

I think the directive title should be a little more unique than title.  pageTitle to go along with the data: object.

Other than that, it is exactly how I saw it in my head.  good work @itelo

trainerbill:
@itelo You should also update the tests:

https://github.com/meanjs/mean/blob/master/modules/articles/tests/client/articles.client.routes.tests.js

Would like to see one passing test to make sure title is there, and one that passes without a title as title should not be required.

Not sure why there are no core route tests anywhere in the project, but the articles routes should suffice.

rhutchison:
In some places, you have "MEAN.js" others "MEAN.JS".  Beside the inconsistency, I do not like that it's hard-coded.  Maybe there can be some sort of template or config value, that would make it easy to change.

itelo:
@rhutchison I totally agree with you, so i did some adjustments (:

rhutchison:
Thanks.  Can you also come up with a different name for the directive?  Maybe [pageTitle](https://github.com/meanjs/mean/pull/1186#issuecomment-182184179) or something else?

itelo:
@rhutchison would you like change the file name too, like @trainerbill said? page-title.client.directive.js

rhutchison:
LGTM - Great job

modules/core/client/directives/page-title.client.directives.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
```
rhutchison: [IIFE](https://github.com/johnpapa/angular-styleguide#iife)

itelo:
Should I put pageTitle directive in john papa sytle code? I have this question due to all core isn't in style code...

trainerbill:
@itelo I would JP stylize it so a PR in the future does not have to.  I would personally rather see pageTitle as it is more unique than title which may collide in a custom submodule.

ilanbiala:
@itelo we're currently converting the code to follow the John Papa style guide, so please do that for this PR as well.

itelo:
@ilanbiala i already do that (:

ilanbiala:
Okay, I just wanted to clarify because I saw some confusion above.

rhutchison:
LGTM

mleanos:
LGTM. I pulled this down, and tested again. Everything is working as expected. Thanks, again @itelo

ilanbiala:
LGTM.

trendzetter:
@itelo I am getting a different error on pages that do not have data in their state definition:
```
Error: currentState.data is undefined
getTitle@http://localhost:3000/modules/core/client/directives/page-title.client.directives.js:30:1
```

trendzetter:
`if (currentState.data && currentState.data.pageTitle) {`
Fixes both errors.

trainerbill:
I am going to get chastised for this, but it is way too complex for what it does.  I pulled this into my fork and tried to write a test and it was just too complex.  So I reworked it

```js
(function() {
  'use strict';

  angular
    .module('core.routes')
    .directive('pageTitle', pageTitle);

  pageTitle.$inject = ['$rootScope', '$timeout', '$interpolate', '$state'];

  function pageTitle($rootScope, $timeout, $interpolate, $state) {
    var directive = {
      retrict: 'A',
      link: link
    };

    function link(scope, element) {

      $rootScope.$on('$stateChangeSuccess', listener);

      function listener(event, toState) {
        if (toState.data && toState.data.pageTitle) {
          element.html(toState.data.pageTitle);
        } else {
          element.html('No Title');
        }
      }
    }

    return directive;
  }
})();
```

It does not do the article title but simplifies the entire directive.  I think this should be in the core and then make some logic in the article module to append the article title... though I don't even think that is necessary.

itelo:
@trendzetter I was thinking about your error:
```bash
Error: text is undefined
$interpolate@http://localhost:3000/lib/angular/angular.js:11739:11
getTitle@http://localhost:3000/modules/core/client/directives/page-title.client.directives.js:32:28
```
this should not happen due to ```if (currentState.data)```, can you specify with an example?

trendzetter:
I get this when I import modules created before the current change:
```
      .state('rulesets.create', {
        url: '/create',
        templateUrl: 'modules/rule-sets/client/views/form-rule-set.client.view.html',
        controller: 'RuleSetsController',
        controllerAs: 'vm',
        resolve: {
          ruleSetResolve: newRuleset
        },
        data: {
          roles: ['user', 'admin']
        }
      })
```

trainerbill:
@codydaig since you merged, what do you think on my comment above?  Up to you.

mleanos:
@trainerbill If I may jump in here.. I was actively involved with this PR. 

I do like how your version is much simpker. Do you see any limitations with doing it your way?

What do you mean by "it does not do the article title"? Also, what should go in core? This directive already reaides in the core module. 

Can you share what the test you wrote for your version looks like?

trainerbill:
@mleanos 

```js
(function() {
  'use strict';

  describe('core.client.directive.page.title.js', function() {
    var $compile,
      $rootScope,
      $state,
      $scope,
      $templateCache;

    beforeEach(module('core'));

    beforeEach(inject(function(_$compile_, _$rootScope_, _$state_, _$templateCache_){
      $compile = _$compile_;
      $rootScope = _$rootScope_;
      $state = _$state_;
      $templateCache = _$templateCache_;
      $scope = $rootScope.$new();
    }));

    it('should inject the state.data.pageTitle into DOM', function() {
      $templateCache.put('modules/core/client/views/core.client.views.home.html', '<div></div>');
      var element = $compile('<div page-title></div>')($rootScope);
      $rootScope.$digest();
      expect(element.html()).to.equal('Welcome');
    });

    it('should inject the core page title constant into DOM if page title doesnt exist on the state', function() {
      $templateCache.put('modules/core/client/views/core.client.views.home.html', '<div></div>');
      $state.get('home').data.pageTitle = undefined;
      $state.transitionTo('home');
      var element = $compile('<div page-title></div>')($rootScope);
      $scope.$digest();

      expect(element.html()).to.equal('Modern MEAN');
    });
  });
})();

```

The complexities of this implementation reside in the fact that he is putting the article title in the title.  My implementation only puts the state.data.pageTitle. I think that this is what needs to be in the core module, and IF you want the article title then you put that in the article module.  In the resolve of the route just change the state.data.pageTitle to concat the article title.

trainerbill:
To take it a bit further I also created a .constants(CORE_CONSTANTS and have a default title instead of 'No Title'.  To me this is just a better implementation.  If someone takes out the articles module they would have no need for half of this directive, so handle all the other jazz in the article module if you want to.

trainerbill:
I don't see any reason to do all the getTitle work in this directive.  It should simply take whats in the state.data.pageTitle and insert it into DOM.  I am fairly confident that you can just update state.data.pageTitle in the route file in the getArticle function when the promise resolves.  Set var resolve = this,  when the article resolves do resolve.data.pageTitle + article.title.  $stateChangeSuccess does not fire until the article promise resolves so the directive would get the update.  However I didn't work on this directly.  I don't even think its necessary to add the article title to the main title, but if its wanted I am fairly confident this proposal is a better one.

mleanos:
I think you're right. However, the current directive does just get the title from the state; unless I'm missing something. Although, it does go about it in a compkex manner. So your version is probably a better way.

trainerbill:
@mleanos True.

PR #1189, opened 2016-02-05
feat(CA Bundle option)

siteshackinc:
Added CA Bundle option for SSL Certs

### DIFF
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -4,7 +4,8 @@ module.exports = {
   secure: {
     ssl: true,
     privateKey: './config/sslcerts/key.pem',
-    certificate: './config/sslcerts/cert.pem'
+    certificate: './config/sslcerts/cert.pem',
+    caBundle: './config/sslcerts/cabundle.crt'
   },
   port: process.env.PORT || 8443,
   // Binding to 127.0.0.1 is safer in production.
--- a/config/lib/socket.io.js
+++ b/config/lib/socket.io.js
@@ -19,9 +19,11 @@ module.exports = function (app, db) {
     // Load SSL key and certificate
     var privateKey = fs.readFileSync(path.resolve(config.secure.privateKey), 'utf8');
     var certificate = fs.readFileSync(path.resolve(config.secure.certificate), 'utf8');
+    var caBundle = fs.readFileSync(path.resolve(config.secure.caBundle), 'utf8');
     var options = {
       key: privateKey,
       cert: certificate,
+      ca: caBundle,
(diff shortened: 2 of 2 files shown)


lirantal:
HI @siteshackinc 

Thanks for contributing to the project.

Why is the CA bundle required? what happens if it's not there? which is the case now.

A couple of comments on my side:
1. Is the file `/config/sslcerts/cabundle.crt` supposed to be created by the ssl generation scripts or should we be adding it as part of the repo?
2. you need to update your commit messages (see the CONTRIBUTING.md file to see how to style them) and also squash your commits to 1

siteshackinc:
If a person buys a Comodo SSL certificate for example, not all browsers will validate the crt, which is why they provide a CABundle.crt or mulple CA's to add to create a CABundle.crt. Adding the CABundle that a Certificate Authority provides (like Comodo), fixes that issue. I bought a Comodo cert, this was required to have it work correctly. So I thought I would share.

codydaig:
I'm not opposed to having it in there as a comment. But, the self-signed certs do not come with a bundle file, and since our instructions and scripts allow for self-assigned, this would break.

siteshackinc:
How would this break? I would love to hear about it. As the cabundle.crt wont break anything its just an added root ca file that accompanies some ssl providers certificates.

siteshackinc:
Tell you what codydaig, create a blank cabundle.crt file. Enable those options in the proper configs. create a self signed cert. Now show me how this breaks...

lirantal:
@siteshackinc what we mean is that if we try to just merge this PR it will break because we are telling the server to look for the `cabundle.crt` file but it doesn't really exist.

This will be the error that the user will see:
```
    Error: ENOENT, no such file or directory 'config/sslcerts/cabundle.crt'
```

jloveland:
What if you add a line in the `generate-ssl-certs.sh` to create an empty `config/sslcerts/cabundle.crt`?
https://github.com/meanjs/mean/blob/master/scripts/generate-ssl-certs.sh#L16

something like `touch ./config/sslcerts/cabundle.crt`

ilanbiala:
@siteshackinc @lirantal let's try to wrap this up. Can we get a consensus on the implementation and do a final review before merging?

ilanbiala:
@siteshackinc can you make the changes @lirantal requested ^^?

lirantal:
@ilanbiala  @siteshackinc my PR for applying this fix is here: https://github.com/meanjs/mean/pull/1342

PR #1191, opened 2016-02-06
feat(gulp): Add the copy:localConfig task to Gulp

mleanos:
Adds a task to copy the example local environment config, and save it as a development environment config.

Releated: #1175

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -4,6 +4,7 @@
  * Module dependencies.
  */
 var _ = require('lodash'),
+  fs = require('fs'),
   defaultAssets = require('./config/assets/default'),
   testAssets = require('./config/assets/test'),
   glob = require('glob'),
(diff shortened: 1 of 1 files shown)


codydaig:
LGTM

rhutchison:
Should copy:localConfig task come before lint? also, I don't think you want to run copy:localConfig in parallel with nodemon/watch? do you?

gulpfile.js
```diff
@@ -202,6 +203,21 @@ gulp.task('imagemin', function () {
     .pipe(gulp.dest('public/dist/img'));
 });
 
+// Copy local development environment config
+gulp.task('copy:localConfig', function () {
```
ilanbiala: The colon syntax is a remnant from Grunt, and I'd rather it not come over to Gulp as I think it's unclear. Can we try to rename these so they make sense? How about `copyLocalConfig`?

mleanos:
@rhutchison I've updated this based on your feedback. Thanks.

Let me know how it looks to you now.

ilanbiala:
That's pretty long, can you make it shorter like `copyLocalEnvConfig`?

mleanos:
@ilanbiala Updated the name. Thanks.

PR #1194, opened 2016-02-08
fix(core): Add "additionalProvidersData" to "safeUserObject"

francisdaigle:
Needed for management of social accounts.

### DIFF
--- a/modules/core/server/controllers/core.server.controller.js
+++ b/modules/core/server/controllers/core.server.controller.js
@@ -18,7 +18,8 @@ exports.renderIndex = function (req, res) {
       profileImageURL: req.user.profileImageURL,
       email: validator.escape(req.user.email),
       lastName: validator.escape(req.user.lastName),
-      firstName: validator.escape(req.user.firstName)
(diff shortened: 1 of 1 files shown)


mleanos:
LGTM. I've tested this, and verify that the social accounts are now showing up correctly.

Thanks @francisdaigle!

codydaig:
@mleanos @francisdaigle Was this a bug? Can someone elaborate slightly more on why this was needed? I don't have an issue with it, just want to vet it before it gets merged. Thanks!!!

mleanos:
@codydaig I hadn't noticed before, but on the latest *master* the social accounts profile view wasn't showing the connected accounts. This is because the `additionalProvidersData` field isn't present on the client-side User object. 

In short, it is a bug that this is fixing.

francisdaigle:
@codydaig: Yep. Came across it while working on our fork. Must have been omitted by mistake. Makes it a fix I would think for the reasons @mleanos cited.

codydaig:
Thanks @mleanos and @francisdaigle for the clarification.

PR #1196, opened 2016-02-10
fix(users): Update assertion for duplicate username/email for newer v…

### LINKED ISSUE #1124: CRUD unit test not passing
Hi there,

I just yesterday cloned the MEANJS repo and installed the latest version of MongoDB on my Windows 10 machine.

Running the `grunt test:server` task throws an AssertionError when it gets to `User CRUD tests should not be able to update own user details with existing username` as well as for an existing email.

In particular:

`Username already exists` was expected
`11000 duplicate key error collection: mean-test.users index: username already exists` was received

Digging around in the code, I found the following line in the `getUniqueErrorMessage` method in `modules/core/server/controllers/errors.server.controller.js` which should indeed fetch the actual field name:

`var fieldName = err.errmsg.substring(err.errmsg.lastIndexOf('.$') + 2, err.errmsg.lastIndexOf('_1'));`

The search for an index of `.$` seems to corroborate the error MongoDB has thrown in the past, e.g.:

`E11000 duplicate key error index: mydb.users.$email_1  dup key: { : null }`

[Source](http://stackoverflow.com/questions/24430220/e11000-duplicate-key-error-index-in-mongodb-mongoose)

Can anyone reproduce this? Could it be that the test case simply isn't up to date with the latest releases of Mongoose/MongoDB? I'm happy to submit a PR if so, although I would suspect it be a breaking change for users still on an older version of Mongo.

Cheers!

rhutchison:
Update assertion for duplicate username/email for newer versions of mongodb.

When running a newer version of mongodb, users are reporting an issue when executing tests because the response from the mongo request has changed.  This change should support new/old versions of the test.

Resolves #1124

### DIFF
--- a/modules/users/tests/server/user.server.routes.tests.js
+++ b/modules/users/tests/server/user.server.routes.tests.js
@@ -742,7 +742,7 @@ describe('User CRUD tests', function () {
               }
 
               // Call the assertion callback
-              userInfoRes.body.message.should.equal('Username already exists');
+              userInfoRes.body.message.toLowerCase().should.containEql('username already exists');
 
               return done();
             });
@@ -794,7 +794,7 @@ describe('User CRUD tests', function () {
               }
 
               // Call the assertion callback
-              userInfoRes.body.message.should.equal('Email already exists');
+              userInfoRes.body.message.toLowerCase().should.containEql('email already exists');
 
               return done();
             });


modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -742,7 +742,7 @@ describe('User CRUD tests', function () {
               }
 
               // Call the assertion callback
-              userInfoRes.body.message.should.equal('Username already exists');
+              userInfoRes.body.message.toLowerCase().should.containEql('username already exists');
```
ilanbiala: LGTM, but can you add a comment saying that this works for x versions of MongoDB.

modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -794,7 +794,7 @@ describe('User CRUD tests', function () {
               }
 
               // Call the assertion callback
-              userInfoRes.body.message.should.equal('Email already exists');
+              userInfoRes.body.message.toLowerCase().should.containEql('email already exists');
```
ilanbiala: Same comment here.

lirantal:
Thanks @rhutchison.

@ilanbiala I think the mongodb version comment is not needed there and actually `toLowerCase` is probably a common method we should employ for the rest of the tests.

The commit is already set correctly so let's just merge.

ilanbiala:
@lirantal I'm pretty sure the message has changed too because it says containEql instead of equal...so the comment is definitely necessary.

lirantal:
@ilanbiala I still don't see why we need the comments there. `containEql` is just a little more forgiving than `eql`, that's all. That test should work just fine with `eql` too I think, and if not because there is more test from the output then still don't see it as an issue.

ilanbiala:
@lirantal because we haven't really ever used containEql, we should specify why it's necessary in this case. Most people when writing new tests probably wouldn't pay attention if they switched MongoDB versions. The comment won't hurt, and it will probably help at least a few people if not us as well to remember why it is that way.

lirantal:
I think a comment there is just nit-picking. So we never used containEql before, big deal. Anyone can just open the shouldjs API docs and read what it does. If we had to comment every new method of some library that we used we'll have 50% ratio of code comments :)

ilanbiala:
@rhutchison can you clarify exactly which MongoDB versions changed this? Mongoose requires 2.1.6, so I'm not sure how you're using a newer version of MongoDB when we don't even have one in the package.json.

rhutchison:
@ilanbiala I am running 3.2.0

>so I'm not sure how you're using a newer version of MongoDB when we don't even have one in the package.json.

I went to https://www.mongodb.org/downloads#production and clicked download.

ilanbiala:
Oh the actual installation changed it, I thought you installed through npm a different version than Mongoose's dependency. In that case I think it's even more important to have the comment. Honestly it won't hurt. @rhutchison can you just add a comment in both places where you made the change saying which versions have which message and that the containEql catches both versions.

rhutchison:
I'm struggling to see the importance of in-line comments for this change.

If someone cares that much, they can read the history, the original issue (#1124), or the manual for shouldjs.

You can merge the PR when you get off your high horse.

ilanbiala:
@rhutchison you may not see it because you'll remember why, but it's important for the thousands of users MEAN.js has and many people will not read the history, the issue, or the manual to know about the MongoDB versions. The comment isn't really for you as much as it is for the rest of the community.

lirantal:
@ilanbiala just let go this time :)
I don't think the in line comments there justify the case at hand.

ilanbiala:
@rhutchison there are also 2 tests that need to be fixed on the frontend I believe.


```
1) Users E2E Tests: Signup Validation Should report Email already exists
   Message:
     Expected '11000 duplicate key error collection: mean-test.users index: email already exists' to be 'Email already exists'.
   Stacktrace:
     Error: Failed expectation
    at [object Object].<anonymous> (/Users/ilan/Desktop/mean/modules/users/tests/e2e/users.e2e.tests.js:275:62)
    at process._tickCallback (node.js:402:9)

  2) Users E2E Tests: Signup Validation Should report Username already exists
   Message:
     Expected '11000 duplicate key error collection: mean-test.users index: username already exists' to be 'Username already exists'.
   Stacktrace:
     Error: Failed expectation
    at [object Object].<anonymous> (/Users/ilan/Desktop/mean/modules/users/tests/e2e/users.e2e.tests.js:294:62)
    at process._tickCallback (node.js:402:9)
```

lirantal:
@ilanbiala how is it possible that the build passes then?

ilanbiala:
These are for when MongoDB 3 is used, which I don't think we are using on Travis.

rhutchison:
> @rhutchison there are also 2 tests that need to be fixed on the frontend I believe.

I'll look it over - thanks.

mleanos:
@ilanbiala I think this has been addressed with #1246

ilanbiala:
Yep, closing.

PR #1197, opened 2016-02-10
fix(users): Correct spelling/grammar mistakes

rhutchison:
Closes #1156

### DIFF
--- a/modules/users/client/services/password-validator.client.service.js
+++ b/modules/users/client/services/password-validator.client.service.js
@@ -11,7 +11,7 @@ angular.module('users').factory('PasswordValidator', ['$window',
         return result;
       },
       getPopoverMsg: function () {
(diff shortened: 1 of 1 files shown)


modules/users/client/services/password-validator.client.service.js
```diff
@@ -11,7 +11,7 @@ angular.module('users').factory('PasswordValidator', ['$window',
         return result;
       },
       getPopoverMsg: function () {
-        var popoverMsg = 'Please enter a passphrase or password with greater than 10 characters, numbers, lowercase, upppercase, and special characters.';
+        var popoverMsg = 'Please enter a passphrase or password with more than 10 characters, numbers, lowercase, uppercase, and special characters.';
```
ilanbiala: I believe it is at least 10 characters, not strictly greater than.

PR #1174, opened 2016-01-30
fix(dependencies)

siteshackinc:
Updated version dependencies, added Kerberos to satisfy a dependency

PR #1198, opened 2016-02-10
feat(core): Update project dependencies

rhutchison:
Closes #1174

update mock socket.io file to pass tests.


rhutchison:
Should we drop node 0.x support?

@codydaig @trainerbill @ilanbiala @lirantal @mleanos @vaucouleur @simison

ilanbiala:
@rhutchison can we keep the socket.io and dependency updates separate?

codydaig:
I'm not opposed to dropping 0.12, especially if we can get builds working on Node 5 before the next release. I agree though that the socket.io stuff should be a separate PR. If we want to drop Node 0.12, that should also be a separate PR.

ilanbiala:
Node 0.12 I think we can drop with 0.5. I'm not sure how we can add support for Node 5, though.

trainerbill:
Its possible.  I have travis building node 5 in a fork.

ilanbiala:
@rhutchison I'm introducing some quick updates as part of #1200 to try to get Node 5 to build properly, so that will probably merge in faster as it has less changes. Just wanted to let you know so that the rebase may be simpler.

rhutchison:
@ilanbiala thanks - I'll clean up this PR later

lirantal:
I re-ran the test but it's still failing so @rhutchison you probably need to update from `master` to get the latest `.travis.yml` config updates

rhutchison:
@lirantal @codydaig @mleanos @trainerbill please review

ilanbiala:
@rhutchison I don't want all the node deprecation and socket change in the same PR, hence #1213.

rhutchison:
@ilanbiala why does it matter?  I made them separate commits which will get you to the same result - just slower.

mleanos:
@rhutchison What is wrong with the Socket.IO tests, that requires you to fix it? Are the tests failing because of the upgrade to Socket.IO?

Can you just fix the tests in this PR, and leave the Node deprecation out of it? There's already an effort to get the project dependencies upgraded as Ilan pointed out. I'm not sure if this was really decided on yet??

Either you want to fix the tests here, or you want to deprecate Node 0.x. 

Also, why remove Node 5 from the allowed failures? Are we confident in it passing already? I'll leave those types of issues to those better suited. However, I think it needs to be coordinated with the others that are already working on it, and not tacked onto a seemingly unrelated fix.

rhutchison:
> @rhutchison What is wrong with the Socket.IO tests, that requires you to fix it? Are the tests failing because of the upgrade to Socket.IO?
Yes

>There's already an effort to get the project dependencies upgraded as Ilan pointed out.
This effort came after this PR.

> Also, why remove Node 5 from the allowed failures? Are we confident in it passing already? 
If you reviewed the travis-ci builds, it's clearly passing.

mleanos:
Yes, Ryan. I saw the builds are passing. My point was more that since there's a lot of activity around the dependencies & the Travis config right now, why not just let things stabilize until we remove it from the allowed. I may be off base here, but I still don't see the point in mixing those changes in with this PR.

rhutchison:
I can add 5 back to the allowed failures, but thought it would make sense to remove it since it's already passing.  Why would we want to allow someone to let it fail again - they should fix it.

ilanbiala:
@rhutchison we also don't really have 100% coverage, so we don't actually know that everything just works in Node 5 as you are assuming.

rhutchison:
@ilanbiala you're correct.

PR #1199, opened 2016-02-10
feat(gulp): Add gulp task mkdir:upload to ensure upload directory exi…

rhutchison:
…sts.

Related #1175

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -218,6 +218,15 @@ gulp.task('copyLocalEnvConfig', function () {
     .pipe(gulp.dest('config/env'));
 });
 
+// Make sure upload directory exists
+gulp.task('makeUploadsDir', function () {
+  return fs.mkdir('modules/users/client/img/profile/uploads', function (err) {
+    if (err && err.code !== 'EEXIST') {
+      console.log(err);
+    }
+  });
+});
+
 // Angular template cache task
 gulp.task('templatecache', function () {
   return gulp.src(defaultAssets.client.views)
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -202,6 +203,11 @@ gulp.task('imagemin', function () {
     .pipe(gulp.dest('public/dist/img'));
 });
 
+// Make sure upload directory exists
+gulp.task('mkdir:upload', function () {
+  return fs.mkdir('modules/users/client/img/profile/uploads', function () { });
```
ilanbiala: @rhutchison need to handle a possible error here.

gulpfile.js
```diff
@@ -202,6 +203,11 @@ gulp.task('imagemin', function () {
     .pipe(gulp.dest('public/dist/img'));
 });
 
+// Make sure upload directory exists
+gulp.task('mkdir:upload', function () {
```
ilanbiala: I think the name doesn't need to be the same, thoughts on maybe `mkUploadDir`?

gulpfile.js
rhutchison: Error is handled with noop

ilanbiala: log it or something maybe? We basically need to notify users of it at the least.

gulpfile.js
codydaig: I hesitate to change since this is what it currently is.

ilanbiala: I know, but Grunt's convention so to use different subtasks, where was can just create a proper task name in Gulp. I feel like we shouldn't keep old conventions.

gulpfile.js
```diff
@@ -218,6 +218,15 @@ gulp.task('copyLocalEnvConfig', function () {
     .pipe(gulp.dest('config/env'));
 });
 
+// Make sure upload directory exists
+gulp.task('makeUploadsDir', function () {
+  return fs.mkdir('modules/users/client/img/profile/uploads', function (res) {
```
ilanbiala: nitpick, res should be err (convention)

ilanbiala:
Also, the scope of the commit message should be gulp, because this is gulp-related, not really "core".

gulpfile.js
rhutchison: Where is this convention defined?

This is a callback method for success/error and is not specifically a callback only for errors.

rhutchison:
@ilanbiala scope is updated to reflect your feedback

gulpfile.js
ilanbiala: It's a common node convention for callbacks to return an error, so err makes more sense.

rhutchison:
@ilanbiala LGTY?

rhutchison:
Closing/Opening to produce clean build.  One node timed out.

Issue #657, opened 2015-07-17
Roadmap planning for MEAN.JS 0.5.0


lirantal:
Features
- ~~0.4.0 - refactor_user #559 (need PR work)~~ added to backlog (mleanos)
- ~~JWT (JSON Web Tokens) authentication method (#1163)~~ added to backlog (mleanos)
- [x] Deprecating Grunt in favor of Gulp (#1175)
- [x] Angular 1.4
- [x] Angular Styleguide - https://github.com/johnpapa/angular-styleguide
  - [X] Articles
  - [x] Core (#1202)
  - [x] Users (#1208)
- [x] #807 [Enhancement] Articles admin @ilanbiala

General Cleanups
- [x] Cleaning up the /client issue #608 ( @trainerbill )

Documentation
- ~~Support deploying MEANJS to Cloud Foundry #681~~ support dropped & feature removed with https://github.com/meanjs/mean/pull/1476 (mleanos)

simison:
https://github.com/johnpapa/angular-styleguide

lukekhamilton:
+1

simison:
Supposing 0.5.0 won't hit the master branch very soon, we could try to include new ng-router. https://github.com/meanjs/mean/issues/530 once its ready/documented.

Using Angular 1.5 would be a good step towards 2.0 some time in the future (0.6.0?)

fatty1380:
> https://github.com/johnpapa/angular-styleguide #490

THIS!!!! Some key highlights that I've adopted in my project:
* "$scope must die!" -  use `controller as` in place of `$scope` [Y030:View Syntax](https://github.com/johnpapa/angular-styleguide#style-y030), [Y031:Controller Syntax](https://github.com/johnpapa/angular-styleguide#style-y031) and [Y032:with vm](https://github.com/johnpapa/angular-styleguide#style-y032]
* IIFE : Immediately Invoked Function Expressions [Style Y010](https://github.com/johnpapa/angular-styleguide#style-y010)
* Standardized and rigorous use of JSHint, JSCS, JSDoc

vaucouleur:
The edit and delete buttons, as well as the "posted on and posted by" at the bottom of every view-x.client.view.html pages are, arguably, a form of code duplication across generated modules.

Custom directives could help factoring out those pieces of code.

- On the Pros side, it would be pushing forward the "angular way" (directives), and help people get a concrete example of one or two simple custom directives.

[...]

PR #1202, opened 2016-02-11
feat(core): Modify core module to implement style guidelines.

rhutchison:
Update the core module to implement the style guidelines.
Reduce size of init.js - moved filter logic out to it's own config.
Rename Menus to menuService

Ref: #657
Ref: #943


modules/core/client/app/init.js
```diff
@@ -1,80 +1,43 @@
...
+    .config(bootstrapConfig);
+
+  function bootstrapConfig($locationProvider, $httpProvider) {
     $locationProvider.html5Mode(true).hashPrefix('!');
 
     $httpProvider.interceptors.push('authInterceptor');
   }
-]);
-
-angular.module(ApplicationConfiguration.applicationModuleName).run(function ($rootScope, $state, Authentication) {
```
rhutchison: Moved to https://github.com/meanjs/mean/pull/1202/files#diff-4f015eca6b0b5f96c11d195d1a2058bc

modules/core/client/app/init.js
```diff
@@ -1,80 +1,43 @@
-'use strict';
+(function (app) {
```
trainerbill: send in document and window and remove app since it is now window.applicationConfiguartion

modules/core/client/app/config.js
```diff
@@ -1,23 +1,29 @@
...
+(function (window) {
+  'use strict';
 
-// Init the application configuration module for AngularJS application
-var ApplicationConfiguration = (function () {
-  // Init module configuration options
-  var applicationModuleName = 'mean';
-  var applicationModuleVendorDependencies = ['ngResource', 'ngAnimate', 'ngMessages', 'ui.router', 'ui.bootstrap', 'ui.utils', 'angularFileUpload'];
+  // Init the application configuration module for AngularJS application
+  var ApplicationConfiguration = (function () {
```
trainerbill: Why another IIFE?  can't you just set window.applicationConfiguration = service?

modules/core/client/app/init.js
```diff
@@ -1,80 +1,43 @@
...
-        } else {
-          $state.go('authentication.signin').then(function () {
-            storePreviousState(toState, toParams);
-          });
-        }
+  
+  bootstrapConfig.$inject = ['$locationProvider', '$httpProvider'];
+
+  //Then define the init function for starting up the application
+  angular.element(document).ready(function () {
```
trainerbill: I would bootstrap document.body instead.  more efficient.

modules/core/client/app/init.js
```diff
@@ -1,80 +1,43 @@
...
+        document.body.scrollLeft = scroll.left;
       }
     }
-  });
 
-  // Record previous state
-  $rootScope.$on('$stateChangeSuccess', function (event, toState, toParams, fromState, fromParams) {
-    storePreviousState(fromState, fromParams);
+    //Then init the app
+    angular.bootstrap(document, [app.applicationModuleName]);
```
trainerbill: document.body

modules/core/client/core.client.module.js
```diff
@@ -1,6 +1,8 @@
-'use strict';
+(function (app) {
```
trainerbill: Why not just send in window and use window.applicationConfiguration.registerModule since that is what you are doing with init.  To me it makes more sense.  If i just open the file i know what window is but have no idea what ApplicationConfiguration is.

trainerbill:
@rhutchison Left some inline comments.  Overall love the PR and is almost spot on with where I was going with it.

modules/core/client/app/config.js
```diff
@@ -1,23 +1,29 @@
-'use strict';
+(function (window) {
```
trainerbill: send in angular to be consistent

modules/core/client/app/init.js
trainerbill: styleize and remove anonymous function

modules/core/client/app/config.js
rhutchison: We don't send in angular anywhere, why here?

modules/core/client/app/init.js
rhutchison: This is a change

modules/core/client/app/config.js
trainerbill: Just a thought since you are now sending window, may as well make it a true IIFE and add angular.

lirantal:
@rhutchison , @trainerbill  once you guys wrap it up, then @codydaig or @ilanbiala can take a look and comment.

zeldar:
@rhutchison Based on a fork by @trainerbill I moved auth-interceptor.service.js to the users module and changed the responseError function to:

```JavaScript
function responseError(rejection) {
      if (!rejection.config.ignoreAuthModule) {
        switch (rejection.status) {
          case 401:
            var auth = $injector.get('Authentication');
            auth.signout();
            $injector.get('$state').transitionTo('authentication.signin');
            break;
          case 403:
            $injector.get('$state').transitionTo('forbidden');
            break;
        }
      }
```

modules/core/client/app/init.js
rhutchison: >Can't the app.applicationModule name just be window.applicationConfiguration.applicationModuleName and same for dependencies?

It can, but I made it shorter.

modules/core/client/app/config.js
rhutchison: I disagree, as it's not part of the style guidelines.  I can remove window if you wish.

rhutchison:
@zeldar that should probably go in a different PR.  This is just to conform to the style-guide.

I think it is ready to go.

modules/core/client/app/config.js
ilanbiala: Why an IIFE inside an IIFE?

ilanbiala: Why the extra IIFE?

modules/core/client/services/interceptors/auth-interceptor.client.service.js
```diff
@@ -0,0 +1,35 @@
...
+            $injector.get('$state').transitionTo('forbidden');
+            break;
+        }
+      }
+      // otherwise, default behaviour
+      return $q.reject(rejection);
+    }
+  }
+})();
+
```
ilanbiala: newline

modules/core/tests/client/header.client.controller.tests.js
```diff
@@ -16,48 +16,45 @@
       $state = _$state_;
       Authentication = _Authentication_;
 
-      HeaderController = $controller('HeaderController', {
+      HeaderController = $controller('HeaderController as vm', {
```
ilanbiala: This controller definition should be changed to be consistent, no?

modules/core/tests/client/home.client.controller.tests.js
```diff
@@ -12,13 +12,9 @@
     beforeEach(inject(function ($controller, $rootScope) {
       scope = $rootScope.$new();
 
-      HomeController = $controller('HomeController', {
+      HomeController = $controller('HomeController as vm', {
```
ilanbiala: Same here

ilanbiala:
Added a few comments.

modules/core/tests/client/header.client.controller.tests.js
rhutchison: What is your recommendation?  This is how we do it for [articles](https://github.com/meanjs/mean/blob/master/modules/articles/tests/client/articles.client.controller.tests.js#L62)

ilanbiala: Oh wait, these are in the tests, sorry I just noticed. Then this looks fine to me and same with the home tests. Only thing is the extra IIFE above that I think can be cut out.

rhutchison:
@ilanbiala took care of double IIFE - just waiting for successful build.

ilanbiala:
@codydaig @lirantal :eyes:?

mg1075:
Please re-consider having _everything_ use `vm` in the html views. 

Yes, using `vm` consistently inside the `controller` function makes sense. 
`var vm = this;` 
Awesome.

However, I do not think it is advisable to have `vm` be the variable inside all the html views. 
`<div class="container" ng-controller="HeaderController as vm">`

There is an important note in John Papa's guidelines about **controllerAs with vm**.
https://github.com/johnpapa/angular-styleguide#style-y032


> Note: When working with larger codebases,**using a more descriptive name can help ease cognitive overhead & searchability** [_emphasis added_]. Avoid overly verbose names that are cumbersome to type.

```
<!-- avoid -->
<input ng-model="customerProductItemVm.text">
```

```
<!-- recommended -->
<input ng-model="productVm.id">
```

rhutchison:
@mg1075 I put forth a [commit](https://github.com/meanjs/mean/commit/84daa57dd4e3b33966a3233cf9438da6164b1ca0) for the community to review.  If they are good with it, I will squash the commit.

@ilanbiala @lirantal @codydaig

rhutchison:
My personal preference would be to stick with 'vm'.

@mleanos 

>I thought we were trying to get away from using the ng-controller attribute in the views.

I don't know if there is a way around it with HeaderController without implementing nested views.  Any ideas?

>The John Papa Styleguide examples aren't very clear on using the descriptive vm naming convention. What's the real benefit of doing this? There's only one controller per view, so it seems that vm would never have conflicts with any other vm.

I believe @mg1075 is saying for searchability in a project.

mleanos:
@rhutchison I haven't tested this, or given it a thorough review. https://github.com/meanjs/mean/pull/1183

rhutchison:
@lirantal @codydaig @ilanbiala do either of you have an opinion on https://github.com/meanjs/mean/pull/1202#issuecomment-184494736

https://github.com/meanjs/mean/commit/84daa57dd4e3b33966a3233cf9438da6164b1ca0

See comments above.

ilanbiala:
When someone defines 2 scopes, both with `$scope.vm = this`, how would it be clear which vm is being referenced?

trainerbill:
I prefer it just as vm.  When using nested views you would have to define multiple variables and remember what to place in each variable.  Using just vm, you can easily just include the view with a different controller and place whatever variables the view needs on vm.  I honestly don't know how that would work with nested views.  I guess normally you would also have a different controller, but there may be times that you would not.

rhutchison:
@ilanbiala the way mean is structured, it's 1 view, 1 controller.  We do not really have the complexity of a larger application.

Even in the enterprise application I am working on, we stick to `controllerAs: 'vm'` and I think it works well - the rule is clear/consistent.

I think descriptive names would be more valuable if we were referencing more than 1 controller scope in a view.

mleanos:
I agree with what @rhutchison is saying. I don't see many use cases where you wouldn't just have a one-to-one relationship between views & controllers.

In the case of using nested-views, I think it's more likely that we would have a view that is used within multiple parent views, that have their own controller. in that case, the controller would be providing smaller sets of data/functionality that the nested-view needs. Correct me if I'm wrong, about that assumption. 

As Ryan mentioned earlier, I don't think we can avoid having `ng-controller` in the Header view.

ilanbiala:
Yeah, if we ng-include the file then we can't really do anything about it I think, we'd have to declare the header as its own ui-view in the layout file.

Meistercoach83:
Why don´t you update to angular 1.5 and use the new component() method? Then you could add your header for example with <mean-header></mean-header>

rhutchison:
@Meistercoach83 or we could just make it a directive

rhutchison:
Is the consensus to revert https://github.com/meanjs/mean/commit/84daa57dd4e3b33966a3233cf9438da6164b1ca0 and squash?  I would like to get this merged.

mleanos:
My vote is Yes; revert the commit. I much prefer using `vm` everywhere, rather than the descriptive naming.

modules/core/client/app/config.js
mleanos: Can we use $window instead?

modules/core/client/app/init.js
```diff
@@ -1,80 +1,45 @@
...
-        }
+  
+  bootstrapConfig.$inject = ['$locationProvider', '$httpProvider'];
+
+  //Then define the init function for starting up the application
+  angular.element(document).ready(init);
+
+  function init() {
+    //Fixing facebook bug with redirect
+    if (window.location.hash && window.location.hash === '#_=_') {
```
mleanos: Same here. Can we use `$window`?

modules/core/client/app/init.js
```diff
@@ -1,80 +1,45 @@
...
-        } else {
-          $state.go('authentication.signin').then(function () {
-            storePreviousState(toState, toParams);
-          });
-        }
+  
+  bootstrapConfig.$inject = ['$locationProvider', '$httpProvider'];
+
+  //Then define the init function for starting up the application
+  angular.element(document).ready(init);
```
mleanos: `$document`?

modules/core/client/app/init.js
ilanbiala: :+1:

modules/core/client/app/init.js
ilanbiala: :+1:

mleanos:
@rhutchison This looks good for the most part. Left a couple line comments. 

Can you rebase? I'd like to pull this down, and test.

mleanos:
@rhutchison I think some of these changes were implemented with fixes, and the Angular 1.5 upgrade. Can you rebase, and verify? Once the remaining issues have been addressed, I think this will be good to go.

rhutchison:
@mleanos I'll try to rebase this evening

modules/chat/client/config/chat.client.config.js
```diff
@@ -5,11 +5,11 @@
     .module('chat')
     .run(menuConfig);
```
mleanos: I think we should rename this file to `chat.client.menus.js`. 

WDYT?

modules/core/client/app/config.js
ilanbiala: $window

modules/core/client/app/init.js
```diff
@@ -1,80 +1,45 @@
...
+    if (window.location.hash && window.location.hash === '#_=_') {
+      if (window.history && history.pushState) {
+        window.history.pushState('', document.title, window.location.pathname);
+      } else {
+        // Prevent scrolling by storing the page's current scroll offset
+        var scroll = {
+          top: document.body.scrollTop,
+          left: document.body.scrollLeft
+        };
+        window.location.hash = '';
```
ilanbiala: $window

modules/core/client/app/init.js
```diff
@@ -1,80 +1,45 @@
...
-          $state.go('forbidden');
-        } else {
-          $state.go('authentication.signin').then(function () {
-            storePreviousState(toState, toParams);
-          });
-        }
+  bootstrapConfig.$inject = ['$locationProvider', '$httpProvider'];
+
+  // Then define the init function for starting up the application
+  angular.element(document).ready(init);
```
ilanbiala: $document

modules/core/client/controllers/home.client.controller.js
```diff
@@ -1,8 +1,17 @@
...
-    // This provides Authentication context.
-    $scope.authentication = Authentication;
+  angular
+    .module('core')
+    .controller('HomeController', HomeController);
+
+  function HomeController() {
+    var vm = this;
+
+    /* implementation details */
```
ilanbiala: Delete this comment, it's unnecessary

modules/core/client/directives/show-errors.client.directives.js
```diff
@@ -1,79 +1,87 @@
...
-      inputName = $interpolate(inputNgEl.attr('name') || '')(scope);
-
-      if (!inputName) {
-        throw new Error('show-errors element has no child input elements with a \'name\' attribute class');
+(function () {
+  'use strict';
+
+  /**
+   * Edits by Ryan Hutchison
+   * Credit: https://github.com/paulyoder/angular-bootstrap-show-errors */
```
ilanbiala: Just leave the HTTP link in a single line comment and the multiline is unnecessary

modules/core/client/services/menu.client.service.js
```diff
@@ -0,0 +1,196 @@
...
+      service.validateMenuExistance(menuId);
+
+      // Return the menu object
+      return service.menus[menuId];
+    }
+
+    function init() {
+      // A private function for rendering decision
+      shouldRender = function (user) {
+        if (!!~this.roles.indexOf('*')) {
```
ilanbiala: Why !!~, shouldn't it just be !!

modules/core/client/services/menu.client.service.js
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    // Remove existing menu object by menu id
+    function removeMenuItem(menuId, menuItemState) {
+      // Validate that the menu exists
+      service.validateMenuExistance(menuId);
+
+      // Search for menu item to remove
+      for (var itemIndex in service.menus[menuId].items) {
+        if (service.menus[menuId].items[itemIndex].state === menuItemState) {
```
ilanbiala: has own property missing

modules/core/client/services/menu.client.service.js
```diff
@@ -0,0 +1,196 @@
...
+      });
+    }
+
+    // Remove existing menu object by menu id
+    function removeMenu(menuId) {
+      // Validate that the menu exists
+      service.validateMenuExistance(menuId);
+
+      // Return the menu object
+      delete service.menus[menuId];
```
ilanbiala: delete typically isn't used

ilanbiala:
@rhutchison left some comments.

modules/core/client/app/init.js
rhutchison: Is `angular.element($document).ready(init);` even valid?

ilanbiala: Yes, but $document.ready also works I believe.

codydaig:
I'd like to use $window and $document everywhere that it can be. Is there any case where we have to use window or document vs. $window and $document?

modules/core/client/app/init.js
rhutchison: init.js:22 Uncaught ReferenceError: $document is not defined

modules/core/client/app/init.js
rhutchison: this will be the same.

modules/core/client/services/menu.client.service.js
rhutchison: Why not?

modules/core/client/app/config.js
rhutchison: I don't think it can be, unless angular is injecting it into a controller/service.

rhutchison:
@ilanbiala @codydaig @mleanos @lirantal this is ready.

modules/core/client/services/menu.client.service.js
rhutchison: because not found for indexOf is -1, which needs to be converted to 0 to evaluate correctly.

modules/core/client/app/init.js
ilanbiala: did you inject it?

modules/core/client/app/init.js
ilanbiala: did you inject?

modules/core/client/services/menu.client.service.js
ilanbiala: Why not just `this.roles.indexOf('*') !== -1`

modules/core/client/services/socket.io.client.service.js
```diff
@@ -1,40 +1,57 @@
...
+    }
 
     // Wrap the Socket.io 'on' method
-    this.on = function (eventName, callback) {
-      if (this.socket) {
-        this.socket.on(eventName, function (data) {
+    function on(eventName, callback) {
+      if (service.socket) {
+        service.socket.on(eventName, function (data) {
           $timeout(function () {
```
ilanbiala: this can just be `$timeout(callback(data));` no?

modules/core/client/services/menu.client.service.js
ilanbiala: Actually this is fine, though it would probably be better if we did `service.menus[menuId] = undefined` because most browsers can more easily optimize that than an object that changes. (https://www.smashingmagazine.com/2012/11/writing-fast-memory-efficient-javascript/) That's a bit old, but written by a software engineer at Google who works on Chrome, so it's probably pretty accurate for most JavaScript engines.

ilanbiala:
@rhutchison left an explanation of delete comment, but other than that the only thing I'm thinking of is the $window and $document stuff that @codydaig is also referencing. @rhutchison did you find any sources that say it isn't possible to use them in scenarios like ours?

rhutchison:
@ilanbiala within the context of angular, I'm sure it's possible, but not sure how it's done in these cases which are bootstrapping angular.  Please let me know if you can find a way and I'll gladly update.

modules/core/client/app/init.js
ilanbiala: @rhutchison did you try injecting?

modules/core/client/app/init.js
ilanbiala: @rhutchison inject

modules/core/client/app/init.js
```diff
@@ -1,80 +1,45 @@
...
-    // only store this state if it shouldn't be ignored
-    if (!state.data || !state.data.ignoreState) {
-      $state.previous = {
-        state: state,
-        params: params,
-        href: $state.href(state, params)
-      };
-    }
+    // Then init the app
+    angular.bootstrap(document, [app.applicationModuleName]);
```
ilanbiala: inject

modules/core/client/services/menu.client.service.js
ilanbiala: @rhutchison have you tried this?

modules/core/client/services/menu.client.service.js
ilanbiala: @rhutchison can you change this?

modules/core/client/services/socket.io.client.service.js
ilanbiala: @rhutchison did this work?

modules/core/client/app/init.js
rhutchison: https://docs.angularjs.org/api/ng/function/angular.bootstrap

modules/core/client/app/init.js
rhutchison: Have you tried it yourself with any success or does this just make sense in your head?

modules/core/client/services/socket.io.client.service.js
rhutchison: I did not try, but it does not look like it would work.

modules/core/client/app/init.js
rhutchison: I have tried different ways, I don't think it's possible.  I have not read or seen any examples that shows otherwise.

modules/core/client/services/menu.client.service.js
rhutchison: it will work, but not sure what this change has to do with angular style guide?  this is not throwing any exceptions, and this is not logic that I introduced. https://github.com/meanjs/mean/blob/master/modules/core/client/services/menus.client.service.js#L14

modules/core/client/services/menu.client.service.js
rhutchison: Not sure why it matters - the method is referenced no where in code.  Additionally, if it's used, it's not going to be used to the point where performance is an issue.  Also, if someone was to iterate over service.menus, then it would be faster if the key was not there.

modules/core/client/services/menu.client.service.js
mleanos: I would prefer using `this.roles.indexOf('*') !== -1` as well. I think it's easier to tell what exactly is going on. However, I agree with @rhutchison about updating this line being out of scope for this PR; as long as it passes the ESLint rules it's not a problem. 

Ryan, if this change can be done here that's great, but if not then no big deal.

mleanos:
I was out of town over the weekend, so I apologize for the lack of feedback. I'm planning on pulling this down in the next day or so.

modules/core/client/app/init.js
mleanos: @ilanbiala @rhutchison If it's not possible (or easily possible) to use `$document` & `$window` in this case, then I say it's fine to leave it the way it is.

For the sake of not holding up this PR we can just focus on the style-guide implementation. However, @rhutchison if we're using `window` and/or `document` anywhere else let's change it to use `$window`/`$document` if it's an easy change. I haven't thoroughly checked for any other other usages.

mleanos:
@rhutchison I just pulled this down, and tested. Overall, it looks pretty good.

I think there are a few remaining comments that need to be addressed/finalized. Other than that this looks ready to me.

modules/core/client/app/init.js
simison: Inject before Angular is even bootstrapped? ;-) Don't think so.

rhutchison:
@mleanos thanks for reviewing.  Please point me to the outstanding items.

codydaig:
@rhutchison Looking at it closer, I see the aspects where you need to use window vs. $window. I'm cool with the way that it is currently. (I don't think I saw window or document being used inside a controller or service).

modules/core/client/services/menu.client.service.js
ilanbiala: @rhutchison I agree, but while we're refactoring for the style guide, we should refactor to make the overall style better. If the code works, then it would be great if you could replace it.

ilanbiala:
LGTM, only change is the !== -1. Other than that, everything is fine.

modules/core/client/services/menu.client.service.js
rhutchison: replaced

rhutchison:
@ilanbiala done

ilanbiala:
@mleanos LGTM.

codydaig:
LGTM. @mleanos Merge before merge conflicts arise!!!!

PR #1203, opened 2016-02-11
feat(articles): Rename list.articles.client.controller.js to list-art…

### LINKED ISSUE #1192: Inconsistent file naming convention for certain files in articles module
If you look at the filenames for the project, they are for the most part very well organized and consistent. There are, however, a few minor exceptions that should probably be fixed in the `articles` module. In most cases, files are named like:

```
edit-user.client.view.html
list-users.client.view.html
view-user.client.view.html
edit-profile.client.controller.js
list-graphs.client.view.html
```

So the format seems to roughly follow the pattern: [hyphenated subject].[client/server].[type].[optional "tests"].js. But then there's:

```
list.graphs.client.controller.js
list.graphs.client.controller.tests.js
```

This may be the only example (swapping a `-` for a `.`), but it probably makes sense for someone more knowledgeable about the codebase to do a quick review of the filenames to make sure they're consistent.

Does this really matter that much? Probably not. But it's an easy enough fix to agree upon and implement.

rhutchison:
…icles.client.controller.js

Closes #1192


lirantal:
LGTM

PR #1215, opened 2016-02-14
feat(gulp): Add mongoose task to Gulp

itelo:


ilanbiala:
@itelo why is this necessary in Gulp?

itelo:
@ilanbiala due to #1175, but i read more and dont know if we need mongoose task...

mleanos:
I don't think we need the `mongoose` task. I know it's on the list in the issue tracking the Gulp migration, but I don't see any use/reason-for-use. We're not even using it with Grunt; the task is only defined, but never used.

itelo:
So, should i close the PR ?

ilanbiala:
We'll reopen if necessary, thanks for the effort though.

PR #1230, opened 2016-02-21
fix(core): Remove the <base> tag.

### LINKED ISSUE #1224: <base href="/"> breaks SVG IRIs and element CSS URLs
Unfortunately, the `<base href="/">` tag in `modules/core/server/views/layout.server.view.html` breaks SVG IRIs and element CSS URLs.

For those very reasonably unsure what I'm referring to:

- [SVG IRIs](https://www.w3.org/TR/SVG/linking.html#IRIandURI) are the formal name for SVG attribute values like `url(#myElement)`, used throughout SVGs for referencing other elements usually defined in `<defs>` tags. They are the only way to do a large number of things in SVGs including filters/manipulations, text paths, marker-start/-mid/-end (the way to make lines into arrows). The [SVG spec 25 different SVG elements](https://www.w3.org/TR/SVG/linking.html#processingIRI) that can use IRIs--many of which only work with IRIs.
- Similarly, element [CSS URLs](https://developer.mozilla.org/en-US/docs/Applying_SVG_effects_to_HTML_content) are commonly used for a number SVG-like effects on regular HTML elements, like `mask`, `clip-path`, and `filter`. These only work with relative URLs.

The problem arises because, as mentioned in [this article](http://webdesign.tutsplus.com/articles/quick-tip-set-relative-urls-with-the-base-tag--cms-21399), `<base href="/">` will apply to *all* URL/IRIs in the document, including SVG IRIs and CSS URLs, which basically always reference a relative element `id` on the same page.

Take [this jsfiddle](https://jsfiddle.net/dofxce2y/). Note the happy little triangle making the `<polyline>` into an arrow. Now look at [the same fiddle with `<base href="/">`](https://jsfiddle.net/dofxce2y/1/). The arrow is no longer there because the IRI is now pointing incorrectly to `https://jsfiddle.net/#Triangle`. You can tell by hovering over the blue-linked `url("#Triangle")` style value in a Chrome style inspector on the `<polyline>`.

Unfortunately, there's only one way to override the `<base>` tag, and it only works in some situations and is pretty unpleasant. [This jsfiddle](https://jsfiddle.net/dofxce2y/6/) demonstrates that you can instead use an absolute URL in the IRI, since `<base>` tags only apply to relative URLs. Unfortunately, this has its downsides:

1. This forces you to add logic in your template to insert the absolute path to the current page into the URL, which is annoying if the URL is dynamic, and frankly, kinda ugly. It's also very fragile, breaking if the page's URL changes, like when you click "Run" on [the same jsfiddle](https://jsfiddle.net/dofxce2y/6/), which just changes the URL. And it only gets more complicated if you have different URLs for the same page (e.g. vanity URLs).
2. This workaround doesn't work at all if you don't have the ability to insert the absolute path into the SVG IRI or CSS URL (perhaps they are generated by a third-party library, or are simply third-party content in the case of SVG artwork).
4. This workaround also doesn't work at all if you have multiple elements with the same `id` in separate `<svg>` elements on the same page and want to refer to them. For example, there is simply no way to make [this jsfiddle](https://jsfiddle.net/dofxce2y/4/) work with a `<base>` tag and absolute URLs, although it's clearly a valid use (since `<svg>` elements kind of have their own documents, so the two `id`s aren't actually colliding).

So what can we do? Well, in [this StackOverflow question](http://stackoverflow.com/questions/12303241/how-to-override-base-tag-without-removing-the-tag-itself), a user asks how to override `<base>`, to which one user responded (back in *2012*):

> The `base` tag is normally not needed these days. It is better to use server-side technologies that let you construct addresses from one or more base addresses. So quite possibly the best approach is to get rid of the tag.

It took me a good 30 mins of probing to get to the bottom of this, and I'd consider myself an expert with JS/HTML/CSS. For less experienced engineers, this could be a real tough bug to track down. And with the rise of visualization libraries like D3, SVG bugs like this are only going to come up more and more often. Since all of the page resources are controlled by a server-side template anyway, I think we should strongly consider prioritizing correctness over convenience and remove the `<base>` tag entirely, switching to fully-qualified path names for external resources.

Sorry for the essay :sweat_smile: Thoughts?

0x24a537r9:
Fixes #1224. Everything looks good in the browser and server/e2e tests, but unfortunately the unit tests are broken, and I'll probably need some guidance from someone more familiar with Angular/Karma.

### DIFF
--- a/modules/articles/client/config/articles.client.routes.js
+++ b/modules/articles/client/config/articles.client.routes.js
@@ -16,7 +16,7 @@
       })
       .state('articles.list', {
         url: '',
-        templateUrl: 'modules/articles/client/views/list-articles.client.view.html',
+        templateUrl: '/modules/articles/client/views/list-articles.client.view.html',
         controller: 'ArticlesListController',
         controllerAs: 'vm',
         data: {
@@ -25,7 +25,7 @@
       })
       .state('articles.create', {
         url: '/create',
-        templateUrl: 'modules/articles/client/views/form-article.client.view.html',
+        templateUrl: '/modules/articles/client/views/form-article.client.view.html',
         controller: 'ArticlesController',
         controllerAs: 'vm',
         resolve: {
@@ -38,7 +38,7 @@
       })
       .state('articles.edit', {
         url: '/:articleId/edit',
-        templateUrl: 'modules/articles/client/views/form-article.client.view.html',
+        templateUrl: '/modules/articles/client/views/form-article.client.view.html',
         controller: 'ArticlesController',
         controllerAs: 'vm',
         resolve: {
@@ -51,7 +51,7 @@
       })
       .state('articles.view', {
         url: '/:articleId',
-        templateUrl: 'modules/articles/client/views/view-article.client.view.html',
+        templateUrl: '/modules/articles/client/views/view-article.client.view.html',
         controller: 'ArticlesController',
         controllerAs: 'vm',
         resolve: {
--- a/modules/articles/client/services/articles.client.service.js
+++ b/modules/articles/client/services/articles.client.service.js
@@ -8,7 +8,7 @@
   ArticlesService.$inject = ['$resource'];
 
   function ArticlesService($resource) {
-    return $resource('api/articles/:articleId', {
+    return $resource('/api/articles/:articleId', {
       articleId: '@_id'
     }, {
       update: {
--- a/modules/articles/tests/client/articles.client.controller.tests.js
+++ b/modules/articles/tests/client/articles.client.controller.tests.js
@@ -83,7 +83,7 @@
 
       it('should send a POST request with the form input values and then locate to new object URL', inject(function (ArticlesService) {
         // Set POST response
-        $httpBackend.expectPOST('api/articles', sampleArticlePostData).respond(mockArticle);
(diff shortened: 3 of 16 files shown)


lirantal:
@meanjs/contributors any say?

ilanbiala:
I remember using SVG and Angular, but I don't think I had to remove the base tag. Sadly I can't access that code anymore, so I can't tell you for sure how I solved that issue. @0x24a537r9 did you find an authoritative Angular contributor/issue saying that the base tag cannot be used?

0x24a537r9:
@ilanbiala please see my explanations in #1224 for why the base tag will break specific SVG features (not all SVGs or Angular). I provide a number of sources detailing exactly when and why the `<base>` tag is problematic.

Unfortunately due to wedding planning and performance assessments at work, I haven't had any time to continue looking into the unit test issue I was running into described in #1224. That said, I really wasn't making any progress anyway, so this might need to be forked and taken up by someone more familiar with Angular testing with Karma.

lirantal:
@0x24a537r9 can you take another look at this PR?

lirantal:
We can really save this PR if someone can help fix the Karma tests...
cc @0x24a537r9 @seanhussey @hyperreality @mleanos

PR #1207, opened 2016-02-11
feat(config): upgrade to angular 1.5

trendzetter:
Upgrade dependencies to more recent versions.
Migration guide: https://docs.angularjs.org/guide/migration

PR #1232, opened 2016-02-25
PhantomJs to 2.x series

usta:
This patch updates PhantomJs to 2.x  and karma-phantom-launcher.
This also shows that an existing but missed by old phantomJs version. After this patch implemented
we need to fix chat module testes but that one is for other PR. (I'm not good at writing tests )

Also this patch is a prerequest for angular1.4 or 1.5 patches


ilanbiala:
Can't merge in this PR since it fails. Need the tests in the same PR.

usta:
so this gives us 2 options:
1- remove all  : meanjs/mean/modules/chat/tests/client/chat.client.controller.tests.js
and mean/modules/core/tests/client/socket.io.client.service.tests.js files until someone rewrite them properly
2- exclude this 2 tests from karma  until someone rewrite them

ilanbiala:
@usta these tests should be fixed in this PR, removing tests doesn't make sense as that regresses our coverage, and excluding them does the same. I don't see how rewriting 2 tests is a huge issue to upgrade to Angular 1.5? Also, why is this not in the Angular 1.5 PR? All of this is related..so it should be in 1 PR, but in separate commits.

trendzetter:
@ilanbiala Concerning the failing tests. I am getting
```
	TypeError: Attempted to assign to readonly property. (line 9)
	io@/modules/core/tests/client/socket.io.client.service.tests.js:9:9
```
I do not understand the issue fully but I found this discussion describing similar behavior:
https://stackoverflow.com/questions/26288963/weird-behaviour-with-use-strict-and-read-only-properties

ilanbiala:
Let's close this PR and keep all discussion on #1207.

lirantal:
@ilanbiala why do you tie this request to upgrade phantom and update whatever tests needed with the angular 1.5 upgrade? are they related?

ilanbiala:
Yes, it all has to do with Angular 1.5 being upgraded.

lirantal:
Sorry I didn't follow-up yet on that PR you have there for 1.5, can you briefly explain where/how it is related?

From this thread it seemed that upgrading phantom requires only to address the tests that fail and re-write them (https://github.com/meanjs/mean/pull/1232#issuecomment-188937951)

ilanbiala:
When you update Angular, some tests fail, and PhantomJS adds some of the methods necessary for the failing tests, but updating PhantomJS causes a couple more to fail, so a few tests need to be fixed.

PR #1236, opened 2016-02-26
fix(users) error objects were not handled

lirantal:


ilanbiala:
@lirantal can we get a test so we know these are necessary in the future. Also, same commit message issue.

lirantal:
@ilanbiala I edited the commit message, thanks.
On the tests - I only added an evaluation of the err object that gets returned from the db, how do you suggest we test it exactly?

lirantal:
@ilanbiala actually, I believe that this test case already asserts that behavior:
https://github.com/meanjs/mean/blob/7852144fc1f71a57a32cfc77afbfbb9670bdbf28/modules/users/tests/server/user.server.routes.tests.js#L274

ilanbiala:
SGTM, merge.

modules/users/server/controllers/users/users.password.server.controller.js
```diff
@@ -32,7 +32,7 @@ exports.forgot = function (req, res, next) {
         User.findOne({
           username: req.body.username.toLowerCase()
         }, '-salt -password', function (err, user) {
-          if (!user) {
+          if (err || !user) {
```
mleanos: Shouldn't we be handling an `err` differently than when we don't find a User?

Our handling of Mongoose errors seems to be a pain point in our coverage. It's difficult (and I don't think we have done it anywhere) to write tests against possible Mongoose errors; other than predictable schema validation errors. So when we write something like `if (err) {...}` when no validation errors are possible (as in this case), we lose coverage.

Maybe it's something we can work on. Although, I'm not sure the exact strategy we should employ. I just feel like when we don't find a User, and when Mongoose throws an unexpected error, are two different cases that should be handled independently.

WDYT?

mleanos:
LGTM. My comment is more of a general concern over our coverage. I think this fix here does the job of fixing the bug. However, I wanted to open up that discussion.

modules/users/server/controllers/users/users.password.server.controller.js
ilanbiala: Yeah, Mongoose coverage is definitely an issue.

codydaig:
LGTM. I do agree that err and !user should be handled differently, but I'm indifferent if that should be a part of this PR or not.

ilanbiala:
I'd rather that be in this PR than another.

lirantal:
We can take that in another PR if any of you has an idea how to add coverage for that.
Because we *are* handling those errors, but rather we don't cover these cases in our tests.

Issue #1032, opened 2015-10-29
Connected Accounts does not account for the initial way you signed up


ervine13:
When viewing your profile and see what social accounts you  have connected, it only shows you the social accounts you connected *after* you signed up already.

Meaning, if I originally signed up using Facebook, then connect G+, the view displays that I have G+ connected. I think it should also show Facebook.

```javascript
// Check if there are additional accounts
    $scope.hasConnectedAdditionalSocialAccounts = function (provider) {
      for (var i in $scope.user.additionalProvidersData) {
        return true;
      }

      return false;
    };
```

lirantal:
@ervine13 this is using the latest `master` branch?

ervine13:
@lirantal yes.

lirantal:
@ervine13 so it's always showing the last social account?
What happens if you sign-up with FB, then G+, then LI? what will it show?

ervine13:
@lirantal I'm sorry I wan't clear. It will only show the subsequent social accounts, not the original.

Scenario 1
1. Sign up with Fb
2. Connect G+
3. Connect Twitter
*I only see G+ and Twitter as my connected accounts*


Scenario 2
1. Sign up with G+
2. Connect GH
3. Connect LinkedIn
*I only see GH and LinkedIn as my connected accounts*

Maybe the question is was this intentional that you only see the subsequent providers and not the original? What I was going to try is check the `user.provider`, not just `user.additionalProvidersData`


```
    // Check if there are additional accounts
    $scope.hasConnectedAdditionalSocialAccounts = function (provider) {
      for (var i in $scope.user.additionalProvidersData) {
        return true;
      }
      //check original
      if ($scope.user.provider === provider) {
        return true;
      }

      return false;
    };
```

mleanos:
@ervine13 I'm going to look into this. One quick question, in Senario 1 are you still logged into the FB provider & then in Scenario 2, you're still logged in with G+? 

I'm wondering if it's just not listing the provider that you're logged in with. IIRC, the listed providers are there for removal. It might not allow you to remove the one you're logged in with; or the fact that it may be considered the *primary*. 

Most of that is speculation, but I wanted to throw that out there as I dive into this.

ervine13:
@mleanos  just tested the following

### Test 1
1. Sign up w FB
2. Connect with G+
3. Go to manage social accounts and only see G+
4. Sign out
5. Sign in with G+
6. Go to manage social accounts and only G+ still

### Test 2
[...]

PR #1241, opened 2016-02-29
fix(users): Missing primary social provider

mleanos:
Adds the User's provider to the list of connect social accounts, when it is also a social account.

This is the first step in solving the issue described in #1032. This at least fixes the bug, of the User's provider not showing up.

We could either merge this to satisfy the bug, and then proceed with the rest of the implementation discussed in the issue. Or we could add it to this PR. We'd still need to discuss, and come to a conclusion of how to handle the different scenarios that end User's can face.

### DIFF
--- a/modules/users/client/views/settings/manage-social-accounts.client.view.html
+++ b/modules/users/client/views/settings/manage-social-accounts.client.view.html
@@ -1,6 +1,11 @@
 <section class="row" ng-controller="SocialAccountsController">
   <h3 class="col-md-12 text-center" ng-show="hasConnectedAdditionalSocialAccounts()">Connected social accounts:</h3>
   <div class="col-md-12 text-center">
+    <!-- If the user's provider field (primary) is a social account, show it here -->
+    <div ng-hide="user.provider === 'local'" class="social-account-container">
+      <img ng-src="/modules/users/client/img/buttons/{{user.provider}}.png">
+      <i class="glyphicon glyphicon-check text-success user-primary-account" data-toggle="popover" title="Primary: {{user.provider}}"></i>
+    </div>
     <div ng-repeat="(providerName, providerData) in user.additionalProvidersData" class="social-account-container">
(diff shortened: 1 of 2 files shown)


codydaig:
I haven't tested but LGTM.

modules/users/client/views/settings/manage-social-accounts.client.view.html
```diff
@@ -1,6 +1,11 @@
 <section class="row" ng-controller="SocialAccountsController">
   <h3 class="col-md-12 text-center" ng-show="hasConnectedAdditionalSocialAccounts()">Connected social accounts:</h3>
   <div class="col-md-12 text-center">
+    <!-- If the user's provider field (primary) is a social account, show it here //-->
```
ilanbiala: end //?

mleanos: Just a habit. I can remove it.

mleanos: Should I remove the whole comment?

ilanbiala: No just the //, the comment is fine

mleanos: Ok. Will do. I think the habit stems from the good ole XML days :)

mleanos:
@ilanbiala I've removed the `//` from the inline comment.

I'm not really a UI guy, so I'm not sure if there's a better way to indicate that a Social provider is the User's Primary (`provider` field) account. Any objections to this?
gh-1241

ilanbiala:
No, I think that's fine. We're not really building a UI framework any, it's just to have some sort of frontend.

mleanos:
Merging.. We can discuss the other features that should go into Manage Social Accounts, and handle those enhancements in a new PR.

PR #1255, opened 2016-03-07
fix(gulp): remove gulp-node-inspector dependency

lirantal:
Due to the following considerations this PR would like to remove gulp-node-inspector
* gulp-node-inspector is yet another dependency for us to rely on
* the module is marked as blacklisted by gulpjs plugins as it is not making the correct use as gulp


mleanos:
LGTM.

codydaig:
LGTM

PR #1259, opened 2016-03-10
fix(client): lint issues for confirm() messages not referencing the `window` object

lirantal:


mleanos:
Could we use Angular's `$window` service instead? I think we're trying to move in that direction with $window, & $document. I mentioned it here https://github.com/meanjs/mean/pull/1202#discussion-diff-54761229

lirantal:
@mleanos you're right, I updated accordingly.

ilanbiala:
LGTM

mleanos:
I just double checked, and the ESLint warnings go away after the changes here are implemented.

PR #1263, opened 2016-03-12
feat(articles): Server-side paging for list

### LINKED ISSUE #1048: Object pagination
Hi there.

Is there any pagination system implemented when listing entities?

Thanks!

mleanos:
Adds capabilities to handle the Articles list paging server-side.

~~Updated tests to reflect changes.~~
EDIT: During review of this PR, some concerns were raised. These latest changes are an attempt to address those concerns. Once the implementation is agreed upon, and finalized, I will add tests.

Closes #1048


modules/articles/client/controllers/list-articles.client.controller.js
```diff
@@ -9,7 +9,45 @@
 
   function ArticlesListController(ArticlesService) {
     var vm = this;
+    var rightNow = new Date(Date.now());
 
-    vm.articles = ArticlesService.query();
+    vm.pageChanged = pageChanged;
+
+    init();
```
simison: BTW Angular 1.5 has a new lifecycle hook `$onInit`: 
http://angularjs.blogspot.lt/2016/02/angular-150-ennoblement-facilitation.html

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -81,7 +81,29 @@ exports.delete = function (req, res) {
...
+  *
+  * The idea behind the filter is that mongoose will start at the beginning of the collection, and move it's way up to 
+  * satisfy the skip condition. This could have performance issues. ref: http://stackoverflow.com/questions/5539955/how-to-paginate-with-mongoose-in-node-js/23640287#23640287
+  */
+
+  //var today = new Date(Date.now());
+  //var dateFloor = (new Date()).setDate(today.getDate() - 365);
+  //var filter = { created: { $gte: req.query.createdAfter || dateFloor } };
+
+  Article.find().limit(take).skip(page * take).sort('-created').populate('user', 'displayName').exec(function (err, articles) {
```
simison: I'm curious why not to use e.g. [mongoose-paginate](https://www.npmjs.com/package/mongoose-paginate) + [express-paginate](https://github.com/expressjs/express-paginate) ?

mleanos: I sort of jumped the gun on submitting this PR. I mostly wanted to get it out there to gain some momentum, and see what we could do without an outside dependency. If we can determine how we want this feature to work by keeping it basic, then we'll be in a much better position to choose the best library (if at all).

ilanbiala: @simison I'm not really a fan of Mongoose plugins, they seem a little shaky and not very stable to me, I really try to avoid them and also because most people who write Mongoose plugins seem less well-maintained.

simison: Actually, seems to be specifically true with mongoose-paginate. The package [got deprecated](https://github.com/edwardhotchkiss/mongoose-paginate/commit/36f3714d19d11f2b8f47f6841e0dfd108519613f) pretty bluntly at some point, but is rolling again it seems.

mleanos: I didn't consider mongoose-paginate at this point because its not compatible with Node < 4.0. Which doesn't matter too much since it looks like we're dropping support for Node .12 anyway.

modules/articles/client/controllers/list-articles.client.controller.js
```diff
@@ -9,7 +9,45 @@
...
+    function init() {
+      vm.itemsPerPage = 20;
+      vm.currentPage = 1;
+
+      vm.statistics = ArticlesService.statistics();
+
+      find();
+    }
+
+    function pageChanged() {
```
ilanbiala: This seems like it's just aliasing, and then you alias this into vm, so why not just do it straight to vm and save on code and memory.

modules/articles/client/controllers/list-articles.client.controller.js
```diff
@@ -9,7 +9,45 @@
...
+        //createdAfter: createdAfter || rightNow.setDate(rightNow.getDate() - 730)
+      }).$promise;
+    }
+
+    function onSearchSuccess(response) {
+      vm.articles = response;
+    }
+
+    function onSearchError(response) {
+      // handle error
```
ilanbiala: need to add code here.

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -81,7 +81,29 @@ exports.delete = function (req, res) {
  * List of Articles
  */
 exports.list = function (req, res) {
-  Article.find().sort('-created').populate('user', 'displayName').exec(function (err, articles) {
+  // Prepare paging parameters
+  var page = Math.max(0, req.query.page -1); // using a zero-based page index for use with skip()
```
ilanbiala: space after -

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -81,7 +81,29 @@ exports.delete = function (req, res) {
...
+  var take = req.query.take || 1000;
```
ilanbiala: Why 1000, that seems a bit high as the default, no?

modules/articles/server/controllers/articles.server.controller.js
ilanbiala: This also seems like it wouldn't be performant on collections with lots of entries. @mleanos have you found any sources saying this is the way to go?

mleanos: This [SO post](http://stackoverflow.com/questions/5539955/how-to-paginate-with-mongoose-in-node-js/23640287#23640287) suggests to use a filter on a date field to limit the amount of documents that are being queried. However, the post doesn't take into account the use of the `skip` method. I think the same would apply here with our implementation, if we just add the date field filter; on perhaps `created`. You can see my commented code.

modules/articles/server/controllers/articles.server.controller.js
mleanos: I can changed it to something like 100. I used 1000 just as an example. But you're right, it's too large for a default.

modules/articles/client/controllers/list-articles.client.controller.js
mleanos: I was going to add `vm.error = response.data.message;` I was hoping to get feedback before I built this out too much on the front-end.

mleanos:
@ilanbiala I've changed the implementation to scale better. I think it's pretty close to the best we can get, with a feature like this. Or maybe it's not :flushed: 

I've tested this with 500,000 Article documents. The response time from the server to retrieve the specific page increased as I moved up the pages. Eventually, at around the 9000th page (of 20 articles per page & querying all 500,000 Articles) the server request timed-out. Up to that point, as I reached this limit, the response would take somewhere around 8500 ms at around page 8000. The first few pages took around 25 ms, and I didn't see much increase until I started to get up to around page 50 (1000th Article). I know this isn't a detailed benchmark report of the performance. But I think we get the idea of how this will perform.

I'm wondering what our expectations of this feature are. How well can we expect server-side paging to perform? I don't know what any other library might be doing to solve the same issues that we're facing. At some point our users will have to decided if the implementation makes sense for them, and even if their data grew to a point where this became a performance issue, they could adjust the defaults that we've set for the paging feature.

I'm not opposed to looking at using a library that may be solving this. My goal at this point is to determine what we want out of this feature, and what hurdles we must overcome.

Let me know your thoughts. I'm more than happy to keep at this feature, until we're happy. There's no real rush here.

@codydaig @lirantal @simison Thoughts?

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -81,14 +81,45 @@ exports.delete = function (req, res) {
...
-      });
-    } else {
-      res.json(articles);
-    }
+  // Prepare paging parameters
+  var page = Math.max(0, req.query.page - 1); // using a zero-based page index for use with skip()
+  var take = req.query.take || 10;
+
+  /** 
+  * I don't like having the filter, so I'm removing it for now. It's going to introduce a bit of complexity.
```
mleanos: I left this comment here, so there is some context of why the filter is there & what the implications of using it are.

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -81,14 +81,45 @@ exports.delete = function (req, res) {
...
+  * satisfy the skip condition. This could have performance issues. ref: http://stackoverflow.com/questions/5539955/how-to-paginate-with-mongoose-in-node-js/23640287#23640287
+  */
+
+  var today = new Date(Date.now());
+  // only articles from the past 30 days
+  var dateFloor = (new Date()).setDate(today.getDate() - 30);
+  var filter = { created: { $gte: dateFloor } };
+
+  // cap the amount of articles queried
+  var query = Article.find(filter).limit(500).sort('-created');
```
mleanos: 500 might too large for a default amount of Articles to be querying. I'm unsure of where we might want to set this at. Obviously, this default isn't relevant until the amount of documents reached this limit. Our user's requirement will vary quite a bit.

With that said, perhaps 100 is better?

modules/articles/server/controllers/articles.server.controller.js
```diff
@@ -115,3 +146,21 @@ exports.articleByID = function (req, res, next, id) {
     next();
   });
 };
+
+/**
+* Articles collection statistics
+*/
+exports.collectionStats = function (req, res) {
```
mleanos: I left this here, in case we decide to keep it or use it.

modules/articles/client/controllers/list-articles.client.controller.js
simison: Actually I take my words back, seems like `$onInit` works only with new modules.

modules/articles/client/controllers/list-articles.client.controller.js
```diff
@@ -10,6 +10,39 @@
   function ArticlesListController(ArticlesService) {
     var vm = this;
 
-    vm.articles = ArticlesService.query();
+    vm.error = null;
+    vm.find = find;
+
+    init();
```
simison: In Jo

mleanos: You're right. We should use `activate()`. I looked into using Angular 1.5's `$onInit`. But I didn't see how it would work in this scenario. Also, I didn't want to over complicate this PR.

mleanos:
@ilanbiala @simison @codydaig Can we pick this back up since the other bigger PR's have been merged already?

I'd like to figure out if this PR is on the right track. If I need to take a different direction, or look for a library to use then I'm more than happy to do so. At the very least, we should figure out what we want from a feature like this.

mleanos:
@meanjs/contributors Need more eyes on this to determine if I should continue, or go down a different path.

lirantal:
Hi @mleanos , I'm overall neutral about it. Since it shows an example of pagination implementation I guess it may contribute to the stack to show users how to do that.

What is the use for the `db-collection-stats` API routes?

codydaig:
I like the general flow from what I can tell.

modules/articles/server/controllers/articles.server.controller.js
ilanbiala: 100 is probably better. People probably won't consider how intensive this could be.

mleanos:
@lirantal the `db-collection-stats` can be used to collect statistics on a particular collection. If we want to display the total count of Articles in the database, then we can use this route to get that info. In this current PR there is no use for it, but I left it there in case we decided to use it. It might be something that would be *only* used by the Admin area.

mleanos:
I'm not opposed to merely having this branch up on my fork, for reference on how to do it. I'm not 100% sold on the idea that we need it included in the framework. It does introduce some complexity, and potential performance issues; if users don't mind the limitations of a feature such as this.

lirantal:
Like I said before - I'm overall neutral. If we do want to merge this feature I'd remove the collection related code as it is not implemented and only adds extra confusion for readers using the framework.

ilanbiala:
I'd also say that we should merge this in and remove the stats code simply because we aren't using it.

mleanos:
Conflicts after merging #807. I'll rebase, and resolve the conflicts.

ilanbiala:
Sounds good.

modules/articles/client/controllers/list-articles.client.controller.js
mleanos: I've addressed this with my recent commits.

modules/articles/client/controllers/list-articles.client.controller.js
mleanos: I've added this error handling.

modules/articles/server/controllers/articles.server.controller.js
mleanos: Changed to "limit" to 100, and modified the default "take" to 50.

modules/articles/client/controllers/list-articles.client.controller.js
mleanos: updated the modified controllers in this PR to use `activate()`.

mleanos:
@ilanbiala Final review?

vaucouleur:
@mleanos 

Thanks for this PR !  Two quick comments.

This is the current implementation:
```
var query = Article.find(filter).limit(100).sort('-created');
query.count(function (err, count) {
....
}
```
The count is indeed useful to get the total number of hits -- but from a performance point of view, is the sort really necessary for the "count query" ? Could the sort be moved to the other query (the one that does the skip and get the results) ? 

Also, from a functional point of view this time, in this current implementation, the "count query" will be impacted by the limit, isn't it ? It's perfectly OK to limit the result set to a maximum of 100 items, but, I guess, the count of the total number items should always be correct. So, may be, the limit should be moved to the other query.

mleanos:
@vaucouleur Thank for jumping in here, and providing your feedback.

The way the queries are setup in this implementation is to avoid an performance pitfalls with large collections.

The filter query only gets the top 100 Articles from the past 30 days. The reason `limit()` is used here is to avoid iterating through a large dataset. For instance, if we have 100k Articles created in the last 30 days, we don't want to traverse such a large data set when we perform our `skip()` & `limit()` in the next part of the query.

Think of the first query (the one with the filter & limit) as **all** pages. The second part (with the skip & limit) is used to get the **current page**.

From my perspective, we need the `sort()` performed on the first query in order to ensure we get *virtually* the same top 100 Articles based on our filter, and we'll use this data-set as our pool of Articles that we're paging through.

Using the `created` data field, and applying a limit of 100 in the filter query is meant as a guide on how to implement pagination server-side. This filter, and limit, seemed appropriate for Articles. However, users of this framework can adjust this to their specific needs.

I combined the suggestion in [this SO answer](http://stackoverflow.com/a/23640287/2228227) with other suggested methods, for this implementation.

vaucouleur:
@mleanos 

We've got two separate query executions. I was only discussing the first one here, the **count** query.

- Regarding the sort on the **count** query: sorting a collection does not change its count, it's just a waste of CPU cycles.

- Regarding the limit on the **count** query:  Say we have 107 articles. If we are displaying "Currently showing articles from x to y (out of 107)", we cannot write 100 instead of 107 here -- that would be a bug, obviously. If we are just showing the "Go Next Page" button, we must still disable this button if we are on the last page -- therefore we must have the *exact* count.

vaucouleur:
@mleanos   I'm read your comment and you code again, and I think I got a better idea of what you tried to achieve with this "30 days filter". You tried to give a pragmatic solution to an extreme case. I would suggest, instead, to give a fully working solution to the average/reasonable case. That is, the exact total count, assuming a reasonable number of articles. If you want to give an example of server-side pagination, my guess is that this what most users/developers will expect. You can slightly increase the performance by doing the count and the actual page query in parallel (via async for example).

vaucouleur:
Of interest, mongoose 4.5:

https://github.com/Automattic/mongoose/issues/1908

ilanbiala:
@mleanos does the cursor functionality help at all? Would it change this implementation?

mleanos:
@ilanbiala We'd be able to implement a Mongoose Streaming type of solution using the Cursor API. However, I'm unsure of the actual impact of the server-side Mongoose query performance. I'm still looking into that. One thing for sure, is that we'd be able to send an Article back to the client one at a time which could improve load times, and take away some of the heavy lifting on the server-side; we'd be sending smaller packets of data back to the client.

I've played around with such an implementation, but it would introduce a bit more changes & complexity. However, it could be a step in the right direction since streaming data is probably the better way to go.

vaucouleur:
Irrespectively of pagination, streaming would be indeed a good idea. 

Some personal opinions, other might disagree:

- By **default**, we should have no sorting. The order of the articles should be similar to what we would get without pagination. This is both from a correctness and a performance point of view.

- But, the implementation should **support** sorting. Meaning that in addition to the "page" and the "take" argument, the client should be able to pass-in two more parameters: the sort direction {up,down,undefined}.. or {+1, -1, undefined}.. or any other variant/encoding.. it does not matter much.. As well as the sort field name, of course.

- Finally, the implementation should support filtering. That means two additional optional parameters, the sort field, and the filter value. 

One cannot do proper sorting on the client-side with server-side pagination; one cannot do proper filtering on the client-side with server-side pagination. Typically, sorting and filtering must be done on the whole collection, not just on the currently displayed page. So: in order, first optional filtering; then optional sorting; then optional paging.

The discussion whether the "total count" should be part of the same request, or part of a separate request is also very interesting. There are interesting pros and cons in terms of performance, reuse, etc. :

- One might prefer to hit the back-end only once; so one might prefer to have both the "total count" and the "page result" as part of the same response.

- Or, one might prefer to have a dedicated service for this, to be able to reuse the "count service" in an other context. For example, a "dashboard tile" might want to show the total number of "messages" in the user's mailbox, but might not need to show the first 20 messages -- one might also want to pass in a filter option for this count request (eg. to get the count of **unread** messages etc. ).

On final point. It's easy to forget about code duplication, because the standard implementation use very few business objects (article and user). But a reasonably large application will have 10, 20 of them, may be more. So the code in the server-side controller should be as small as possible. Ideally the pagination logic should be factored out. (This also applies to the client side; for example for services. IMHO, it's a bad idea to duplicate code across 20 or 30 services).

It will be interesting to see how this PR moves forward.

hyperreality:
This is a fantastic PR and I would like to see it merged soon. Great points from @vaucouleur too. @vaucouleur are you suggesting that we should add parameters for filtering and sorting in the route to provide a good example? That shouldn't be too hard given the great readable code by @mleanos.

Glancing through there appears to be one potential security issue - the `take` parameter is specified by the user so presumably they could set it to an arbitrarily high number which would defeat the point of paginating for performance reasons. I suggest that a `maxTake` variable is added (say 150), and an error is returned if user tries to grab more results than that at once.

lirantal:
@mleanos what's the state of this one?
Ready? can we merge?

vaucouleur:
@hyperreality That's right. My point is that it cannot be done later. Server-side optional sorting, server-side optional filtering, and server-side pagination must be delivered at the same time. Take your gmail web account for example. It's paginated, most likely. Yet when you enter, "meanjs" in the gmail search-box, the default behavior is to filter on all your emails -- not just the ones that are currently displayed on the current page. This is the default behavior in most cases, and rightly so.

lirantal:
@mleanos ? :)

vaucouleur:
@mleanos Let me know if I can help you. Further testing, documentation, a mongoose plugin, etc. I'm happy to do anything that could help moving your PR forward.

mleanos:
I'm still trying to figure out how to approach this PR, after the great feedback from @vaucouleur. I sort of put it as a lower priority, since this is a feature enhancement & there are other pressing issues/features in this milestone.

I agree with pretty much everything @vaucouleur is saying as well. I just want to do this right, and not rush it.

@hyperreality Thanks for pointing out the vulnerability. I'll adjust to take that into account.

lirantal:
@mleanos no problems. just stressing out that long-open PRs seems to never get in so you might want to push some effort and tight it down or we'll just end up closing it a few months later.
Either way, I totally appreciate all the great PR, review, and work you always put so thumbs up from me any time :)

simison:
@mleanos @lirantal I think it would be important to have this feature in — it's pretty basic functionality that different apps hit pretty early, yet it's kinda complicated to get right without good examples.

lirantal:
@simison I do but I am under the impression that this PR isn't yet ready to merge.

modules/articles/client/views/admin/form-article.client.view.html
```diff
@@ -3,7 +3,7 @@
     <h1>{{vm.article._id ? 'Edit Article' : 'New Article'}}</h1>
   </div>
   <div class="pull-right">
-    <a class="btn btn-primary" ng-click="vm.remove()">
+    <a class="btn btn-primary" ng-click="vm.remove()" ng-if="!vm.isNew">
```
mleanos: This is unrelated to what this PR is trying to accomplish, but I noticed a bug while developing this new feature. The "Delete" button was visible when the user was in "create" mode here.

modules/articles/tests/client/admin.list.articles.client.controller.tests.js
```diff
@@ -78,7 +78,6 @@
         // Set POST response
         $httpBackend.expectGET('api/articles').respond(mockArticleList);
 
-
```
mleanos: Removed extra empty line.

mleanos:
@vaucouleur @simison @meanjs/contributors 

I've refactored this PR to include filtering, and sort, along with the pagination functionality. This implementation should perform well, even on large collections (with the exception of the optional sorting, which I left an inline comment for).

You'll see the general direction that I'm going with the new Core server-side controller *mongoose-filtering*. I'm hoping to make the pagination generic enough to be used with any of Mongoose collections that users may add to their project.

I'd appreciate a review, and testing. I tested this with 1 million Articles in my database & didn't have any performance issues, other than the sorting. However, if I sorted on a field that is indexed then there was no longer any performance issues, even with the 1 million Articles. I explained about the indexing in the comment too.

mleanos:
I'll work on adding tests, while this is being reviewed and tested :)

mleanos:
Please note, I haven't implemented the filtering mechanisms on the client-side. I'm still a little unsure of how we want to implement this. We have a few considerations..

1) What type of filters should we use? equality checks, comparisons, something similar to a "like", etc.
2) Do we want to use Mongoose `$or`'s or `$and`'s.. or both?
3) How do we want to translate the client-side filters on the back-end? For instance, we may have some input that does an equality check, and others that are providing an `$in` type of search.

I definitely would appreciate someone helping out here that knows Mongoose querying well. I haven't written many complex Mongoose queries in some time, so I'm a bit rusty. And to boot, I'm trying to do this in a DRY manner :)

modules/core/server/controllers/mongoose-filtering.server.controller.js
```diff
@@ -0,0 +1,54 @@
+﻿// Service for managing Filtering/Parameterized-Queries.
+
+var _query = null;
+var _modifiedQuery = null;
+var _totalDocumentCount = null;
+var _page = 0;
+var _take = 50; // default to prevent unreasonable take() requests
```
mleanos: I think I should move the default, and cap (below) for `_take` responsibility to the collection controller. We may want have different defaults & caps on each collection. For instance, a messaging module might want only the last 50 messages returned, and Articles may want 50.

simison: Different clients might also want different number of pages (mobile), so it
should probably be possible to se it at URL/headers.

mleanos: Any client request can set the *take*, but I think the server should have the responsibility of managing a maximum *take*. This would protect from malicious, or erroneous, requests.

I've set the maximum in this case to be 150. However, I think it should be higher. Either way, it should be configurable to each implementation of the `mongoose-filtering` functionality; i.e. the calling controllers should be in control of the maximum & default *take*.

simison:
Did you have some script at hand to quickly create those million articles? :-)

mleanos:
@simison Here's a Gist. The maximum that I could seed per script run, without it timing out, is 100,000. So I just fired up 10 instances of a terminal and started the script :)

https://gist.github.com/mleanos/81fa1ca3730fc266df95b2946f0f0aec

modules/core/server/controllers/mongoose-filtering.server.controller.js
mleanos: I just re-read your comment. I didn't understand it correctly before. To be clear, you would like the "parameters" (take, & page) to be accessible in the URL? The API route `/api/parameterized-query/articles` is a POST since we could be sending complex data that we can't put in the query-string. I can pull the *take*, and *page* out of the request body & put them into the URL as query-string parameters.

Is that what you were driving at?

hyperreality: I think it makes sense to make *take* and *page* query-string parameters. It's how I've seen a lot of other websites do it, because it allows people to (at least for a short period of time) return to a similar place in a results feed rather than having to start from page 1 again when returning to the old URL.

mleanos:
Why so many messages Coveralls? :confused:

hyperreality:
That's a real mystery. They all link back to the same page and were sent simultaneously. My first thought was that they were for different builds, but there are 12 builds and 9 notifications  so that doesn't quite add up. Googling doesn't find any history on this issue whatsoever.

Perhaps @lirantal could disable coveralls notifications until we're at the bottom of it and we've also found a way to get client-side coverage reports going?

lirantal:
Let's keep monitoring it to figure out what's going on before disabling.

modules/core/server/controllers/mongoose-filtering.server.controller.js
```diff
@@ -0,0 +1,97 @@
...
+
+      // Send back the modified query
+      resolve({
+        query: self._modifiedQuery
+      });
+    }
+  });
+}
+
+function page(query, take, page) {
```
vaucouleur: Equivalent to what you wrote, but please consider the following:

function page(query, take, page) {
 if (!take) {
   return query;
 }
 return !page ? query.limit(take) : query.skip(page * take); 
}

modules/core/server/controllers/mongoose-filtering.server.controller.js
```diff
@@ -0,0 +1,97 @@
...
+  if (take !== null) {
+    query = query.limit(take);
+  }
+
+  if (page !== null & take !== null) {
+    query = query.skip(page * take);
+  }
+
+  return query;
+}
```
vaucouleur: Equivalent to what you wrote, but please consider the following:
 if (!take) {
   return query;
 }
 return !page ? query.limit(take) : query.skip(page * take);

vaucouleur:
@mleanos Naive cosmetic question: is the use of "underscore prefix" for variable names now encouraged ? Personally, It's never been my cup of tea, but I know that there are plenty of "underscore enthusiasts" out there ! :) Thanks a lot for your follow-up on this PR. I'll try to give further feedback soon.

modules/core/server/controllers/mongoose-filtering.server.controller.js
mleanos: Good suggestion. Thanks. I'll add comments as well, to be clear on what this is doing.

mleanos:
@vaucouleur I'm not a huge fan of using the underscores either. However, for these purposes I think it may be appropriate. I'm mimicking the `class` feature of ECMAScript 6. We're not quite there yet, to be able to fully support ES6. A feature of this new standard, is [Symbol's](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) which is used to implement *private* members of classes; even-though they're not **truly** private, I guess it's just the suggested way. In which case using names like "take", rather than "_take" make more sense. 

My reasoning is that when a developer is utilizing a class, they should understand that a property named with "underscore prefix" is signifying that it's meant to be "private". Of course, again, we can't make them truly private. A developer might see the "underscore prefix", and think "Maybe I shouldn't be using this property directly." :smile_cat:  Eventually, I would like to re-write this "service" in TypeScript, or at the very least using ES6. 

If the `mongoose-filtering` service proves to be useful, and we want to enhance it to include all the bells & whistles that one might expect from this type of service, then I would prefer to build this out into it's own library; i.e. a new Mongoose plugin. I'm integrating this into the Articles module, with this in mind. So there's a clear SOC between our API/Mongoose-models & our filtering service.

mleanos:
I've made some changes to the implementation. I think it's much closer to representing the most common use cases for server-side filtering, sorting, and pagination.

Admin's can use the filtering, and sorting, along with the pagination. The `/api/parameterized-query/articles` route is restricted to the "admin" role. Generally, a public facing API route wouldn't have a need to provide advanced filtering & sorting. Consequentially, the `api/articles` GET route method (`articles.list`) **only** allows for the paging options to be considered.

Aside from a very simple search box, on the Articles Admin feature, I don't think we should spend the effort on a full implementation of the advanced search capabilities on the client-side. At least not at this time. Our main concern should be providing a solution on the back-end for pagination, and providing simplistic examples for the UI. Most users would need very custom behavior to suite the needs of their own application.

@meanjs/contributors @simison @vaucouleur @hyperreality @Wuntenn Let's start leading/pushing-for a review. I know some have expressed their desire to have this merged. But we need to be sure we're comfortable with this implementation.

mleanos:
The build didn't start, so I'm going to close, and re-open, this PR. Hopefully, it will start the Travis builds, and Coveralls.

mleanos:
Travis build started, but no Coveralls.. all I did was rebase from master, to get the changes to the coverage that were just merged in.

@hyperreality @lirantal Any ideas?

hyperreality:
@mleanos I think the coveralls build will kick off after Travis is done now.

hyperreality:
I wonder if the Coveralls spam has something to do with this: https://coveralls.zendesk.com/hc/en-us/articles/203484329

Still a bit of mystery why this is happening to this PR and not others, though.

mleanos:
Thank you @hyperreality. Looking forward to the feedback.

mleanos:
FYI, I just deleted 8 of the duplicate Coveralls notifications in this PR.

lirantal:
Cool, thanks.
Would be great to test it out and merge.

mleanos:
I've been testing it. I encourage others to test manually. More importantly, I need a proper review of the implementation. If you see my comments above, there are some points of discussion still unaddressed.

Mainly, the client-side input for filters. What types of filters should we expect (and/or require) from the client-side request to the API? We need to consider different types of filtering: `$in`/`$nin`, a sort of `like('%xxxxx%')` comparison, `$eq`, etc.

Currently, the `mongoose-filtering` service does just a simple equality check based on the filter(s) (`[{ title: 'somesearchstring` }]`) sent with the request to the API. This works for a simple example, and we may just want to leave it at that. We could just expect that our users would extend this service/feature to suit their own needs.

Other than that, it's just a matter of reviewing and getting some LGTM's. If I can see that this implementation is generally accepted, then I will finalize the tests & we should be good to merge.

modules/articles/client/controllers/admin/list-articles.client.controller.js
```diff
@@ -10,6 +10,63 @@
...
+    vm.filter = filter;
+    vm.reset = reset;
+
+    activate();
+
+    function activate() {
+      vm.itemsPerPage = 10;
+      vm.currentPage = 1;
+      vm.filters = [];
+      vm.sorting = '-created';
```
hyperreality: Why specify these here when they are only going to be plugged directly into the find() function?

modules/articles/server/policies/articles.server.policy.js
```diff
@@ -20,6 +20,9 @@ exports.invokeRolesPolicies = function () {
     }, {
       resources: '/api/articles/:articleId',
       permissions: '*'
+    }, {
+      resources: '/api/parameterized-query/articles',
```
hyperreality: I see why you called this route parameterized-query, and we probably want to keep backwards-compatibility, but it does make me wonder whether we could retitle the existing articles routes /api/articles/list and /api/articles/list/:articleId and rename your query /api/articles/query

hyperreality:
@mleanos Just gave it a spin and it's great. Even into the tens of thousands of articles performance is retained. Made just a couple of nitpicky comments on the code.

For the filters I suggest one that finds any article with a title including the search term, ignoring case. I can't see how we could implement any more advanced filters at this stage. Although it would be nice to have a button to flip articles being sorting by ascended and descended creation date.

modules/articles/client/controllers/admin/list-articles.client.controller.js
mleanos: The point is that you may have multiple views, or user inputs, that change the filters and/or sorts.

Furthermore, you could have many different clients sending API requests to the back-end with varying filters & sorts.

modules/articles/server/policies/articles.server.policy.js
mleanos: I see what you mean, but I definitely prefer `/api/articles/:articleId`. It's more standard. The reason why I've used `parameterized-query/articles` is for two reasons. The first, you already pointed out, that there's a conflict with the `/:articleId` route, if I were to use `articles/query`. The second, is that any other collections that we may want to expose this feature to, could use the same pattern.

We may end up with something like:
`/api/parameterized-query/articles`
`/api/parameterized-query/users`
`/api/parameterized-query/some-other-data-collection`

I can see this type of API route configuration as a benefit, even if this *may* seem like a band-aid to a problem with how we're declaring our routes; given that we could have these conflicts. We can definitely look at our router configuration & solve the issue. But it's not something I wanted to do in this PR.

mleanos:
@hyperreality I agree that it would be nice to have the search using an "in-string" type of comparison, and to have sorting exposed with a user input. However, my worry is that we'd end up making the UI implementation of this too opinionated. Where our users will most likely have varying requirements for how the UI interacts with this feature. By keeping the example really simple, we can show the usefulness of the feature. While at the same time making it easy for our users to "roll-their-own" UI for it.

Thank for the feedback. If you wanted to test out even more Articles, I left a link (above) to a Gist with a script that seeds many Articles. Once the collection starts to grow substantially, the sorting should be modified to either use an indexed field, or no sort at all (thus, using the default on the server).

vaucouleur:
@mleanos Michael, I not understand the necessity for the "default sort" on _id. It breaks the natural order of mongodb, and according to my benchmark, it does no provide anything.  There are actually 3 possibilities: "asc sort", "desc sort", and "no sort" -- and "no sort" should be the default.  The only time I had to use a "default sort" on an id column was in the context of server-side pagination with a relational database (the query plan for a relational database might change depending on the page number, hence sorting becomes a necessity). As far as I know, this is not the case with mongodb, hence the default sort is not required. Am I missing something ?

https://github.com/meanjs/mean/pull/1263/files#diff-89b7f92b3fc6548ce65ea9989199bc83R84

lirantal:
Too big of a PR for me to dive into, would appreciate if you guys could come to an agreement.

mleanos:
@vaucouleur The default sort is just an example. Sure, it does break the natural order of the documents. Perhaps it was a bad choice for a default. However, I would like to include a default sort with the Admin list (`/api/parameterized-query/articles`) on the server-side. This is meant to show the capabilities of our paging-filtering-sorting solution. Would it be better if I changed the default in this method to sort on `created`, or `title`? I can also, update the comment to clearly explain more about this. I purposely chose to add the default sort to the list that's being accessed by the Admin area. For the public facing list of Articles, there's no default sorting. 

In the case where I was sorting, with 1M Articles in my db, there was no significant differences in performance than when I had 100,000 articles. Just as long as I was sorting on an indexed field. So I wanted to show that as well.

I understand the differences you're talking about with relational databases, and MongoDB, and their query execution plans. However, sorting is a very common use case for this type of feature, so why not show it?

Other than that, do you have feedback on the other aspects of these changes? I think we're close on this one, but there are few things to wrap up. The issue you brought up, being one of them. Thanks!

vaucouleur:
@mleanos I'm happy to agree with most things since I can always change the code to my liking, including keeping the natural order of the documents when "no sorting" is specified by the user :) So, not a big deal.

I did not see any other issues. It looks good, and I'm sure it will be very useful to many meanjs followers. Thanks again for your good work !

mleanos:
@hyperreality @codydaig @ilanbiala ?

lirantal:
@hyperreality @vaucouleur @codydaig @simison 
Merging this PR is now or never... can you please give your +1 or any constructive feedback to @mleanos to address standing issues?

hyperreality:
All my comments were nitpicks, the implementation looks good. I still think it would be best to provide an in-string search out of the box because it is such a common-use case which could be easily adapted or removed as the dev pleases, but I also understand why @mleanos doesn't want to add it in.

mleanos:
> I did not see any other issues. It looks good, and I'm sure it will be very useful to many meanjs followers. Thanks again for your good work !

@vaucouleur Thank you for your feedback, and the benchmark work you put into helping get this PR ready. This implementation wouldn't have been as good without your involvement! I salute you good sir :)

> Merging this PR is now or never...

Harshness.. haha. Thanks for putting a fire under our feet lately @lirantal (it's a good thing), leading to our final push for 0.5.0!

lirantal:
My pleasure :)

mleanos:
@simison Have you had a chance to review this? You had brought up using the http headers to pass back the pagination info in our Gitter discussion. Is that something you would like to see in implementation? I just want to double check with you before I merge, since Trustroots has a much different pattern for server-side Pagination. So you probably have good insight into whether or not this would work; I know you're trying to move away from mongoose-paginate.

Issue #1260, opened 2016-03-10
TODOs on the Client side


lirantal:
We have a few TODOs that were added, let's see if they make sense or clean them up:
* articles.client.controller.js#L34 - https://github.com/meanjs/mean/blob/master/modules/articles/client/controllers/articles.client.controller.js#L34
* users.client.service.js#L14 - https://github.com/meanjs/mean/blob/master/modules/users/client/services/users.client.service.js#L14

@mleanos would you like to take a stub at this?

mleanos:
I've been thinking about this. Both TODO's have the potential to be big changes.

1. Articles client-controller: May need a big refactor of the service. This could be a first phase of exploring client-side models.

2. Users Admin client-service: This could require some back-end router changes; similar to @trainerbill's suggestion of using `Express.router` in each route configuration.

I'll look into both of these to see if I can come up with a good solution to each, and how much work/refactoring is involved.

lirantal:
Ok so it's either we decide to fullfill the TODOs or we remove them from the code.
@ilanbiala , @codydaig to review if the TODOs are really necessary or not.

ilanbiala:
Both seem valid, and I think they could just be a quick refactor. @mleanos for the first we could just move the code, right? For the second, it might be a bit more refactoring in other places where we use this, so I'm not sure how the dependencies would work out, but if they stay clean across modules then it shouldn't be an issue.

lirantal:
One issue we tackled and merged.
Remaining issues?

mleanos:
@lirantal This [TODO](https://github.com/meanjs/mean/blob/master/modules/users/client/services/users.client.service.js#L73) can wait until after we release 0.5.0. This change will be a bit involved, and I'm still not sure how we want to go about it.

lirantal:
Sure

PR #1266, opened 2016-03-13
feat(articles): ArticlesService extended $resource

mleanos:
Extends the ArticlesService $resource object to include a custom method for creating, or updating, an Article instance.

Related #1260

### DIFF
--- a/modules/articles/client/services/articles.client.service.js
+++ b/modules/articles/client/services/articles.client.service.js
@@ -8,12 +8,46 @@
   ArticlesService.$inject = ['$resource'];
 
   function ArticlesService($resource) {
-    return $resource('api/articles/:articleId', {
+    var Article = $resource('api/articles/:articleId', {
       articleId: '@_id'
     }, {
       update: {
         method: 'PUT'
       }
     });
+
+    angular.extend(Article.prototype, {
+      createOrUpdate: function () {
+        var article = this;
+        return createOrUpdate(article);
+      }
+    });
+
+    return Article;
+
+    function createOrUpdate(article) {
+      if (article._id) {
+        return article.$update(onSuccess, onError);
+      } else {
+        return article.$save(onSuccess, onError);
+      }
+
+      // Handle successful response
+      function onSuccess(article) {
+        // Any required internal processing from inside the service, goes here.
+      }
+
+      // Handle error response
+      function onError(errorResponse) {
+        var error = errorResponse.data;
+        // Handle error internally
+        handleError(error);
+      }
+    }
+
+    function handleError(error) {
+      // Log error
+      console.log(error);
+    }
   }
 }());
--- a/modules/articles/client/controllers/articles.client.controller.js
+++ b/modules/articles/client/controllers/articles.client.controller.js
@@ -31,12 +31,10 @@
         return false;
       }
 
-      // TODO: move create/update logic to service
-      if (vm.article._id) {
-        vm.article.$update(successCallback, errorCallback);
-      } else {
-        vm.article.$save(successCallback, errorCallback);
-      }
+      // Create a new article, or update the current instance
+      vm.article.createOrUpdate()
+        .then(successCallback)
+        .catch(errorCallback);
 
       function successCallback(res) {
         $state.go('articles.view', {


mleanos:
If we're happy with these changes, and going this direction with extending the Angular Resource objects, I'll refactor the rest of the client-side Articles code to follow this pattern. For instance, I'll add a method for the `ArticlesService.query()` method called `find()`/`list()`.

I'll also add some extra logic to the `createOrUpdate` method to return a new Promise, so we can catch errors inside ArticlesService & handle them accordingly.

ilanbiala:
@mleanos take a look at https://github.com/angular/angular.js/issues/10692#issuecomment-69327003

mleanos:
I originally came across this implementation, when investigating https://github.com/meanjs/mean/pull/1149. However, when I came across http://sauceio.com/index.php/2014/07/adding-custom-methods-to-data-models-with-angular-resource/, i like it much better.

Using `angular.extend` allows us to add Class level methods, as well as instance methods. We can separate instance methods from the Class methods like so..
```javascript
var Article = $resource('api/articles/:articleId', {
      articleId: '@_id'
    }, {
      update: {
        method: 'PUT'
      }
    });

    // instance methods
    angular.extend(Article.prototype, {
      createOrUpdate: function () {
        var article = this;
        return createOrUpdate(article);
      }
    });

    // Class methods
    angular.extend(Article, {
      find: function () {
        var articlesResource = this;
        return articlesResource.query();
      }
    });

    return Article;

    function createOrUpdate(article) {
      if (article._id) {
        return article.$update();
      } else {
        return article.$save();
      }
    }
  }
```

Using `Article.prototype.find` would be ambiguous as to whether or not we're working with an instance or the Class itself. And would require a bit more logic.

mleanos:
@codydaig @ilanbiala @lirantal Any thoughts on this implementation?

mleanos:
@ilanbiala Thoughts?

rhutchison:
I personally am a huge fan of angular-model-factory.

modules/articles/client/services/articles.client.service.js
```diff
@@ -8,12 +8,29 @@
...
+    angular.extend(Article.prototype, {
+      createOrUpdate: function () {
+        var article = this;
+        return createOrUpdate(article);
+      }
+    });
+
+    return Article;
+
+    function createOrUpdate(article) {
```
ilanbiala: I would just consider moving this above the return, some people might get confused and forget about hoisting.

ilanbiala:
LGTM other than the one coder order concern I mentioned above.

modules/articles/client/services/articles.client.service.js
mleanos: I think having the function declarations after the return statement make it easier to read, and understand the implementation. 

It follows this [style from John Papa's guide](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#accessible-members-up-top). I don't feel super strong about it, but I think it would help others that are following, or aware of, the style-guide.

mleanos:
@rhutchison I really like the idea of having client-side model's. However, angular-model-factory didn't really impress me. I felt like it added more complexity than it's worth to the project. I'd love to see how you would integrate it into this framework though. Perhaps, there's a benefit that I just didn't see.

Our framework is quite lightweight, so it might be better to give our user's a simple approach, and let them decide if they wanted to implement something like angular-model-factory.

trendzetter:
Thanks! I think it is useful as example of how to use the service to create a separate client side data layer instead of preparing your data in the controller.

ilanbiala:
@mleanos other than adding tests, which may or may not be necessary, LGTM.

mleanos:
I'm pretty sure we already have tests for this functionality. I'll double check though.

ilanbiala:
@mleanos if we don't need to add tests, LGTM.

mleanos:
@ilanbiala I rebased this and added some additional error/callback handling from inside the service.

I verified that we already have test coverage for this functionality. We're testing the `save()` method of the Article client controller, which now implements this new service method.

@codydaig @ilanbiala @lirantal @simison LGTY? If so, I'll merge.

ilanbiala:
LGTM.

mleanos:
@meanjs/contributors I'll merge this in in the next few days, if there are no objections & I can get another LGTM. I just want to make sure we're still ok with this change.

lirantal:
I guess it's ok by me.

codydaig:
LGTM.

PR #1271, opened 2016-03-18
fix(dependency): Bump Mongoose to 4.4.8

mleanos:
Upgrades the Mongoose version to 4.4.8.

The need to limit Mongoose version to be less than 4.4.7 is no longer necessary due to a bug fix released with 4.4.8.

See ref: https://github.com/Automattic/mongoose/issues/3972

Resolves the need for #1264

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "method-override": "~2.3.5",
     "gulp-wiredep": "~0.0.0",
     "mocha": "~2.4.5",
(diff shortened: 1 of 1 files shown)


lirantal:
I was just following that thread too, LGTM.
If build succeeds just go ahead and merge.

mleanos:
Thanks Liran.

PR #1273, opened 2016-03-19
fix(config): typo

eduardojunio:
fix(config): typo

Fixed a typo in a comment.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -112,7 +112,7 @@ var validateSessionSecret = function (config, testing) {
 };
 
 /**
- * Initialize global configuration files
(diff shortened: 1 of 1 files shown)


lirantal:
that function still technically processes files though so I'm not 100% sure

ilanbiala:
It really doesn't matter, honestly the function does both, so I would say this is unnecessary.

lirantal:
I feel somewhat as Ilan does unless anyone else cares to chime in?

ilanbiala:
@lirantal I think we can close this. We'll reopen if someone feels differently.

PR #1276, opened 2016-03-19
Enhancement(core):Relative position attribute added to the header.

### LINKED ISSUE #1275: On small screen devices (width less than 767px), the expanded navbar covers the main page content.
Steps: 
1. Reduce the screen size to 767px orr . 
2. click on the collapsible button to expand the nav bar.
 After this, you will see the expanded nav bar covering the content section as shown in the images bellow:
screen shot 2016-03-19 at 3 21 26 pm

It will be great experience to have an enhancement to the boiler plate that keeps the main page visible as users dropdown the menu instead of covering the main page. This change overall gives a better User Experience.

generalgmt:
Enhancement(core): Relative position attribute added to the header.

Relative position attribute added to the header. proposed by @generalgmt

Fixes #1275

### DIFF
--- a/modules/core/client/css/core.css
+++ b/modules/core/client/css/core.css
@@ -1,5 +1,5 @@
-.content {
-  margin-top: 50px;
+.relative-header{
+  position: relative;
 }
 .undecorated-link:hover {
   text-decoration: none;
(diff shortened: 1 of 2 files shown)


codydaig:
LGTM

PR #1293, opened 2016-04-04
fix(user-Model): social signup email fix

jmclemore368:
fix(model): Fixed email portion of user schema

- Emails are flagged unique with a default of the empty string ''.
- However, Social signups, particularly Twitter, do not guarantee email
- If social signup results in an empty string, then the empty string is indexed as unique
- Furthermore, subsequent signups [without an email] will throw Mongo error 11000.
- This is due to the empty string now being indexed by mongo
- Fix: add 'sparse' and remove 'default' in the schema 
- Result: empty string is no longer indexed for the email field
- Social signups can now resolve successfully without providing an email

### DIFF
--- a/modules/users/server/models/user.server.model.js
+++ b/modules/users/server/models/user.server.model.js
@@ -47,9 +47,10 @@ var UserSchema = new Schema({
   email: {
     type: String,
     unique: true,
+    sparse: true,
     lowercase: true,
     trim: true,
-    default: '',
+    //default: '',
     validate: [validateLocalStrategyEmail, 'Please fill a valid email address']
   },
   username: {
--- a/modules/users/server/models/user.server.model.js
+++ b/modules/users/server/models/user.server.model.js
@@ -50,7 +50,6 @@ var UserSchema = new Schema({
     sparse: true,
     lowercase: true,
     trim: true,
-    //default: '',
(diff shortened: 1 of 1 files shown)


modules/users/server/models/user.server.model.js
```diff
@@ -47,9 +47,10 @@ var UserSchema = new Schema({
   email: {
     type: String,
     unique: true,
+    sparse: true,
     lowercase: true,
     trim: true,
-    default: '',
+    //default: '',
```
simison: Remove instead of `//`?

codydaig:
LGTM

mleanos:
We already have a PR that solves this. #1250.

codydaig:
@mleanos My bad. Forgot about that one sitting there. 

@jmclemore368 Thanks for the PR, but @mleanos is right, the PR already exists.

jmclemore368:
Thanks for the feedback. I'll be more mindful of existing PR's in the future :)

mleanos:
@jmclemore368 No worries. Thanks for the contribution. I can see how that PR slipped under your radar. I need to get back on it :)

@jmclemore368 Can you confirm that the PR does indeed resolve the issues your reported here?

PR #1296, opened 2016-04-05
[WIP] feat(core): use npm commands

simison:
Replace variety of gulp/grunt/sh/npm commands with npm commands.

Todo:
- [x] Add commands
- [ ] Test
- [ ] Test with Docker
- [ ] Update readme

See #1258 for discussion.

PR #1298, opened 2016-04-09
feat(gulp): test:server:watch task onlyChanged

### LINKED ISSUE #1297: Should gulp test:server:watch default to "onlyChanged"?
This:
```
gulp test:server:watch --onlyChanged
```
...should really be default when hitting that test-watch task (`gulp test:server:watch`).

Are there situations when you'd want to run all the tests when changing something at a single test file? Wouldn't it be enough to run all the tests only at the end of development manually?

Thoughts?

mleanos:
The test:server:watch Gulp task has a parameter `onlyChanged` that is used to specify that if a server test file was modified, then only run that test suite. This removes the parameter, and sets the task behavior to the aforementioned.

Closes #1297

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,11 +19,13 @@ var _ = require('lodash'),
   pngquant = require('imagemin-pngquant'),
   path = require('path'),
   endOfLine = require('os').EOL,
-  argv = require('yargs').argv,
   protractor = require('gulp-protractor').protractor,
   webdriver_update = require('gulp-protractor').webdriver_update,
   webdriver_standalone = require('gulp-protractor').webdriver_standalone,
   KarmaServer = require('karma').Server;
+
+// Local settings
+var changedTestFiles = [];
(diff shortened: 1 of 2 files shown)


mleanos:
@codydaig @ilanbiala @simison Not trying to complicate the issue here, but I did a refactor on the implementation for this Gulp task. I did this in a separate commit so it can easily be reverted if these changes aren't accepted.

I think this is a bit cleaner, and it works as a stand-alone watch task. I don't think it should be grouped in with the other watches, since this is meant for a very specific development task.

Also, I removed the dependency on the yargs package, from this task. If the refactor is agreed upon, I'll remove the package from the library as well.

Any thoughts?

ilanbiala:
LGTM.

mleanos:
I'll merge if the build passes, since this is a pretty simple change & so that this isn't blocking the progress of #1296.

PR #1300, opened 2016-04-10
Feature/jwt auth lib

mleanos:
I've simplified the JWT Authentication implementation from #1163. Everything seems to be working correctly, but this definitely needs more eyes & manual testing.

I've managed to keep the changes to a minimum, so that it's easier to review. Also, I added separate commits for the changes to the test coverage.

Let me know how it looks, or if there's something missing/broken.


mleanos:
It looks like the Social Accounts OAuth isn't working. Does someone want to take a look at that as well?

Anyone, feel free to submit PR's to my branch if you'd like to contribute here.

config/lib/socket.io.js
```diff
@@ -6,11 +6,9 @@ var config = require('../config'),
   fs = require('fs'),
   http = require('http'),
   https = require('https'),
-  cookieParser = require('cookie-parser'),
   passport = require('passport'),
   socketio = require('socket.io'),
-  session = require('express-session'),
-  MongoStore = require('connect-mongo')(session);
+  ExtractJwt = require('passport-jwt').ExtractJwt;
```
ilanbiala: Where do you use this?

modules/users/server/config/strategies/jwt.js
```diff
@@ -0,0 +1,32 @@
...
+ * Module dependencies
+ */
+var passport = require('passport'),
+  JwtStrategy = require('passport-jwt').Strategy,
+  ExtractJwt = require('passport-jwt').ExtractJwt,
+  User = require('mongoose').model('User');
+
+module.exports = function (config) {
+  var opts = {
+    jwtFromRequest: ExtractJwt.versionOneCompatibility({ tokenQueryParameterName: 'auth_token' }),
```
ilanbiala: Why v1?

modules/users/server/config/strategies/jwt.js
```diff
@@ -0,0 +1,32 @@
...
+    secretOrKey: config.jwt.secret
+    // opts.issuer = "accounts.examplesoft.com",
+    // opts.audience = "yoursite.net"
+  };
+
+  passport.use(new JwtStrategy(opts, function (jwt_payload, done) {
```
ilanbiala: camel case

modules/users/server/config/strategies/jwt.js
```diff
@@ -0,0 +1,32 @@
...
+    User.findById({ _id: jwt_payload.user }, '-salt -password', function (err, user) {
```
ilanbiala: camel case

modules/users/server/config/strategies/local.js
```diff
@@ -21,9 +21,9 @@ module.exports = function () {
         return done(err);
       }
       if (!user || !user.authenticate(password)) {
-        return done(null, false, {
+        return done({
```
ilanbiala: Why is this syntax different than before?

ilanbiala:
Looks like a good start. I'll look again tomorrow afernoon/Tuesday.

config/lib/socket.io.js
mleanos: I'm not. It was left over from my attempts to get SocketIO working with the JWT. I'll remove this.

modules/users/server/config/strategies/jwt.js
mleanos: I needed to add to get our SocketIO to authenticate. `versionOneCompatibility()` will first look for the JWT Token in the header, if it doesn't find it there it will look for it in the querystring. Express will send the token in the header, and SocketIO will send it in the querystring.

simison: I was digging up more info about this:

- http://stackoverflow.com/a/27618749/1984644
- https://github.com/facundoolano/socketio-auth
- https://github.com/socketio/engine.io-client/pull/379

modules/users/server/config/strategies/local.js
mleanos: I think this is the proper signature for the callback here.

mleanos:
@ilanbiala @simison I modified the SocketIO configuration to authenticate using the JWT Token. Thanks for steering me to the `socketio-auth` package @simison. I like their implementation, and it's super simple. Now the SocketIO configuration is much simpler :) 

However, the client-side SocketIO tests are failing. If someone wants to help out here, that would be much appreciated.

mleanos:
Tests are passing. I haven't looked at the OAuth issues yet. I'll try to work on that in the next few days.

In the meantime, please feel free to review & provide feedback.

config/lib/express.js
```diff
@@ -89,6 +90,21 @@ module.exports.initMiddleware = function (app) {
   // Add the cookie parser and flash middleware
   app.use(cookieParser());
   app.use(flash());
+
+  // Authorize JWT
+  app.use(function (req, res, next) {
```
mleanos: I think it would be better to use the new `config/lib/authorization` service as the middleware here.

Something like:
```javascript
app.use(authorization.authorize);
```

Any objections? Or other input on this part of the implementation?

trendzetter:
I am having a hard time trying to understand what happens exactly, not that it's bad, it's just complicated and elaborate. I would consider spitting up the tests in a separate commit so just reading what changes is less unwieldy. 
One consideration that came to my mind is why signing only the id as payload? Shouldn't the roles also be signed as "claims" so no fiddling with that by enterprising users? Maybe I am just not far enough in my understanding of what happens and how jwt is used, just a first thought.
Does it work, should I test the functionality?

config/lib/authorization.js
```diff
@@ -0,0 +1,35 @@
...
+
+var auth = {
+  signToken: signToken
+};
+
+// Export the token auth service
+module.exports = auth;
+
+// Sign the Token
+function signToken(user, options) {
```
trendzetter: Are there any planed uses for the options parameter?

mleanos:
@trendzetter I broke things down into more focused commits. While looking at the changes here, you can specify a single commit to view. That should makes things easier for you to review.

Yes. Everything appears to be working, except for the Social OAuth functionality. I haven't looked into why it's broken, but my guess is the difference in the request handling. If you would like to look into that I would appreciate it. Otherwise, feel free to take this for a test drive.

mleanos:
> One consideration that came to my mind is why signing only the id as payload? Shouldn't the roles also be signed as "claims" so no fiddling with that by enterprising users?

I it may be a good idea to include the roles with the payload. Also, adding the expiration date might be a good idea as well, to further secure the payload by adding a dynamic field.

Anyone else have thoughts on that idea?

config/lib/authorization.js
mleanos: this parameter is there for anyone that wants to add/override any options coming from the env configuration.

See below where we're merging the two objects.

trendzetter:
I will take another look to the separate commits

simison:
@trendzetter Check also https://github.com/sindresorhus/refined-github — it has some features that ease reading longer PR's.

config/lib/express.js
trendzetter: So you move the authorize method out of express into it's own authorization "class". Seems a good idea to separate those
+ it makes the file a little shorter, not that it is too big now but it could easily become too big as someone needs to add something more.

PierreBrisorgueil:
Hello @mleanos  (sorry for my english :))

For me, we have a little problem from this comit 
https://github.com/trainerbill/mean/commit/09389f035d4f204180de67b5a0a80b6673003cb8

on modules/core/client/app/init.js

```
Authentication.ready
    .then(function (auth) {
```

In your pull request it's this 

```
    Authentication.ready
      .then(function (auth) {
        $rootScope.$on('$stateChangeStart', stateChangeStart);
      });
```

Authentication.ready return readyPromise.promise, bu you never go in 'then' so you never go in 'stateChangeStart' and never check roles on my side. I have a look on this.

mleanos:
@PierreBrisorgueil I'm not quite following where this issue is. Is the issue with the implementation of the `Authentication.ready` method in this PR?

What do you mean by "never go in 'then'"? Maybe provide inline comments to the commits in this PR to clearly point out where/what the issue is.

Thank you for your feedback. I'd like to get this PR back on track, so I appreciate any contribution.

PierreBrisorgueil:
@mleanos Yes it's this, 

in fact in this PR (https://github.com/meanjs/mean/tree/b14a5bf5b657e05d62899f752cf6a9913ae49a44) 
based on https://github.com/trainerbill/mean/commit/09389f035d4f204180de67b5a0a80b6673003cb8


If we add a console.log : 

```
  Authentication.ready
      .then(function (auth) {
        console.log("MeanJS");
        $rootScope.$on('$stateChangeStart', stateChangeStart);
      });
```

In my test with @farhatmo, if you are sign in, after you will never see the console.log("MeanJs"), when you change state. That's what I meant by you will "never go in then", and never check roles in front part (when user is connected).

In fact, in https://github.com/meanjs/mean/blob/b14a5bf5b657e05d62899f752cf6a9913ae49a44/modules/users/client/services/authentication.client.service.js 

```
    var readyPromise = $q.defer();

    var auth = {
      user: null,
      token: null,
      login: login,
      signout: signout,
      refresh: refresh,
      ready: readyPromise.promise
    };
```

var auth.ready never get the new status and update from readyPromise.promise. So the 
information is not propagated to core.client.route-filter.js and Authentication.ready var is never update by promise. So you never go in "then". You never get the resolved status of promise in auth.ready. 

For us (me and @farhatmo), you need to make something like this : 

**authentication.client.service.js**

```
    //var readyPromise = $q.defer();

    var auth = {
      user: null,
      token: null,
      login: login,
      signout: signout,
      safe: safe,
      refresh: refresh,
      readyPromise: $q.defer(),
    };
```

`        readyPromise.resolve(auth);
`
=>

`        auth.readyPromise.resolve(auth);
`

**core.client.route-filter**

```

    Authentication.readyPromise.promise
    .then(function(auth) {
      $rootScope.$on('$stateChangeStart', stateChangeStart);
    });

```

mleanos:
@PierreBrisorgueil Thank you for your in depth explanation. I'll implement your suggested fix for this.

PierreBrisorgueil:
@mleanos perfect 👍  :)

We have one last problem, now (with this modifications) we tcheck users role in order to protect front part display.  But we do this only if we  execute "stateChangeStart". 

This event is called, only if we change state. If you refresh the page, actually, you never go in stateChangeStart, and don't tcheck roles. So the user will see the template. It's not critical, it's only in client side .. 

But i think it's not really clean, we probably can protect this case. ( create a function to check roles from two events, not only stateChangeStart  ... i d'ont know)

I have a look on this quickly if possible.

simison:
BTW, some food for thought: http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/

farhatmo:
@simison Thanks for sharing the article.

I find the article very interessting but i missed some explanations. 

I think that :

- JWT is much more scalable than sessions: session autho needs to store those sessions in db and for some application with more then 1 millions user it can be a problem (there are some solutions) but for JWT we handle only one key that can be sotored in one file.  Why we need to deal other framework in order to ensure scalability ?

- Stateless || Stateful :  should be a choise of the application. It depends of the contexte . Both can be mixed for sensitive data. 

- Security: In theory , JWT is as secure as sessions.

I think the solution implemented by default of JWT is suitable for 90% of applications. Still 10% of sensitive applications, they should  use most appropriate solutions for their context without excluding completely JWT.

ilanbiala:
@mleanos if we are still moving ahead with JWT, can you rebase this so we can take a look and finalize everything?

mleanos:
@meanjs/contributors I've rebased this and fixed a couple minor conflicts. I've spent some time testing this & much of the functionality appears to be working with the new Token-based Authentication.

I'll go back, and look at any unaddressed issues that have been brought up in this discussion. One thing, IIRC, is that there are issues with the social logins. I'm not sure if they're issues that were pre-existing (currenly in *master*), or if these changes introduced them.

It would be a huge help if we can start reviewing, and continuing the conversation here. I'd really like to get feedback & get this PR back on track.

lirantal:
@mleanos so we're only left with the problems of social logins?

mleanos:
Actually, I'm also working on the issue brought up here: https://github.com/meanjs/mean/pull/1300#issuecomment-224539828

So the only two issues I'm aware of are the above, and the social logins. We're pretty close :) 

One concern that I'd like to express is that we shouldn't merge this in *just* because it's working, and all the functionality is there. 

For a change of this magnitude, and importance, we should make sure we really want to move to Token-based Authentication & that this implementation is as secure as we need it to be. I'd say a real thorough review is in order.

lirantal:
Ok, then I'll keep this open and definitely not for 0.5.0

simison:
FYI, some reading: https://angularjs.blogspot.lt/2016/11/easy-angular-authentication-with-json.html

Ghalleb:
First, thanks for the good job. I'm using 0.4.2 and it works like a charm.

But I was expecting JWT authentification to be in 0.5.0 so I can integrate my project with a mobileapp...
Is there a chance for this feature to be in 0.5.0 release or in near future?

Mean.js is a great framework, but with JWT integration and the ability to use it as a pure back end for mobile app, it would be perfect.

PierreBrisorgueil:
Hello @Ghalleb you can use it with an app without jwt auth, with basic auth and https :) 
I have a sample like this with iOS (swift1.2 or 2 don't remember) if you have some questions

But if we switch to JWT, always agree for my side, done in my fork, I will try to push samples Apps switft, iOS & tvOS maybe :)

Always in the idea of having a better and quicker starter for projects :)

Ghalleb:
@PierreBrisorgueil 
If I can have access to your sample, it would be nice.
Does it work with social login as well?
It can do the job while waiting for JWT Authentication...

PierreBrisorgueil:
@Ghalleb no, only basic inscription, and login, i will have a look on this quickly an share you a quote of code here if i can, as soon as possible :) 

The require https for security, with Let's Encrypt, it's perfect. 

can't make a sample for the moment, to much work, but i will try after ;)

protheantom:
Please don't forget this

JWT is necessary for better scaling and for reusing the api for mobile applications (yet discussed some comments ago)

Thanks for the work @mleanos

mleanos:
I've managed to get everything working, except for the social logins. For some reason, the callback is not receiving the `req` object that contains the `user`. It could be that since sessions aren't being used any longer, that passport isn't able to retrieve the user when the social provider responds to the callback. Although, it seems that the request **should** still have the user embedded. Any help here would be much appreciated; I could use a second set of eyes on it. 

As for why the build is failing, I'm unclear since the build & tests pass locally (Windows 10 x64, and Node `6.6.0`).

Ghalleb:
Hi,

Any news on JWT integration and Social Login?
It doesn't seems to have a lot of commits lately

Issue #1294, opened 2016-04-04
Running an AngularJS App in Production


simison:
Two things from https://docs.angularjs.org/guide/production

- [ ] Disabling Debug Data
- [x] Strict DI Mode

Thoughts how to implement these?

codydaig:
I think to disable debug data, we have to somehow load config into the client side. This is something that has come up a lot, but has not been implemented. 

However, I think we can do the strict DI mode with the index.html template in swig.

pgrodrigues:
I opened a [similar issue](https://github.com/meanjs/mean/issues/903) a few months ago.

lirantal:
Will you guys work on a PR for this, sounds like a good idea.

lirantal:
@simison you handled in the recent PR #1457  the debug data. What about the ng-strict-di? how do we handle that?

PR #1302, opened 2016-04-11
feat(core): Use ngStrictDi in production

pgrodrigues:
Slightly changed layout template to use Angular's Strict
Dependency Injection if the app is running in production.

Fixes part of meanjs/mean#1294

### DIFF
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -1,5 +1,5 @@
 ﻿<!DOCTYPE html>
-<html lang="en">
+<html lang="en" {% if env === "production" %}ng-strict-di{% endif %}>
 <head>
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -25,6 +25,7 @@ var config = require('../config'),
  */
 module.exports.initLocalVariables = function (app) {
   // Setting application local variables
+  app.locals.env = process.env.NODE_ENV;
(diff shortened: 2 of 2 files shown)


modules/core/server/views/layout.server.view.html
```diff
@@ -1,5 +1,5 @@
 ﻿<!DOCTYPE html>
-<html lang="en">
+<html lang="en" {% if env === "production" %}ng-strict-di{% endif %}>
```
ilanbiala: I don't see a reason to not have this turned on all the time. Is there?

simison: @ilanbiala  Wouldn't work if you use ngAnnotate?

pgrodrigues: @ilanbiala right now ngAnnotate is only being used in build task for production so I kept ngStrictDi only for production otherwise it will break.

ilanbiala: We could just annotate in development too, which I'm not opposed to. Honestly I think ngStrictDi would prevent errors in production if we also check it in development.

lirantal:
@meanjs/contributors any of you care to take it further to implement @ilanbiala suggested changes with making it work regardless of the environment? it's a rather small change.

lirantal:
@pgrodrigues you're also welcome to just update the PR to accommodate the changes.
And of course congrats on passing a stage in the euro 2016 stages! :-)

portugal

pgrodrigues:
@codydaig is already taking care of this. Time to celebrate the Euro2016 win :)

lirantal:
definitely! :)

PR #1303, opened 2016-04-12
fix(gulp): Trailing spaces lint error in gulpfile

mleanos:
Removes the trailing spaces in the gulpfile causing a linting error.


mleanos:
Merging since it's a small fix.

PR #1314, opened 2016-04-17
feat(docker): persisting the container names

lirantal:
Persisting the container names for the docker containers that get created allows to easily use an alias instead of the container id which is automatically generated and is usually different.


ilanbiala:
LGTM.

simison:
Tested, all good.

lirantal:
Thanks guys.

PR #1318, opened 2016-04-20
fix(models, routes, and sockets): incorrect default asset paths

Wuntenn:
Models, routes and sockets currently work because of the 
catch-all AllJS path in the same file above my changes. We
can't rely on this to be true indefinitely, therefore I've 
updated the paths to maintain their integrity and to reflect the
convention currently in use.

### DIFF
--- a/config/assets/default.js
+++ b/config/assets/default.js
@@ -51,9 +51,9 @@ module.exports = {
     gruntConfig: ['gruntfile.js'],
     gulpConfig: ['gulpfile.js'],
     allJS: ['server.js', 'config/**/*.js', 'modules/*/server/**/*.js'],
-    models: 'modules/*/server/models/**/*.js',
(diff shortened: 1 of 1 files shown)


ilanbiala:
@Wuntenn build is failing, can you take a look?

codydaig:
For the way that glob works, these should be `**/*.js`. This will include all js files in that directory and any nested directories. Just doing `*.js` does not go into nested directories.

Wuntenn:
So my changes would've forced rigid use of a single directory beneath model, sockets, and routes and created a dependency on AllJS for all other cases - This passes all tests locally. Good thing it failed here! Thanks @codydaig!

ilanbiala:
This can be closed right?

Wuntenn:
Yup this can be closed!

PR #1327, opened 2016-04-28
fix(redirect url): on login from social

### LINKED ISSUE #1284: Twitter login doesnt' redirect properly because passport.authenticate

Thanks in advance!
Regards

snlacks:
Issue originally noted in Twitter by @jorgeram
Builds on workaround presented by @ChrisMaher

Logging in with social is broken for some, the redirectURL did not appear to be used and was causing an error as it wasn't the type the redirect function was expecting.

Fixes #1284,

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -116,7 +116,7 @@ exports.oauthCallback = function (strategy) {
           return res.redirect('/authentication/signin');
         }
 
-        return res.redirect(redirectURL || sessionRedirectURL || '/');
+        return res.redirect(sessionRedirectURL || '/');
       });
     })(req, res, next);
   };


ilanbiala:
@mleanos do you know if this will be fixed/affected by JWT?

mleanos:
With the JWT implementation, we won't be setting the `req.session` any longer.

@snlacks Can you come up with an alternative that will work with JWT. See #1300. 

Perhaps just adding the `redirect_to` as a request parameter? It probably should be stored in the querystring any way.

lirantal:
Just stepping in to note that JWT is not going to be a replacement for working with sessions I assume? we need to support both IMO so I wouldn't be dismissing the need for `req.session` entirely.

ilanbiala:
@lirantal I'm pretty sure JWT and sessions would conflict, it's either one or the other if I'm not mistaken. @mleanos can you elaborate?

mleanos:
My understanding is that we wanted to move away from sessions all together. Going completely session-less, which is how mobile apps tend to work.

@lirantal Are you suggesting that we use JWT & Sessions with Express simultaneously? I may not fully understand your perspective. Could you provide more details?

lirantal:
Problem already fixed in another PR superset originated by @fauria in PR: https://github.com/meanjs/mean/pull/1388

PR #1328, opened 2016-05-01
feat(CI): Add Node v6 to Travis CI + remove V5 from allow_failures

simison:
Add Node v6 to Travis CI, but let it fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 node_js:
   - 4
   - 5
+  - 6
 os:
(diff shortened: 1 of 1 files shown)


simison:
v5 seems to be passing atm. Should I remove it from `allow_failures`?

lirantal:
Yeah we can probably remove it as it seems stable enough

simison:
Done + squashed.

codydaig:
LGTM

PR #1283, opened 2016-03-28
fix(users): Incorrect UsersService injection

t-jennings:
Updated edit profile controller to use the users service correctly. Edit profile now working. Closes #1282.

PR #1329, opened 2016-05-02
feat(users): Edit Profile client controller tests

mleanos:
Adds client-side tests for the Users Edit Profile client controller.

1) should have user context
2) should update the user profile
3) should set vm.error if error

Related #1283

### DIFF
--- a/modules/users/tests/client/edit-profile.client.controller.tests.js
+++ b/modules/users/tests/client/edit-profile.client.controller.tests.js
@@ -1,0 +1,92 @@
+﻿(function () {
+  'use strict';
+
+  describe('Edit Profile Controller Tests', function () {
+    // Initialize global variables
+    var EditProfileController,
+      $scope,
+      $httpBackend,
+      $location,
+      Authentication,
+      UsersService;
+
+    // The $resource service augments the response object with methods for updating and deleting the resource.
+    // If we were to use the standard toEqual matcher, our tests would fail because the test values would not match
+    // the responses exactly. To solve the problem, we define a new toEqualData Jasmine matcher.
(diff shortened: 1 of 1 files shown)


modules/users/tests/client/edit-profile.client.controller.tests.js
```diff
@@ -0,0 +1,92 @@
...
+
+      it('should update the user profile', inject(function (UsersService) {
+        // Set PUT response
+        $httpBackend.expectPUT(/api\/users/).respond();
+
+        // Run controller functionality
+        $scope.vm.updateUserProfile(true);
+        $httpBackend.flush();
+
+        expect($scope.vm.success).toBe(true);
```
trendzetter: Is it sufficient to check if success is true? I think maybe it should also check if the properties of the mock users are really changed? Just trying to offer some feedback and learn a thing or 2 at the same time.

mleanos: If our controller sets the `success` view model property to `true', then we can assume our controller is behaving as expected. I don't see a need to test the actual back-end functionality here as well. We already have CRUD server tests for that.

ilanbiala:
@mleanos any progress with this?

modules/users/tests/client/edit-profile.client.controller.tests.js
trendzetter: Ok I understand

mleanos:
@ilanbiala I think this is ready. I've just been waiting for others to give their approval. Is there something specific that seems to be outstanding with this?

lirantal:
@mleanos go ahead and merge then?

mleanos:
Merged. Thanks @lirantal

PR #1337, opened 2016-05-18
Fix ESLint console warnings

### LINKED ISSUE #1284: Twitter login doesnt' redirect properly because passport.authenticate

### LINKED ISSUE #1336: ESLint console warnings
When running ```grunt```, ```eslint:target``` warns ```warning  Unexpected console statement  no-console``` on the following files: ```gruntfile.js```, ```config/config.js```, ```config/lib/app.js```, ```config/lib/express.js```, ```config/lib/logger.js```, ```config/lib/mongoose.js```, and ```config/lib/seed.js```.

```
node -v
v5.10.0

npm -v
3.8.3
```

### LINKED ISSUE #1347: issues about use social account to login
1. after executing passport.authenticate method,  the program will execute _"function (err, user, redirectURL) {}"_  callback function, I try to use github strategy to login, and found that the **redirectURL**  will become an object when we do not set any value, and it will cause the application redirect to a wrong url such as "/api/auth/github/%5Bobject%20Object%5D". 

2. in SocialAccountsController controller I found that the '$scope' still in using:
_function hasConnectedAdditionalSocialAccounts() {
      return ($scope.user.additionalProvidersData && Object.keys($scope.user.additionalProvidersData).length);
    }_
need use vm to instead.

fauria:
fix (config): ESLint console warnings

Adds ```'no-console': 1``` option to ```.eslintrc```.

Fixes #1336

**Update:**

fix redirect to ```'[object Object]'``` after login with social accounts. 

Fixes #1284 

**Update 2:**

fix ```TypeError: Cannot read property 'additionalProvidersData' of undefined at SocialAccountsController.hasConnectedAdditionalSocialAccounts```. Use ```vm``` instead of ```$scope```.

Fixes #1347


fauria:
Those Travis failures seem to be unrelated to this PR, although they should be checked, specially #3 (```Error: expected 200 "OK", got 400 "Bad Request"```)

ilanbiala:
Reran build, I think that was just a temporary thing.

fauria:
Fixes #1284

fauria:
Should not affect code coverage...

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -116,7 +116,7 @@ exports.oauthCallback = function (strategy) {
           return res.redirect('/authentication/signin');
         }
 
-        return res.redirect(redirectURL || sessionRedirectURL || '/');
```
ilanbiala: What does this fix?

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -104,7 +104,7 @@ exports.oauthCallback = function (strategy) {
     var sessionRedirectURL = req.session.redirect_to;
     delete req.session.redirect_to;
 
-    passport.authenticate(strategy, function (err, user, redirectURL) {
```
ilanbiala: Why change variable name?

fauria:
The variable name has been changed to make it consistent with [Passport.js documentation](http://passportjs.org/docs). Currently its not being used.

modules/users/server/controllers/users/users.authentication.server.controller.js
fauria: The variable name has been changed to make it consistent with Passport.js documentation. Currently its not being used.

modules/users/server/controllers/users/users.authentication.server.controller.js
OneOfTheWorld: seems can not simply update: 
_return res.redirect(redirectURL || sessionRedirectURL || '/');_
to:
_return res.redirect(sessionRedirectURL || '/');_

because before **oauthCallback**  function called, there will be another function **saveOAuthUserProfile** execute before it and have the following code snippet to pass value to redirectURL(changed to info): 
      `user.save(function (err) {
        return done(err, user, '/settings/accounts');
      });`
so, if pass value to redirectURL, the value will be right, if do not pass anything, the value will become an object.

for your reference, thanks.

lirantal:
@fauria so where does this PR stand?
To be honest it would've been better if you had made a separate PR for each fix, because this way we're coupling all the fixes to this one PR. Not ideal...

fauria:
@lirantal Yeah, I know. I thought they were going to be merged faster, so I didn't think on multiple branches...

lirantal:
So right now I'm not sure on the state of this PR - is everything really taken care of? or do you want to create 3 new PRs for us to review and assess them each on it's own?

fauria:
They are quite trivial, doesn't look like they need there own PR...

modules/users/server/controllers/users/users.authentication.server.controller.js
lirantal: @fauria can you please refer to this comment by @OneOfTheWorld ?

lirantal:
@fauria I ended up fixing the issue mentioned by @OneOfTheWorld in another PR: https://github.com/meanjs/mean/pull/1388

@fauria thanks for all your help! Hope to see you again with another PR :)

modules/users/server/controllers/users/users.authentication.server.controller.js
fauria: I get @OneOfTheWorld point, the thing is, ```oauthCallback``` is actually called before ```saveOAuthUserProfile``` so even if we do:
```
user.save(function (err) {
  req.session.redirect_to = '/settings/accounts';
  return done(err, user);
});
```
the redirection will still not be applied.

lirantal: @fauria I already merged your commit with @OneOfTheWorld change.
If it's still not solving the problem, do you want to issue a PR that does?

fauria: A PR would be fine, I think it would be a good idea to review the authentication process and rework the redirections.

lirantal: We're always open for improvements. Go ahead and open a PR and let's see what's on the table.
We already had a big refactoring to the authentication and authorization code one time.

PR #1340, opened 2016-05-19
doc(getting started): Add missing word

gitname:
I think the author meant to write "using *the* `wget` command," instead of 
"using `wget` command."


lirantal:
Thanks @gitname

PR #1362, opened 2016-06-13
Fix(editorconfig): Fix file path patterns

shanavas786:
Fix Editorconfig file path patterns.

Editroconfig doesn't support patterns like `[**.js, **.json]`. Pattern `[**.{js,json}]` should be used instead as specified in [editorconfig](http://editorconfig.org/#wildcards)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,8 +11,8 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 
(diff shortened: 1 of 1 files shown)


lirantal:
very cool @shanavas786, thanks!
pr-fun

codydaig:
LGTM

PR #1366, opened 2016-06-19
Fix(config): Use single pattern for routes

shanavas786:
Use single pattern for routes in `config/assets/default.js` 

`'modules/*/server/routes/**/*.js'` instead of 
`['modules/!(core)/server/routes/**/*.js', 'modules/core/server/routes/**/*.js']`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,11 @@ _site/
 logs/
 build/
 
+# Emacs editor
+# ============
+*~
+.dir-locals.el
+
 # Sublime editor
 # ==============
 .sublime-project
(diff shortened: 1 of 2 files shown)


lirantal:
@shanavas786 build is failing.

codydaig:
I'm fairly confident that it was broken out into two because it needed to load all the module routes before loading the catch all that sends requests back to the front-end.

codydaig:
I'm fairly confident that it was broken out into two because it needed to load all the module routes before loading the catch all that sends requests back to the front-end in the core module.

https://github.com/meanjs/mean/blob/master/modules/core/server/routes/core.server.routes.js#L14

mleanos:
My understanding is the same as @codydaig. We need to load the module routes before core routes.

codydaig:
I'll close for now.

PR #1382, opened 2016-06-29
Adding node-email-verification

whitef0x0:
This is currently a WIP

### DIFF
--- a/modules/users/client/controllers/verify.client.controller.js
+++ b/modules/users/client/controllers/verify.client.controller.js
@@ -1,0 +1,105 @@
+(function () {
+  'use strict';
+
+  angular
+    .module('users')
+    .controller('VerifyController', VerifyController);
+
+  VerifyController.$inject = ['$scope', '$state', '$http', 'Authentication', '$stateParams', '$location'];
+
+  function VerifyController($scope, $state, $http, Authentication, $stateParams, $location) {
+    var vm = this;
+
+    vm.authentication = Authentication;
+
+    vm.isResetSent = false;
+    vm.credentials = {};
+    vm.error = '';
+    vm.validateVerifyToken = validateVerifyToken;
+    vm.resendVerifyEmail = resendVerifyEmail;
+
+    // Get an eventual error defined in the URL query string:
+    vm.error = $location.search().err;
+
+    // If user is signed in then redirect back home
+    if (vm.authentication.user) {
+      $location.path('/');
+    }
+
+    // Submit forgotten password account id
+    function resendVerifyEmail() {
+      $http.post('/api/auth/verify', { email: vm.credentials.email }).success(function(response) {
+        vm.success = response.message;
+        vm.credentials = null;
+        vm.isResetSent = true;
+      }).error(function(response) {
+        vm.isResetSent = false;
+        vm.credentials.email = null;
+        vm.error = response.message;
+      });
+    }
+
+    // Validate Verification Token
+    function validateVerifyToken() {
+      if ($stateParams.token) {
+
+        var token = $stateParams.token;
+        var validTokenRe = /^([A-Za-z0-9]{48})$/g;
+        if (!validTokenRe.test(token)) {
+          throw new Error('Error token: ' + token + ' is not a valid verification token');
+        }
+
+        $http.get('/api/auth/verify/' + token).success(function(response) {
+          vm.success = response.message;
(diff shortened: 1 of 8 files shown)


modules/users/client/controllers/verify.client.controller.js
```diff
@@ -0,0 +1,105 @@
...
+        }).error(function(response) {
+          vm.isResetSent = false;
+          vm.credentials.email = null;
+          vm.error = response.message;
+        });
+      }
+    }
+
+
+    function signup(isValid) {
```
mleanos: Why is signup here now? I think it would be better to refactor the existing sign-up functionality to include this extra verification step, rather than extracting it out into this new controller.

modules/users/client/views/verify/verify-account.client.view.html
```diff
@@ -0,0 +1,14 @@
+<section style="margin-top:50px;" class="row text-center" ng-init="vm.validateVerifyToken()">
+
+  <section class="row text-center" ng-if="vm.isResetSent">
+    <h3 class="col-md-12"></h3>
+    <a href="/#!/signin" class="col-md-12">Continue to Sign In</a>
+  </section>
+
+  <section class="row text-center" ng-if="!vm.isResetSent">
```
mleanos: What is "vm.isResetSent"? How does it being truthy tell us that the verification was a success?

modules/users/client/views/verify/verify-account.client.view.html
```diff
@@ -0,0 +1,14 @@
+<section style="margin-top:50px;" class="row text-center" ng-init="vm.validateVerifyToken()">
```
mleanos: Can we remove the `ng-init` attribute from this directive? I think it would be better to initiate `validateVerifyToken()` from the controller.

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -7,48 +7,159 @@ var path = require('path'),
...
+  nodemailer = require('nodemailer');
 
 // URLs for which user can't be redirected on signin
 var noReturnUrls = [
   '/authentication/signin',
   '/authentication/signup'
 ];
 
+// NEV setup and configuration ================
+var config_nev = function () {
```
mleanos: This configuration should be integrated into our existing environment configs.

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -7,48 +7,159 @@ var path = require('path'),
...
+  // Init Variables
   var user = new User(req.body);
+
+  // Add missing user fields
   user.provider = 'local';
-  user.displayName = user.firstName + ' ' + user.lastName;
+  user.username = user.email;
+
+  // Then save the temporary user
+  nev.createTempUser(user, function (err, newTempUser) {
```
mleanos: Can you explain what your "email-verification" library does when it creates a "temp user"? Does it just create a new user in the existing User collection, with a "not-verified" flag?

modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -7,48 +7,159 @@ var path = require('path'),
...
+
+  nev.generateTempUserModel(User, function(err) {
+    if (err) {
+      throw err;
+    }
+  });
+
+};
+
+config_nev();
```
mleanos: All of this "node-email-verification" functionality should be moved to it's own server controller. or server-side module (config/lib).

I think we'd be able to minimize the impact that this new feature has on the existing sign-up process.

mleanos:
Generally, I'm ok with the goal of this PR. I'd like to see this integrated into our framework in a less impacting way; making it more autonomous. This feature might be a little out of the scope of our project since there are many different ways that our users might want to implement email verification, or two-step verification. However, it might work out if it's it's own meanjs server-side module (config/lib).

If we can implement this in a way where our users could easily turn this on/off that would make it more modular & allow for easier integration of other methods of verification.
 
@lirantal WDYT?

lirantal:
Can we sum up the progress on this one? I'm seeing many comments being made on the code, which can't be merged and travis fails too so this doesn't seem in a good shape yet.

ping @whitef0x0 @mleanos

lirantal:
Closing as this is un-maintained and seems to be unsuitable for the project. See reference info in #1500

PR #1383, opened 2016-06-30
Improving README file

RoySegall:
The project now support V5.


lirantal:
Thank you, great attention to details!

codydaig:
Should we replace it with Node 6 not being supported?

lirantal:
@codydaig good point, although I think node v4 is the mainstream now and also the LTS version so possibly this is just an overflow of information to the user in the README.

codydaig:
@lirantal Fair enough

Issue #1398, opened 2016-07-14
Bug: Remove gulp-wiredep dependency


lirantal:
`nsp` reports `gulp-wiredep` as problematic due to dependencies which are out of date and are subject to RegEx attacks. That plugin is anyway another `leftpad` so we really need to take it out of our build system in gulp.

![image](https://cloud.githubusercontent.com/assets/316371/16853712/3d764a32-4a16-11e6-8197-3f31aa310b67.png)

lirantal:
A couple of suggestions for the upcoming PR:
1. Use the native `wiredep` library that is compatible with gulp's stream system
2. Use `gulp-inject`

shanavas786:
`nsp` reports lot more.

            | Regular Expression Denial of Service                  
------------|-------------------------------------------------------
 Name       | minimatch                                             
 Installed  | 2.0.10                                                
 Vulnerable | <=3.0.1                                               
 Patched    | >=3.0.2                                               
 Path       | meanjs@0.5.0-beta > forever@0.15.2 > forever-monitor… 
 More Info  | https://nodesecurity.io/advisories/118                
            |                                                       

            | Regular Expression Denial of Service                 
------------|-------------------------------------------------------
 Name       | minimatch                                             
 Installed  | 2.0.10                                                
 Vulnerable | <=3.0.1                                               
 Patched    | >=3.0.2                                               
 Path       | meanjs@0.5.0-beta > gulp-wiredep@0.0.0 > wiredep@2.2… 
 More Info  | https://nodesecurity.io/advisories/118                

            | Regular Expression Denial of Service                
------------|-------------------------------------------------------
 Name       | minimatch                                             
 Installed  | 0.3.0                                                 
 Vulnerable | <=3.0.1                                               
 Patched    | >=3.0.2                                               
 Path       | meanjs@0.5.0-beta > mocha@2.5.3 > glob@3.2.11 > mini… 
 More Info  | https://nodesecurity.io/advisories/118                

            | DoS due to excessively large websocket message 
------------|-------------------------------------------------------
 Name       | ws                                                    
 Installed  | 1.1.0                                                 
 Vulnerable | <=1.1.0                                               
 Patched    | >=1.1.1                                               
 Path       | meanjs@0.5.0-beta > socket.io@1.4.8 > engine.io@1.6.… 
 More Info  | https://nodesecurity.io/advisories/120                

            | Regular Expression Denial of Service                 
------------|-------------------------------------------------------
[...]

PR #1402, opened 2016-07-16
Feaure(grunt): Wiredep task for grunt

### LINKED ISSUE #1360: Lint error when building assets
When running ```gulp build``` or ```grunt build```, trailing commas are added to ```config/assets/production.js``` resulting in:

```
...
Running "eslint:target" (eslint) task

/Directory/config/assets/production.js
  11:65  error  Unexpected trailing comma  comma-dangle
  28:60  error  Unexpected trailing comma  comma-dangle

✖ 2 problems (2 errors, 0 warnings)

Warning: Task "eslint:target" failed. Use --force to continue.
```

Edit: Only ```gulp build``` adds those commas, ```grunt build```works fine.

shanavas786:
Update dependencies and remove trailing comma from asset files.

Fixes #1360


package.json
```diff
@@ -51,6 +51,9 @@
     "glob": "~7.0.0",
     "grunt": "~1.0.1",
     "grunt-cli": "~1.2.0",
+    "grunt-rtc": "^0.1.1",
```

gruntfile.js
```diff
@@ -74,6 +74,37 @@ module.exports = function (grunt) {
         }
       }
     },
+    wiredep: {
```
ilanbiala: Where is the wiredepnocomma task specified?

ilanbiala: Were we not using wiredep in Grunt before?

shanavas786:
No.

ilanbiala:
@lirantal I think you were working on something with grunt-wiredep, can you take a look at this too?

lirantal:
@ilanbiala indeed, thanks for the heads up.
@shanavas786 can you please take a look at my open issue related to this one? https://github.com/meanjs/mean/issues/1398. The fix should be simpler, and address gulp more specifically.

shanavas786:
@lirantal Well, this PR doesn't do much with `gulp-wiredep`. It just removes the trailing commas created by `wiredep` from asset config files which causes lint errors (#1360).

This PR primarily addresses `grunt-wiredep`. Since that also adds trailing commas, `grunt-rtc` is used to remove them.

lirantal:
@shanavas786 first of all I sincerely appreciate the effort in your PRs.
I don't want to add more dependencies to the project than what is *really* needed which is why I don't want to continue the route this PR is taking.

Both `gulp-wiredep`, and `grunt-wiredep` aren't really doing a lot, just take a look here: https://github.com/stephenplusplus/grunt-wiredep/blob/master/tasks/wiredep.js, we can easily just add `wiredep` itself and use it in either `gulp` or `grunt`.

As to the trailing commas, it sounds like we have some bug there that needs to get resolved. It doesn't make sense that we need a package to remove trailing commas.

So getting back to the point - can you update this PR to take my comments above into consideration and this way to solve both #1360 as well as #1398

shanavas786:
@lirantal sounds cool.
Had a look at `gulp-wiredep` too. [it](https://github.com/ivoputzer/gulp-wiredep/blob/master/index.js) has only two lines of code in it. So both `gulp-wiredep` and `grunt-wiredep` dependencies can be removed easily.

For handling trailing commas, it needs a little [parsing)(https://github.com/shanavas786/remove-trailing-comma/blob/master/index.js) using `esprima`. Where it would be appropriate to place that piece of code ?, in `{gulp,grunt}file.js` or in a dedicated file?

lirantal:
You totally understood my points, great :)

As for the commas, why does it even happen? do other projects need to forcely remove the trailing commas in their build process? This sounds like something we aren't doing right in the build process.

shanavas786:
@lirantal ok, let me explain, basically `wiredep` is for updating `bower` dependencies in `html` source.
It just replaces
```
<body>
  <!-- bower:js -->
  <!-- endbower -->
</body>
```
with something like
```
<body>
  <!-- bower:js -->
  <script src="bower_components/jquery/dist/jquery.js"></script>
  <script src="bower_components/angular/angular.js"></script>
  <!-- endbower -->
</body>
```
The replacement template, though it is customizable, is uniform for all dependencies (in this case `<script src="path-to-bower-dep"></script>`). Since we are not directly wiring to html source, but to a javascript array object for later use, we use a replacement template which may look like `path-to-bower-dep,` with a **trailing comma** [see gulpfile](https://github.com/meanjs/mean/blob/master/gulpfile.js#L215-231).

The trailing comma is necessary in replacement as it ends up as an array element, and since the replacement template is uniform, `wiredep` can't handle the trailing comma of last element specifically.

There comes need for a `remove-comma` task.

shanavas786:
@lirantal Just one more thought.
I think its better to wire https://github.com/meanjs/mean/blob/master/modules/core/server/views/layout.server.view.html directly. It also removes the dependency on server for a task that could be performed statically.

lirantal:
So to be honest I think that the trailing comma here is still just a symptom of the project not managing correctly the assets to wire. I suggest we do the following in this order too:
1. First let's get rid of the security issues with the gulp/grunt wiredep modules and work only with the wiredep library
2. I'm not sure the `config/assets/production.js` assets file is actually set up correctly because changing it means that we need to commit things to the repository. Let's leave this issue aside for a sec.
3. Since this is really just a linting issue how about we add a lint ignore statement on that file?

If we're still not in agreement on this then let's start with a PR on (1) to resolve the security issue and figure out how to solve (3) in a later PR.

shanavas786:
:+1: for 1.

3 would only be a temporary fix. The proper solution would be to wire the template file directly and remove the bower dependencies from asset files. That will fix 2 too.

lirantal:
Great. So let's proceed with a PR for (1) and I'll be happy to review your proposed solution for (3) in another PR.

ilanbiala:
SGTM too.

gruntfile.js
```diff
@@ -228,7 +257,6 @@ module.exports = function (grunt) {
 
   // Load NPM tasks
   require('load-grunt-tasks')(grunt);
-  grunt.loadNpmTasks('grunt-protractor-coverage');
```
lirantal: @shanavas786 why are we removing this?

lirantal:
@shanavas786 looks like we're done there? just see my comment on the protractor task removal.
can you also build it for production and test it is working well?

shanavas786:
@lirantal `load-grunt-tasks` package in `require('load-grunt-tasks')(grunt);` loads all the grunt tasks. So need not load `grunt-protractor-coverage` task explicitly. 
All the development, test and production builds were successful with each of `grunt` and `gulp` :).

lirantal:
You're right. That's great.
Super job making this pokemon happy! :)

pokemon

PR #1226, opened 2016-02-19
fix(core): fix page-title directive

itelo:

PR #1404, opened 2016-07-18
fix(core): fix page-title directive

mleanos:
This is in response to #1226 

Resolved merge conflicts.


mleanos:
I'm not sure why @itelo's commit isn't reflected after merging this, even though it's shown as the commit here.

PR #1408, opened 2016-07-23
Refactoring for express-session deprecated options

lirantal:
Updating express session name variable from `key` which is deprecated to the new variable name `name`

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -24,8 +24,8 @@ module.exports = {
   },
   // sessionSecret should be changed for security measures and concerns
   sessionSecret: process.env.SESSION_SECRET || 'MEAN',
-  // sessionKey is set to the generic sessionId key used by PHP applications
-  // for obsecurity reasons
+  // sessionKey is the session name and is set to the generic sessionId key
+  // used by PHP applications for obsecurity reasons
   sessionKey: 'sessionId',
   sessionCollection: 'sessions',
   // Lusca config
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -24,8 +24,7 @@ module.exports = {
   },
   // sessionSecret should be changed for security measures and concerns
   sessionSecret: process.env.SESSION_SECRET || 'MEAN',
-  // sessionKey is the session name and is set to the generic sessionId key
-  // used by PHP applications for obsecurity reasons
+  // sessionKey is the cookie session name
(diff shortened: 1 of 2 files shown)


config/env/default.js
```diff
@@ -24,8 +24,8 @@ module.exports = {
   },
   // sessionSecret should be changed for security measures and concerns
   sessionSecret: process.env.SESSION_SECRET || 'MEAN',
-  // sessionKey is set to the generic sessionId key used by PHP applications
-  // for obsecurity reasons
+  // sessionKey is the session name and is set to the generic sessionId key
+  // used by PHP applications for obsecurity reasons
```
simison: "PHP applications"?

lirantal: Yeah. I also put it there before too, just organized the comment.
Do you find it odd? do you understand what this actually says/means?

simison: I'm totally puzzled by it. :-) This config var sets the name for the session cookie, right?

So I suppose it's not referring to PHP, the language?

Also first time seeing used "obsecurity" when saying "security by obscurity", but I like it!

codydaig:
@lirantal Your commit isn't authored correctly, it isn't linking to your github account.

config/env/default.js
ilanbiala: This comment is a little confusing

hyperreality: I understand the point - it helps to give as little information about your framework as possible in order to minimize automated attacks. In this case the objective appears to be to save cookies that look like cookies saved by PHP frameworks. However, after a quick Google I couldn't find a documented practice of using "sessionId" as a default name for PHP cookies. Looking through my own cookies, every website seemed to use a different name for session id anyway. 

I don't think this should be a security concern and perhaps the comment should be removed to prevent confusion.

lirantal: The actual use of PHP cookie name for session IDs is 'PHPSESSIONID' but I thought putting that might be confusing. Alternative of 'JSESSIONID' as well so maybe we should just take out the comment and leave the generic cookie name as is.
I'll commit a change.

lirantal:
@codydaig weird, but thanks for the update. I'll check what's up with this new environment setup I have.

codydaig:
LGTM

@lirantal Your latest commit looks good!

lirantal:
Thanks!

PR #1409, opened 2016-07-23
feat(forever): removing forever dependency

lirantal:
cleaning up project dependencies by removing the forever package which is only really used by Heroku's
procfile to start up the node service


codydaig:
LGTM

PR #1413, opened 2016-07-27
Updating environment variable name for mongodb with the new one for Heroku

### LINKED ISSUE #1411: Unable to deploy in Heroku
I am trying to deploy my application in Heroku, but when i open the link it say application error.

Steps Followed
```
$ grunt build --force
$ heroku create
$ git status
$ git add -A
$ git commit -m “ready to deploy”
$ heroku config:set NODE_ENV=production
$ heroku addons:create mongolab
$ git push heroku master
```

lirantal:
Updating environment variable name for mongodb with the new one for Heroku (according to https://devcenter.heroku.com/articles/mongolab)
Fixes #1411

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -4,7 +4,7 @@ var defaultEnvConfig = require('./default');
 
 module.exports = {
   db: {
-    uri: process.env.MONGOHQ_URL || process.env.MONGOLAB_URI || 'mongodb://' + (process.env.DB_1_PORT_27017_TCP_ADDR || 'localhost') + '/mean-dev',
+    uri: process.env.MONGOHQ_URL || process.env.MONGODB_URI || 'mongodb://' + (process.env.DB_1_PORT_27017_TCP_ADDR || 'localhost') + '/mean-dev',
     options: {
       user: '',
       pass: ''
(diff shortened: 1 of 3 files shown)


codydaig:
LGTM

ilanbiala:
LGTM.

mleanos:
:+1:

mleanos:
I looked into this a little further, and realized that this might not be a necessary change. Just as long as the name used on Heroku's environment config settings matches up with what our application uses, then there shouldn't be a problem. If someone is using Heroku's MongoDB plugin, then there may be a mismatch if they don't set an environment variable that matches the name that our framework uses.

My concern here is that we can't account for every variation of naming convention that's used by all possible deployment environments. For instance, what if AWS uses a different name for their MongoDB connection strings?

To me the name **we** use is irrelevant, and it's up to the developer to configure their app to accommodate whichever environment they are deploying to; to ensure their MEANJS app matches up with their deployment environment.

lirantal:
@mleanos I understand what you're trying to say but we aren't accommodating every possible config, but rather just herokus, and they indeed changed their config variable from the old one to the new one so I don't see an issue here. (I also validated with other repositories performed this change)

Unless you really think there's something else to consider here I don't see why not to make this change.

mleanos:
@lirantal I'm fine with this change. I just wanted to express my point of view.  To me there isn't an impact if we do, or do not, make this change. So lets roll with this to keep our naming consistent with Heroku's.

lirantal:
Cool, merging.

PR #1418, opened 2016-08-06
fix(user): fix changeProfilePicture

### LINKED ISSUE #1415: Unlink old(previous) profile images
Currently old profile images are not unlinked upon change and remain on disk. If a user changes his/her profile image frequently that can lead to disk space leakage.

aanev:
* use async to simplify callbacks

* use fs.unlink to delete old picture once the user profile is updated

Fixes #1415


lirantal:
Thanks @aanev 
Can you replace async with a promise-based approach?

ilanbiala:
Yes, promises will make this quite a bit cleaner.

PR #1419, opened 2016-08-07
fix(chat): fixed profile image not displaying properly on disconnect

9ac345a5509a:
In the chat example, the profileImageURL is not sent to clients when a user disconnects.

Before:

before picture

After:

after picture


mleanos:
LGTM

PR #1421, opened 2016-08-07
Fix(users): Don't update secure profile fields

### LINKED ISSUE #1420: Secure fields can be updated throgth profile udpate
Since profile update only filters out `_id` and `roles` fields (see [user.profile.server.controller.js](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.profile.server.controller.js#L22-L26)), other secure fields like `password, salt ..etc` can be updated by including those fields in request body. It is a security flaw.

shanavas786:
Avoid updating secure fields as password, salt ..etc through
user profile update.

Fixes #1420

### DIFF
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -12,21 +12,18 @@ var _ = require('lodash'),
   config = require(path.resolve('./config/config')),
   User = mongoose.model('User');
 
+var whitelistedFields = ['firstName', 'lastName', 'email', 'username'];
+
 /**
  * Update user details
  */
 exports.update = function (req, res) {
   // Init Variables
   var user = req.user;
-  var secureFields = ['firstName', 'lastName', 'email', 'username'];
 
   if (user) {
-    // For security measurements update only secure fields
-    secureFields.forEach(function (field) {
-      if (req.body.hasOwnProperty(field)) {
-        user[field] = req.body[field];
-      }
-    });
+    // Update whitelisted fields only
+    user = _.extend(user, _.pick(req.body, whitelistedFields));
 
     user.updated = Date.now();
     user.displayName = user.firstName + ' ' + user.lastName;
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -18,16 +18,16 @@ var _ = require('lodash'),
 exports.update = function (req, res) {
   // Init Variables
   var user = req.user;
-
-  // For security measurement we remove the roles from the req.body object
-  delete req.body.roles;
-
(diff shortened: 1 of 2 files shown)


modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -18,16 +18,16 @@ var _ = require('lodash'),
 exports.update = function (req, res) {
   // Init Variables
   var user = req.user;
-
-  // For security measurement we remove the roles from the req.body object
-  delete req.body.roles;
-
-  // For security measurement do not use _id from the req.body object
-  delete req.body._id;
+  var secureFields = ['firstName', 'lastName', 'email', 'username'];
```
mleanos: The naming here is a little confusing. It sounds like these are the fields that are **not** allowed.

mleanos:
Thank you for taking the initiative & submitting this PR. And for the added test :)

I'm a little unsure of this solution though. I think a better solution would be to maintain two lists at the top level of this server controller.

1) Whitelist - Fields that can be updated and/or be sent back to the client.
2) Blacklist - Fields that must be hidden from the clients and/or cannot be updated.

We can keep the line with `user = _.extend(user, req.body);`, and then just iterate through the Blacklist & remove those fields from the User object before we delete.

WDYT?

modules/users/server/controllers/users/users.profile.server.controller.js
shanavas786: Yeas :). whitelist would be far better.

shanavas786:
@mleanos I took the black list approach because if request contains some invalid fields like `invalidkey:"invalidvalue"`, `extend` will assign that also to user object. 
```
x = { a: "a" };
_.extend(x, { b: "b" };
```
makes `x` `{ a: "a", b: "b" })`.
But I could not observe such alteration in user object. I guess mongoose might be handling it (any clarification is appreciated :)).

mleanos:
What do you think about having a function defined in the server controller, that "cleans" the user object coming in from `req.body`? It could return an object that only includes the white-listed field.

You may end up with something like this for when we receive a user object from the request:
```javascript
// only extend with fields we want to accept from the request
user = _.extend(user, whitelistedUser(req.body));
```
and for returning to the client:
```javascript
// only send fields that are safe to be sent back to the client
res.json(sanitizedUser(user));
```

Thoughts on this approach?

My understanding is that Mongoose won't update fields that aren't present in the model definition. Unless, it's an object field (dynamic?) with no strict definitions.

shanavas786:
That looks better. 
I'm a little stretched out now. Will update accordingly when I get some spare time.

mleanos:
I think this is good now. Thanks.

@meanjs/contributors Any feedback?

lirantal:
Looks good, thanks @mleanos and @shanavas786

PR #1425, opened 2016-08-11
fix(user): fix changeProfilePicture

### LINKED ISSUE #1415: Unlink old(previous) profile images

aanev:
fix(user): fix changeProfilePicture

* use promises to simplify callbacks

* use fs.unlink to delete old picture once the profile is updated

Fixes #1415

### DIFF
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -60,13 +60,13 @@ exports.changeProfilePicture = function (req, res) {
   var user = req.user;
   var upload = multer(config.uploads.profileUpload).single('newProfilePicture');
   var profileUploadFileFilter = require(path.resolve('./config/lib/multer')).profileUploadFileFilter;
-  var tmpImageUrl;
+  var existingImageUrl;
 
   // Filtering to upload only images
   upload.fileFilter = profileUploadFileFilter;
-  
+
   if (user) {
-    tmpImageUrl = user.profileImageURL;
+    existingImageUrl = user.profileImageURL;
     uploadImage()
       .then(updateUser)
       .then(deleteOldImage)
@@ -87,6 +87,7 @@ exports.changeProfilePicture = function (req, res) {
     return new Promise(function (resolve, reject) {
       upload(req, res, function (uploadError) {
         if (uploadError) {
+          console.log(uploadError);
           reject({
             message: 'Error occurred while uploading profile picture'
           });
@@ -112,11 +113,12 @@ exports.changeProfilePicture = function (req, res) {
 
   function deleteOldImage () {
     return new Promise(function (resolve, reject) {
-      if (tmpImageUrl !== User.schema.path('profileImageURL').defaultValue) {
-        fs.unlink(tmpImageUrl, function (unlinkError) {
+      if (existingImageUrl !== User.schema.path('profileImageURL').defaultValue) {
+        fs.unlink(existingImageUrl, function (unlinkError) {
           if (unlinkError) {
+            console.log(unlinkError);
             reject({
-              message: 'Error occurred while uploading profile picture'
+              message: 'Error occurred while deleting old profile picture'
             });
           } else {
             resolve();
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -87,10 +87,7 @@ exports.changeProfilePicture = function (req, res) {
     return new Promise(function (resolve, reject) {
       upload(req, res, function (uploadError) {
         if (uploadError) {
-          console.log(uploadError);
-          reject({
-            message: 'Error occurred while uploading profile picture'
-          });
+          reject(errorHandler.getErrorMessage(uploadError));
         } else {
           resolve();
         }
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -60,39 +60,83 @@ exports.changeProfilePicture = function (req, res) {
   var user = req.user;
   var upload = multer(config.uploads.profileUpload).single('newProfilePicture');
   var profileUploadFileFilter = require(path.resolve('./config/lib/multer')).profileUploadFileFilter;
+  var tmpImageUrl;
 
   // Filtering to upload only images
   upload.fileFilter = profileUploadFileFilter;
-
+  
   if (user) {
-    upload(req, res, function (uploadError) {
-      if (uploadError) {
-        return res.status(400).send({
-          message: 'Error occurred while uploading profile picture'
-        });
-      } else {
-        user.profileImageURL = config.uploads.profileUpload.dest + req.file.filename;
-
-        user.save(function (saveError) {
-          if (saveError) {
-            return res.status(400).send({
-              message: errorHandler.getErrorMessage(saveError)
-            });
-          } else {
-            req.login(user, function (err) {
-              if (err) {
-                res.status(400).send(err);
-              } else {
-                res.json(user);
-              }
-            });
-          }
-        });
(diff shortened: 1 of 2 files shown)


modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -60,39 +60,83 @@ exports.changeProfilePicture = function (req, res) {
   var user = req.user;
   var upload = multer(config.uploads.profileUpload).single('newProfilePicture');
   var profileUploadFileFilter = require(path.resolve('./config/lib/multer')).profileUploadFileFilter;
+  var tmpImageUrl;
 
   // Filtering to upload only images
   upload.fileFilter = profileUploadFileFilter;
-
+
```
mleanos: Remove spaces here.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -60,39 +60,83 @@ exports.changeProfilePicture = function (req, res) {
...
-        });
-      } else {
-        user.profileImageURL = config.uploads.profileUpload.dest + req.file.filename;
+    tmpImageUrl = user.profileImageURL;
+    uploadImage()
+      .then(updateUser)
+      .then(deleteOldImage)
+      .then(login)
+      .then(function () {
+        res.json(user);
```
mleanos: Will the user sent back with the response here be the updated user (after the update method has completed)? Even though `user` is defined within the same scope, and is available in the `updateUser` closure, it's never updated; see my comment below in the `updateUser` method.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -60,39 +60,83 @@ exports.changeProfilePicture = function (req, res) {
...
-          if (saveError) {
-            return res.status(400).send({
-              message: errorHandler.getErrorMessage(saveError)
+  function deleteOldImage () {
+    return new Promise(function (resolve, reject) {
+      if (tmpImageUrl !== User.schema.path('profileImageURL').defaultValue) {
+        fs.unlink(tmpImageUrl, function (unlinkError) {
+          if (unlinkError) {
+            reject({
+              message: 'Error occurred while uploading profile picture'
```
mleanos: Can the error message here be more specific? Perhaps stating that the error occurred while deleting the file?

mleanos:
Overall, this looks good to me. Thanks for the PR!

I left a few comments, but nothing major. Good work on this :)

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -60,39 +60,83 @@ exports.changeProfilePicture = function (req, res) {
   var user = req.user;
   var upload = multer(config.uploads.profileUpload).single('newProfilePicture');
   var profileUploadFileFilter = require(path.resolve('./config/lib/multer')).profileUploadFileFilter;
+  var tmpImageUrl;
```
mleanos: Naming? Could it be `existingImageUrl`, or something similar?

modules/users/server/controllers/users/users.profile.server.controller.js
mleanos: Also, what info does `unlinkError` contain? Would it be useful to provide that with the reject?

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -60,39 +60,83 @@ exports.changeProfilePicture = function (req, res) {
...
+      message: 'User is not signed in'
+    });
+  }
+
+  function uploadImage () {
+    return new Promise(function (resolve, reject) {
+      upload(req, res, function (uploadError) {
+        if (uploadError) {
+          reject({
+            message: 'Error occurred while uploading profile picture'
```
mleanos: Similar to my comment below. Can we use any of the info from `uploadError` to clearly state what went wrong?

modules/users/server/controllers/users/users.profile.server.controller.js
aanev: I decided to not provide unlinkError info to the end user because that typically would be an administration issue (e.g. deleted file).

modules/users/server/controllers/users/users.profile.server.controller.js
aanev: I absolutely agree that these methods need to be extracted. However, that would mean major refactoring so I decided in favour of taking smaller steps.

modules/users/server/controllers/users/users.profile.server.controller.js
aanev: IMHO I'd rather not pass multer errors to the end user.

simison: These might come handy: 

```javascript
// An error occurred when uploading
// See Multer default error codes:
// @link https://github.com/expressjs/multer/blob/master/lib/make-error.js
if (err) {
  var errorMessage,
      errorStatus;
      
  if (err.code && err.code === 'UNSUPPORTED_MEDIA_TYPE') {
    // Unsupported media type -error
    // This error is generated from ./config/lib/multer.js
    errorMessage = errorHandler.getErrorMessageByKey('unsupported-media-type');
    errorStatus = 415;
  } else if (err.code && err.code === 'LIMIT_FILE_SIZE') {
    // Too big file
    // 413: "Request Entity Too Large"
    errorMessage = 'Image too big. Please maximum ' + (config.maxUploadSize / (1024 * 1024)).toFixed(2) + ' Mb files.';
    errorStatus = 413;
  } else if (err.code && err.code === 'LIMIT_UNEXPECTED_FILE') {
    // Field doesn't exist -error
    errorMessage = 'Missing `avatar` field from the API call.';
    errorStatus = 400;
  } else {
    // Any other error
    errorMessage = errorHandler.getErrorMessageByKey('default');
    errorStatus = 400;
  }
  
  return res.status(errorStatus).send({
    message: errorMessage
  });
}
```

via https://github.com/Trustroots/trustroots/blob/master/modules/users/server/controllers/users/users.profile.server.controller.js#L107-L137

mleanos: Thank you @simison! This looks very handy.

@aanev Could something like this be incorporated into our [existing error handling](https://github.com/meanjs/mean/blob/master/modules/core/server/controllers/errors.server.controller.js) server-side module?

These codes could be useful for the client application.

mleanos:
@aanev Have you seen @simison's [comment](https://github.com/meanjs/mean/pull/1425#discussion_r74685783) above? I think these codes could be useful to send back to the client application. I've also commented on it above.

WDYT?

modules/users/server/controllers/users/users.profile.server.controller.js
aanev: It looks good. Definitely better that the current message.

farhatmo:
Hello, 
I devlopped a version for the ProfilePicture upload  with storage in MongoDB instead of
saving on the FS. @mleanos do you think that can be the subject of new PR ?

lirantal:
@farhatmo  I think it's a bit of an overkill...

mleanos:
@farhatmo I agree that it's a bit overkill. 

In one of my projects I was using Heroku to host, and ran into the issue (early on in the project) with Heroku overwriting your entire app directory on each deployment/remote-build; the profile images were cleared on each deployment. I decided to save the profile images in the database, but eventually realized it was a waste of db storage. I moved to using AWS S3, and have been happy using that solution ever since. I think the decision to store binary data in MongoDB is something that each developer must answer for themselves; IMO, it should be avoided when possible.

GridFS is a neat feature of MongoDB, but you should tread lightly when implementing it in a large scale application. Hence, it's probably not a suitable solution for MEANJS as a framework.

PR #1426, opened 2016-08-11
Fix(server): Remove unused loadModels()

simison:
Remove unused `loadModels()` from app.js, as it doesn't seem to be used anywhere.

— or am I missing something?

It was added here: https://github.com/meanjs/mean/commit/edb62344bc4af3cb28c3469dc7463c13a2fc76a1


lirantal:
I added it as a helper function back then, we can remove it.

simison:
👍 Thanks for explaining & merging.

PR #1427, opened 2016-08-12
feat(users):  Update social login design

benobab:
## New design for social links

- responsive images 
- grid for mobile layout as well

### DIFF
--- a/modules/users/client/css/users.css
+++ b/modules/users/client/css/users.css
@@ -34,3 +34,14 @@
   min-height: 150px;
   max-height: 150px;
 }
+
+.social {
+   -webkit-transition-duration: 0.4s;
+   -moz-transition-duration: 0.4s;
+   -o-transition-duration: 0.4s;
+   transition-duration: 0.4s;
+}
+
+.social:hover {
+    opacity: 0.65;
+}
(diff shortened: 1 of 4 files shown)


mleanos:
LGTM. I've tested this as well.

Thanks @benobab!

hyperreality:
One small gripe I have with this commit is that it works great if you want to authenticate with all 6 networks, but in most cases that won't be the case, and after removing a button or two it's hard to get a nice centred design.

Furthermore, I tried shrinking the page in Firefox and below a certain size it's not possible to click the buttons at all.

Also why we're at it why don't we make the buttons interactive? Add them to the social class and put this in core.css:

```
.social {
   -webkit-transition-duration: 0.4s;
   -moz-transition-duration: 0.4s;
   -o-transition-duration: 0.4s;
   transition-duration: 0.4s;
}

.social:hover {
    opacity: 0.65;
}
```

benobab:
@hyperreality I've updated the code with your css and fixed the bug where you can't click the buttons at all (xs rather than md for bootstrap).

lirantal:
Thanks a lot @benobab, I fixed the conflict and added to your commits a new PR which can be merged at: https://github.com/meanjs/mean/pull/1450

lirantal:
closed through PR #1450

Issue #27, opened 2014-04-21
enhancement: i18n support


frederik:
I currently have the need to localize a lot of core to display authentication information (in German).

To do that I added angular-translate and a local core.client.translation.js with some basics (these files should go into json files one day). I'll have to change all templates (especially the header to now use the translate directive a la {{ 'SIGNOUT' | translate }}).

Any ideas on how to incorporate i18n on the server side so we could re-use translations? I'd be willing to provide a fork with the changed the templates, German translations etc. I'd wait for 0.3 to ship though, and then fork the new development branch.

afucher:
Hi, is there someone working on it?? So I could help??

codydaig:
@afucher I'm assuming since this issue has been stale for over a year that it's all yours in you want to submit a PR.

afucher:
@codydaig I'm starting to develop my first app, using MEAN stack. But, I need to use i18n.. so, if I do any implementation, I'll submit a PR..  :)

lirantal:
@afucher great, I'm closing this issue and looking forward for your i18n PR :)

pun4drunk:
Hi guys! Any progress with i18n?

afucher:
@pun4drunk I'm not working with MEAN, so I don't have news....

ilanbiala:
@pun4drunk we probably won't have i18n for 0.5.0, so we'll look into it for the next few versions.

pun4drunk:
can I make a PR? :)

PR #1432, opened 2016-08-17
feat(i18n): i18n support 

pun4drunk:
feat(i18n): i18n support for MEAN.js

Using angular-translate: https://github.com/angular-translate/angular-translate
Using angular-dynamic-locale: https://github.com/lgalfaso/angular-dynamic-locale
Using i18n-node for backend: https://github.com/mashpie/i18n-node

Features: #27


pun4drunk:
So far I experience some karma:unit tests failing due to additional GET request. Can somebody advise an easy way to fix this?

simison:
So I guess the question is, should strings be translated from English, or from keys. With keys there's much less work maintaining multiple language files when something changes.

pun4drunk:
Yes, I see both approaches are good enough to work together:
English strings in templates are easier to locate and maintain
Translation IDs are good when it comes to translate the whole paragraphs with hyperlinks and other markup

pun4drunk:
Guys, what about this?

So far I experience some karma:unit tests failing due to additional GET request. Can somebody advise an easy way to fix this?

pun4drunk:
We can add a new request handler i.e
<code>$httpBackend.whenGET(/i18n\/\.*/).respond(200, '');</code>
into each controller test file.
Would it be an appropriate solution?

mleanos:
@pun4drunk Yes, adding the request handler to the client-side test files is OK. We're already doing that in some of the Karma tests.

mleanos:
Overall, I welcome the addition of localization & cultures into this framework. In addition, the use of angular-translate seems like the correct choice. With that said, I think this PR can be slimmed down to merely add support for i18n. At most, I think we can show how to implement this using **just** the Chat example. I don't see a need to implement this into every aspect of our front-end. Most users of this framework won't need this feature out-of-the-box.

I don't want to end up maintaining this implementation throughout our client-side controllers, views, directives, etc. This will end up being a maintenance nightmare, as our framework changes & expands. Whoever needs this feature will be able to easily add it since the framework already will have standards, and support in place.

@meanjs/contributors @simison Any thoughts?

public/i18n/en-us.json
```diff
@@ -0,0 +1,115 @@
+{
```
mleanos: The files in *public/i18n/* are auto-generated using the Gulp task, correct? If that's the case, we should add this folder to our *.gitignore*.

pun4drunk: Gulp task creates the files themselves but sure does not fill any translations for you. You must fill them manually

pun4drunk:
@mleanos I find your opinion very reasonable. Although perhaps chat module is not so suitable to demonstrate i18n support - there is not so much to translate there and there is at least header that will stay in english anyway. Maybe we could have a new module instead with just a single page (and language dropdown) to show i18n support in few markup examples

public/i18n/en-us.json
mleanos: What's the difference between these files and the language *.json* files in the root of */i18n*? Why don't the translations in those files exist here?

mleanos:
@pun4drunk The precise reason I think that we should implement i18n in **just** the Chat module, is because there wouldn't be much to translate. Once a user of this framework decides that they want to take full advantage of this localization implementation, they can use our minimal example shown in the Chat module to implement it elsewhere in their application. This approach would help us keep i18n lightweight & have a very low impact on our framework.

public/i18n/en-us.json
pun4drunk: public/i18n holds client translations and is handled by a gulp task
/i18n holds server translations and is handled by i18n-node during invocation

pun4drunk:
Issues I've faced implementing i18n were as follows:

**Main Menu**
To translate main menu, you should append translate filter to the labels. You should also use translate service in each module's menu config to ensure translations where generated.

_Using chat module as an example can not demonstrate how to implement core translations._

**Dynamic Locale**
With articles module translations (or any other future CRUD that handles any kind of dates with it), comes the necessity to auto-translate dates/currencies etc. This is where angular-dynamic-locale comes in handy.

_Using chat module as an example can not demonstrate how to do this since it has no dates._

**Server translations**
To translate messages coming from server (i.e. restricted access message in CRUD module) i18n object is being registered as a global variable and used for each message.

_This is an only place that more or less related to chat module, because it's "Is now connected" message is coming from server._

**Pagination** (and probably more bootstrap widgets)
To translate pagination I needed to overwrite default template, since it used one time bindings. This issue is reported but authors don't think it [should be resolved anyway](https://github.com/angular-ui/bootstrap/issues/5460).

_Using chat module as an example can not show this since it has no pagination._

Now you can imagine how much work you still need to do to implement i18n into your project using obviously poor example from chat module.

Finally, if we implement i18n globally but will only use it in one module, this will not save us from an additional GET request so it still should be handled in tests.

Good thing is that I don't see so much trouble supporting i18n (all you need to do is to call translate filter/service/method in the right place and if you don't need translations you just don't use it) and tests do not depend on this.

I've even added e2e test for language dropdown. 

Maybe you could think about it like a side branch I can provide support to, so it can be used with [MeanJS Generator](https://github.com/pun4drunk/generator-meanjs)? I've almost finished crud-module translations so that in sum this could seriously increase your (our) community.

mleanos:
@meanjs/contributors Would you like to review this as well? IMO, this introduces too many changes for the value it adds to the framework. It seems that most users wouldn't need this functionality right out of the box. I don't feel comfortable with the complexity this implementation is adding. Also, see my comments above.

I'm still open to adding this, but I'd like to see it done in a more configurable manner, and seamlessly fit into the framework.

lirantal:
My personal take is that this looks quite a piece of code to keep maintaining.
Who knows, we might be moving out to ng2/react in the near future :)

My other take on this is that if this PR adds a complete, solid i18n capability, with tests, and @pun4drunk can hang out with us on MEAN.JS to further maintain it then I'm ok with merging this in.

ilanbiala:
Is it possible to have all the config, but just have this set up for one module just to show how it works?

mleanos:
I think this will end up being too much of a maintenance nightmare. And the implementation seems to be one that you would see in an instance of this framework where i18n integration is a requirement.

I like the idea of this being a part of the generator project. @ilanbiala Would that be doable? Also, I did suggest implementing this for just the Chat module. Since the Chat module has very little content. Even though, I'm not a fan of the inline hooks to i18n & front-end views.

Overall, this implementation seems unnatural to a framework like ours. I like what this is attempting to solve though. Can we find a better library/implementation?

lirantal:
@pun4drunk what do you think about doing this as part of the generator maybe?

simison:
Some food for thought — Angular 2 will have i18n support:

>Internationalization (i18n) support. Add an i18n attribute, then use the ng-xi18n tool to extract strings to a standard XLIFF file. See [this example](https://github.com/StephenFluin/i18n-sample) to try it out.

via https://angularjs.blogspot.lt/2016/09/angular-2-rc6_1.html

lirantal:
Even stronger point not to merge in yet.
Thanks Mikael

pun4drunk:
@lirantal, I'm ok about making this a part of a generator. Please contact me to discuss this. My skype is vladwaxme. Thank you

lirantal:
Vlad, @codydaig and @ilanbiala are maintaining the generator, let's steer this discussion over in the generator repo and see if it makes any sense.

pun4drunk:
Fine! I just should note that in the generator they have a list of versions to pick from:
````
var versions = {

  'master': 'master',

  '0.4.0': 'v0.4.0',

  '0.4.1': 'v0.4.1',

  '0.4.2': 'v0.4.2'

};
````
which is basically a git entities taken from this repo. So anyway if we make i18n-supported version a part of a generator, it should somehow get into this repo as a branch

trendzetter:
I don't think there is any path ready to migrate to angular2. This could be very useful for me too.

pun4drunk:
Friends, please keep in mind that migration to angular2 itself is not enough to provide i18n support since MEAN.js also uses messages coming from server side (mentioned above in details)

PR #1449, opened 2016-08-26
Support gulp for bluemix deployment

### LINKED ISSUE #1441: Deployment to IBM Bluemix fails
I can upload the code from Git terminal, but after "Uploading droplet" message, no instance is running.  I can manually run the instance from Bluemix website, but after clicking the URL, it states:  "Requested route does not exist."

I was able to deploy my code to Heroku after looking at the closed issue, which suggested to add gulp dependencies from devDependencies to Dependencies.  However, the same code does not work for Bluemix.

I was able to upload mean.js 0.4.2 to Bluemix, so the problem is with mean.js 0.5.0-beta.

lirantal:
fix(bluemix): refactoring bluemix config to support gulp build process instead of grunt
fixes #1441

### DIFF
--- a/.bluemix/pipeline.yml
+++ b/.bluemix/pipeline.yml
@@ -10,20 +10,14 @@ stages:
   - name: meanjs-build
     type: builder
     artifact_dir: ./
-    build_type: grunt
+    build_type: gulp
     script: |-
       #!/bin/bash
       # Set Node version to 0.12
       export PATH=/opt/IBM/node-v0.12/bin:$PATH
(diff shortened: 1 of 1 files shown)


codydaig:
LGTM as far as I can tell. I cannot test as my Bluemix trial expired.

lirantal:
@codydaig yep, I have no idea how better to support Brian there on that issue so we might end up closing this PR if we can't help him.

lirantal:
As we can't make any progress in testing or fixing up the bluemix deployment I am closing this PR.

Issue #1397, opened 2016-07-14
noscript support


WanderZ:
I would like to clarify where would should the `<noscript>` tag/text/content be at?

e.g. `<noscript>Please enable javascript to use this application</noscript>`

Problem is that the entire body is of class `ng-cloak` and I didn't want users to be looking at a completely blank screen.

Sorry if this question is unrelated or irrelevant to the development

lirantal:
Just put it before or after the body?

WanderZ:
Sorry, neither worked. 

i tried both
```<noscript>Message here</noscript>```

and 

```<div id="noscript">Message here</div>```

**Comment by shanavas786** [comment]
`ng-cloak` uses css property `display` to hide elements. https://github.com/angular/angular.js/blob/master/css/angular.css#L3-L7
Can't you override `display` property for id `noscript` with some thing like this ?
```
#noscript {
  display: block !important;
}
```

**Comment by WanderZ** [comment]
Pardon for the delay, the CSS property did not help when JavaScript is disabled.

It will, however, create a flicker when JavaScript is not disabled. (on first page load)

**Comment by hyperreality** [comment]
Perhaps what is happening is that the text is appearing behind the blank navbar.

If I put the following at the end of layout.server.view.html, after the final </html> tag, I can see it just below the navbar:
   <noscript style="display: block !important;">Message here</noscript>

**Comment by WanderZ** [comment]
Sorry, @hyperreality could you elaborate more on the final tag? 

I'm also modifying layout.server.view.html, and I tried to include it within the `<head></head>`, within `<body></body>`.

I tried placing it after the final `</html>` as well, it didn't appear for me. I tried placing it after the final `</body>` as well.

**Comment by hyperreality** [comment]
Simply putting 

`<noscript>Please enable Javascript to use this application</noscript>`

just before the final </body> tag in layout.server.view.html worked for me (using Firefox with Noscript plugin). It also shows up in console browsers like links.

Which browser are you using? Have you modified the navbar from the default app?

**Comment by WanderZ** [comment]
Sorry, I meant to ask if you could tell me which tag exactly do you mean, because I don't know which tag you are referring to via `final`.

Currently i'm using chrome, and using the no-script suite lite.

You can see my attempt here: 
https://github.com/WanderZ/mean/blob/master/modules/core/server/views/layout.server.view.html

**Comment by shanavas786** [comment]
@WanderZ try this.

```
.....
  <noscript>
    <style>
[...]

PR #1454, opened 2016-08-27
Added noscript tag to layout

hyperreality:
feat(core): Add noscript tag to layout

See #1397 

It's true that not many users block JavaScript these days, but it's certainly 
nice (and costs nothing) to confront those users with a friendly message 
rather than just a blank screen.


lirantal:
Thanks @hyperreality

PR #1457, opened 2016-08-28
feat(core): disable Angular debug data in production

simison:
Disable Angular debug data in production for a significant performance boost.

Passes environment variable from template to app config and from there to Angular bootstrap config.

https://docs.angularjs.org/guide/production#disabling-debug-data

See #1294


lirantal:
Keep it up Mikael!

Issue #1455, opened 2016-08-27
Gulp test silently fails if mocha encounters fatal error


hyperreality:
Accidentally ran `gulp test` using Node 6.2, an unsupported version, and stumbled across a little error. Here is the offending code: https://github.com/meanjs/mean/blob/730cca7/gulpfile.js#L310

If Mocha encounters a fatal error, `end` is never returned, no error is logged to the terminal and Gulp just hangs. Grunt encounters no such problem: the unsupported version error is correctly reported. Perhaps errors should be logged to the console as they occur? Paging @ilanbiala

simison:
Yup, just experienced this myself.

Error was: `Unsupported Node version: 6.3.1`

Added `console.error(err);` to Gulp's [mocha task][1] to see this.

Time to upgrade to [Mocha v3][2]?

[1]: https://github.com/meanjs/mean/blob/master/gulpfile.js#L310-L313
[2]: https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#300--2016-07-31

simison:
Aight, fixed by running `npm install mock-fs@~3.11.0 --save-dev`

I'll make a PR with this + mocha v3. Someone else can tackle this original issue with `.on('error'...` causing problems. It was last time modified at #453

simison:
BTW last time this happened: https://github.com/mochajs/mocha/issues/2102

hyperreality:
@simison Great stuff. Pretty surprised that this bug has been around for so long.

As it happens pretty much everything else seemed to work under Node 6.2.2, I wonder if we can move to supporting it soon.

PR #1458, opened 2016-08-28
fix(core): Fixes issues with tests and Node v6

simison:
Fixes issues with Mocha tests and Node v6.

Basically updating `mock-fs` would fix the issue, but updating Mocha to v3 as well while I’m at it…

Tweaked tests to pass also under Mocha v3 and Node v6.

Related to #1455


lirantal:
Very cool, thanks @simison!

lirantal:
@simison noticed that I closed #1455 or should fixes still be applied for it?

simison:
@lirantal this PR just fixes error we stumbled upon while looking into the issue, but doesn't fix the initial problem he/she was reporting: `.end()` doesn't get called if `.error()` happens.

...at least we now know how to cause mocha error easily. ;-)

PR #1272, opened 2016-03-18
fix(users) Fix bug file mime-type validate does not work 

trendzetter:
Bad use of file type valiator for multer.fileFilter
I have been trying some more things and indeed it uploads a text file if you rename it to jpg. So it validates only the extension on the client and not the mime-type in the multer on the server. After applying the suggested change at least config/env/multer.js is called, but it is still not filtering correctly for me: all files appear to be image/jpeg when I add   console.log(file.mimetype); on line 4.

Fixes #1270

PR #1465, opened 2016-08-29
fix(users) MIME-type checking fixed on both client and server-side

hyperreality:
fix(users) Client-side MIME-type checking for picture upload

@mleanos pointed out that the MIME-type checking for profile pictures
wasn't working on the client-side as expected, so I fixed it to display
a warning message if the user attempts to upload a different kind of file.

In the meantime I realised that MIME-type checking was broken on the
server-side too, as noted in #1272, so I applied the necessary fixes and 
added a test that attempts to upload a JS file as a profile picture.

As @simison said on that thread, the multer isn't infallible and magic bytes
checking would be best, so perhaps that is the next stage.


hyperreality:
Note that this doesn't fix the issue reported in #1272, it is still possible to rename any file to a .jpg for instance and upload it through the API route. To fix that we need to apply a fix like @simison implemented on his website - temporarily upload the file, magic-byte check it, then discard if it doesn't match.

simison:
@hyperreality yeah I'm pretty thin on time. Ping me if ya need any help (see [my chat contact infos](http://www.mikaelkorpela.fi/))

lirantal:
@hyperreality the temp upload and mime type checking is a bit of an overkill for the framework so I don't want to go there.

lirantal:
So what's the status here?

modules/core/server/controllers/errors.server.controller.js
```diff
@@ -45,8 +45,11 @@ exports.getErrorMessage = function (err) {
       case 11001:
         message = getUniqueErrorMessage(err);
         break;
+      case 'UNSUPPORTED_MEDIA_TYPE':
+        message = 'Unsupported filetype';
+        break;
       case 'LIMIT_FILE_SIZE':
-        message = 'Image too big. Please maximum ' + (config.uploads.profileUpload.limits.fileSize / (1024 * 1024)).toFixed(2) + ' Mb files.';
+        message = 'Image too big. The maximum size allowed is ' + (config.uploads.profileUpload.limits.fileSize / (1024 * 1024)).toFixed(2) + ' Mb';
```
simison: There could be a test for this, too.

hyperreality: @lirantal I just saw some bad English there and took the opportunity to clean it up.

@simison is right, while we are at it we should add tests to make sure that those errors are correctly triggered too.

hyperreality:
@lirantal right, I won't bother adding that. But I did have a concern about malicious users finding out that a particular website used the MEAN framework and reading issues like this, and realising they can upload arbitrary data through the route.
The data will not get executed - it just seems wrong to allow it, and it's a potential vulnerability if somebody misconfigures their server in other ways.

modules/core/server/controllers/errors.server.controller.js
simison: https://github.com/Trustroots/trustroots/blob/master/modules/users/tests/server/user.server.routes.tests.js#L1165-L1191 ;-)

lirantal:
I get what you mean, and because of that you should delete your comment and this PR altogether once we merge it so no one can find out about it!!!

On a serious note :) - mime-type checking in real environments actually happen on an isolated system which needs to run some tests, kind of like a service. So a real secure handling is more complex.

hyperreality:
Haha! In that case I think we should add something to the docs about rolling your own MIME-type checking for production environments.

You're right, there are a lots of great pull requests here that got abandoned as developers apparently found all their free time taken up. I really appreciate your dedication to pushing towards version 0.5!

lirantal:
Honestly I think it will overload the README.md but if you meant to add it to the docs at meanjs.org then definitely!

lirantal:
@hyperreality ping me back when you've added tests and this PR is ready to merge

hyperreality:
@lirantal added the image size test. Might be worth noting that the crop directive already tends to shrink the picture size down to <150kB, so it's highly unlikely that people will hit the 1 megabyte limit through the profile picture upload form anyway.

I will be adding @simison's magic byte checker to my own project, but unless there's anything else you want added to the MEAN framework, this one should be complete.

lirantal:
@simison any other comments on this PR before merging it in?

lirantal:
@hyperreality could you check the build erroring out due to bower conflicts for an angular library?

bower.json
```diff
@@ -13,7 +13,7 @@
     "angular-ui-router": "~0.2.18",
     "bootstrap": "~3.3.6",
     "ng-file-upload": "^12.1.0",
-    "ng-img-crop": "ngImgCrop#^0.3.2",
+    "ng-img-crop-full-extended": "ngImgCropFullExtended#~6.0.1",
```
simison: Why `ngImgCropFullExtended#~6.0.1` and not just `~6.0.1`?

bower.json
```diff
@@ -13,7 +13,7 @@
     "angular-ui-router": "~0.2.18",
     "bootstrap": "~3.3.6",
     "ng-file-upload": "^12.1.0",
```
simison: While at it, this could be changed to `~`.

config/assets/default.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
         'public/lib/angular-animate/angular-animate.js',
         'public/lib/angular-bootstrap/ui-bootstrap-tpls.js',
         'public/lib/ng-file-upload/ng-file-upload.js',
-        'public/lib/ng-img-crop/compile/unminified/ng-img-crop.js',
```
simison: Better include unminified version, easier for debugging and our build process should do the minification instead.

config/assets/default.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
...
+        'public/lib/ng-img-crop-full-extended/compile/minified/ng-img-crop.js',
```
simison: Better include unminified version, easier for debugging and our build process should do the minification instead.

config/assets/default.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
         // bower:css
         'public/lib/bootstrap/dist/css/bootstrap.css',
         'public/lib/bootstrap/dist/css/bootstrap-theme.css',
-        'public/lib/ng-img-crop/compile/unminified/ng-img-crop.css'
+        'public/lib/ng-img-crop-full-extended/compile/minified/ng-img-crop.css'
```
simison: `unminified` instead of `minified`

config/env/test.js
```diff
@@ -28,6 +28,14 @@ module.exports = {
   app: {
     title: defaultEnvConfig.app.title + ' - Test Environment'
   },
+  uploads: {
+    profileUpload: {
```
simison: Could be called just 
```
uploads: {
  profile: {
  ...
```
?

config/assets/default.js
lirantal: @hyperreality FYI

config/assets/default.js
lirantal: @hyperreality FYI

bower.json
hyperreality: Bower does this automatically, I think because the package is named `ng-img-crop-full-extended` but the Github repo is `ngImgCropFullExtended`.

For that reason it won't work if we do:
`"ng-img-crop-full-extended": "~6.0.1",`

But we could do this:
`"ngImgCropFullExtended": "~6.0.1",`

config/env/test.js
hyperreality: Good point. I see that in Trustroots you simplified function names like profileUploadFileFilter too, because that function really works on any image, not just the profile uploader.

hyperreality:
Hmm. The `ngImgCropFullExtended` module requires angular 1.4.0 which is why the bower builds failed. Furthermore they appear to have messed up their version numbers by jumping from 0.6.0 to 6.0.1 in the last update. Not sure at all about this module, what do you think @simison?

simison:
@hyperreality Yeah, not very convincing. :-/ Not sure if there's anything very professional / stable out there.

hyperreality:
In that case I'm not averse to rolling back the cropping functionality and just leaving it around here in case somewhat specifically wants it.

lirantal:
@hyperreality any chance we can find something else for cropping? there surely must be something that is widely used

simison:
@lirantal @hyperreality they merged my PR's regarding version issues really fast. It's much better than that badly outdated package, anyway.

hyperreality:
@simison yes work appears to be underway on that package very fast. Thanks for going there and fixing the obvious issues. Do you think it is appropriate to go into the MEAN framework?

lirantal:
@hyperreality did that fix got merged? if so, can you please update this PR status?

simison:
@lirantal @hyperreality it's still a fork of a fork, which is kinda off-putting.

I opened a question about it https://github.com/CrackerakiUA/ngImgCropFullExtended/issues/161

lirantal:
Ugh :(

hyperreality:
No response from the Angular UI team yet about including that crop module.

mleanos:
@lirantal @simison @hyperreality Any update here? Can we move this to the backlog?

hyperreality:
@mleanos Status:
- Image cropping currently uses a now unmaintained library so this should be acted on
- The search for newer libraries wasn't very fruitful, and I'm not convinced that the seeming best choice https://github.com/CrackerakiUA/ngImgCropFullExtended is good enough for MEAN.js. In particular, it seems to have poor compatibility with older browsers and smartphones.
- My conclusion: rollback the cropping functionality. The other stuff in this PR is good and should be included. 

I'll look into updating this tomorrow.

Additionally, I know that @lirantal said above that magic byte checking is outside the scope of the project but want to re-open a discussion on including it. Because right now, the framework is vulnerable out of the box.

mleanos:
@hyperreality Thank you for the update. We're getting ready to release 0.5.0, and would like to wrap up all remaining PR's that are *critical* or fix bugs.

Does it seem reasonable to move it to the backlog? Or is this something that cannot wait?

hyperreality:
@mleanos Apologies for not getting back to this in a timely fashion, it's been another busy week. Because there have been so many intervening changes I'm continuing work at #1589

PR #1468, opened 2016-08-30
Corrected SYNTAX some fixes

frank-dspeed:
## Commit Message Guidelines
```
Header
Blank Line
Body
Blank Line
Footer

The header should look like:
<type>(<scope>): <subject>

The body should have any necessary detailed info about the commit:
An example, references as to where this idea came from, etc.

The footer should have all the issues tagged:
Fixes #123, Fixes #456

So a commit should look like:
feat(users): Add new Yahoo authentication

Yahoo authentication idea proposed by @codydaig
Example implementation in file.js

Fixes #82
```

* Types: 
  * feat - Features, Enhancements, and overall Improvements
  * fix - Fixes, Bugs, HotFixs, etc...
  * doc - Changes to the Documentation and doesn't actually touch any code.
* Scope:
  * The scope should be where the change took place.
  * Examples: users, core, config, articles
* Subject:
  * The subject line should be clear and concise as to what is being accomplished in the commit.
* General Rules:
  * No Line in the Commit message can be longer than 80 characters.
* Reference: [Angular Conventions](https://github.com/ajoslin/conventional-changelog/blob/master/conventions/angular.md)

You did a lot of overhead now the resulting images are much smaller and it will build much faster
greetings.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,64 +6,55 @@
 #
 # Compose:
 # docker-compose up -d
-
 FROM ubuntu:latest
 MAINTAINER MEAN.JS
-
-# Install Utilities
-RUN apt-get update -q
-RUN apt-get install -yqq wget aptitude htop vim git traceroute dnsutils curl ssh sudo tree tcpdump nano psmisc gcc make build-essential libfreetype6 libfontconfig libkrb5-dev
-
-# Install gem sass for grunt-contrib-sass
-RUN apt-get install -y ruby
-RUN gem install sass
-
-# Install NodeJS
-RUN curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -
-RUN sudo apt-get install -yq nodejs
-
-# Install MEAN.JS Prerequisites
-RUN npm install --quiet -g grunt-cli gulp bower yo mocha karma-cli pm2
-
-RUN mkdir /opt/mean.js
-RUN mkdir -p /opt/mean.js/public/lib
-WORKDIR /opt/mean.js
-
-# Copies the local package.json file to the container
-# and utilities docker container cache to not needing to rebuild
-# and install node_modules/ everytime we build the docker, but only
-# when the local package.json file changes.
-# Install npm packages
-ADD package.json /opt/mean.js/package.json
-RUN npm install --quiet
-
-# Install bower packages
-ADD bower.json /opt/mean.js/bower.json
-ADD .bowerrc /opt/mean.js/.bowerrc
-RUN bower install --quiet --allow-root --config.interactive=false
-
-# Share local directory on the docker container
-ADD . /opt/mean.js
-
-# Machine cleanup
-RUN npm cache clean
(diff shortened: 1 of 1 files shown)


lirantal:
@frank-dspeed care to explain?

hyperreality:
Looks like he improved the Docker build process, but the message is certainly misleading and a mess. I'm not running Docker but could somebody who is test this?

frank-dspeed:
i can explain a bit every RUN Instruction runs in its own Layer and Container on docker build so every added run instruction adds a layer example i do RUN apt-get update and then RUN apt-get install and after this RUN apt-get clean then !!!! now the magic happens apt-get clean is useless because the apt-get list is still in the 2 layers before but when i for example do RUN apt-get update && apt-get install git && apt-get clean then its 1 layer that will not include the apt mirror and packet lists because i removed them befor the build container got commited!

frank-dspeed:
Then the next change docker syntax don't accepts EXPOSE 80:80 and this stuff because its simply not allowed to map host ports for security reason sure you can do this on docker run but not in docker build

also every single line EXPOSE ends up in its own build container and image layer so the logic solution is EXPOSE 80 443 and so on 
EXPOSE is only for linking some info to the docker daemon and user that on this ports something could run

but sorry that i got not much more time to explain maybe read the Docker Docs about building images to get deeper into that 

or simply do docker export yourimage > bla.tar
then build my version do docker export myimage > bla2.tar
and see the diff between that results also mesure build time

lirantal:
@frank-dspeed got it. Though I don't see any issue with the separate RUN layers, if it overall simplifies the build process then it's ok with me. Did you test this changed Dockerfile with the rest of the docker setups that were recently added?

@FedeG how does this play well with your recent `Dockerfile-production` and updates to the `docker-compose.yml` file?

FedeG:
This Dockerfile is very simplified but some layers were important for developers.
For example:
1- Layer npm before `ADD . /opt/mean.js.`
Code: 
```
ADD package.json /opt/mean.js/package.json
RUN npm install --quiet
```
If a developer changes files in proyect but don't modify package.json, this layers are caching in future builds.

2- Layer bower before `ADD . /opt/mean.js.`
Code: 
```
ADD bower.json /opt/mean.js/bower.json
ADD .bowerrc /opt/mean.js/.bowerrc
RUN bower install --quiet --allow-root --config.interactive=false
```
If a developer changes files in proyect but don't modify bower.json or .bowerrc, this layers are caching in future builds.

If docker have caching npm/bower commands, the future builds are very fast.

FedeG:
For developers is very important measure build time for successive builds.
In your dockerfile, the time is equal in successive builds but in the old dockerfile the time is minor because the old dockerfile used caching.

FedeG:
My solution: https://gist.github.com/FedeG/1adef09fd0604d9b2050d63aaf611883

lirantal:
@FedeG @frank-dspeed can you guys please work to fix this PR with both of your comments?

PR #1469, opened 2016-08-31
Adds a generic DOMAIN configuration environment

lirantal:
Generic DOMAI configuration environment variable, useful for setting links to an app
in reset email templates, and other cases.

Fixes #871 and #847

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -9,6 +9,9 @@ module.exports = {
   },
   port: process.env.PORT || 3000,
   host: process.env.HOST || '0.0.0.0',
+  // DOMAIN config should be set to the fully qualified application accessible
+  // URL. For example: https://www.myapp.com (including port if required).
+  domain: process.env.DOMAIN,
   // Session Cookie settings
   sessionCookie: {
     // session expiration is set by default to 24 hours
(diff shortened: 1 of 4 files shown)


config/env/default.js
```diff
@@ -9,6 +9,9 @@ module.exports = {
   },
   port: process.env.PORT || 3000,
   host: process.env.HOST || '0.0.0.0',
+  // DOMAIN config should be set to the fully qualified application accessible
+  // URL. For example: https://www.myapp.com (including port if required).
+  domain: process.env.DOMAIN,
```
codydaig: If not set, should this default to localhost? Or at least default to localhost in the development config?

lirantal: We kind of take care of that in https://github.com/meanjs/mean/pull/1469/files/77d4a990f52b83d684a6eb73c7e019576163467e#diff-c3841fb7af244bab7b73ea4c3d72055cR64

It also makes it much complicated to default it to something because you don't know if the server is running in http or https, and this causes many more logic tests to run. It's a simple domain/url setting, I don't think we need to make it anymore complex than that.

codydaig: SGTM

PR #1471, opened 2016-09-01
feat(users) Improved social login buttons UI

hyperreality:
feat(users): Improved social login buttons UI

All social buttons are now responsive and display properly on mobile
The grid is not broken if one is removed
Cleaned up redundant CSS

### DIFF
--- a/modules/users/client/css/users.css
+++ b/modules/users/client/css/users.css
@@ -7,46 +7,27 @@
   display: inline-block;
   position: relative;
 }
-.btn-remove-account {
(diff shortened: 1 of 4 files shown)


hyperreality:
I'm aware that this to some extent undoes the mobile grid layout pushed here #1427

Unfortunately this grid layout was off-center, the buttons were tiny on mobile, and also if a single button was removed (very likely that users won't want to authenticate through all 6 systems) the whole thing would look asymmetrical on either desktop or mobile + the HTML was hard to fix.

PR #1476, opened 2016-09-03
chore(core): Remove Bluemix

codydaig:
Since Bluemix and Cloud Foundry can't be supported and tested, this PR removes it. It's a rapidly iterating service that breaks easily.


lirantal:
Let's merge.

mleanos:
SGTM.

PR #1480, opened 2016-09-05
fix(tests): Reorganized coverage tests

hyperreality:
fix(tests): Reorganized coverage tests

Moved coverage configuration from karma.conf to the gulpfile:
- The server-side coverage tests are already configured in the gulpfile so I think the client-side ones should be too, cleaning up the karma.conf file.
- The previous method to tell karma to run coverage tests involved flipping an environment variable which karma.conf then read. I suspect that this is what caused client-side Travis builds not to send reports to coveralls

LCOV report file concatenation and sending to coveralls has been moved from the gulpfile to .travis.yml, as per @mleanos [suggestion](https://github.com/meanjs/mean/commit/7be191042066e58850899000ee02b6557d744c18#commitcomment-18891729)
- Now coverage tests can be run locally using gulp test:coverage without error if you don't have a coveralls account
- Unfortunately the Karma coverage summary cannot currently be printed to the console due to [a bug](https://github.com/karma-runner/karma/issues/1788) (well, it can be printed, but it causes the karma:coverage gulp task to hang for about 30 seconds).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ before_install:
 after_script:
   - nsp audit-package
   - gulp test:coverage
+  - node_modules/.bin/lcov-result-merger 'coverage/**/lcov.info' | node_modules/coveralls/bin/coveralls.js
 notifications:
   webhooks:
     urls:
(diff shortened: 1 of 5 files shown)


hyperreality:
Boom! Looks like it's covering both client and server side tests. Now so long as it doesn't send out 9 coveralls notifications instead of one...

lirantal:
Cool, thanks!

PR #1485, opened 2016-09-07
feat(views) configure handlebars in separte file

Wuntenn:
feat(views) configure handlebars in separte file

- Moved configuration of handlebars to separate file: 
- Added helper for parsing JSON within templates

Should help to keep `config/lib/express.js` leaner if you want to add quite a few helpers.

### DIFF
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -56,7 +56,7 @@
 
   <!--Embedding The User Object-->
   <script type="text/javascript">
-    var user = {{{ user }}};
+    var user = {{ json user }};
     var env = "{{ env }}";
   </script>
 
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -16,7 +16,7 @@ var config = require('../config'),
   cookieParser = require('cookie-parser'),
   helmet = require('helmet'),
   flash = require('connect-flash'),
-  hbs = require('express-hbs'),
+  handlebars = require('./handlebars'),
   path = require('path'),
   _ = require('lodash'),
   lusca = require('lusca');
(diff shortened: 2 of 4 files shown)


Wuntenn:
The error that I was getting where:
```
user = ;
```
Doesn't seem to be present when testing on a MeanJS empty project. I'm just pushing incase it's useful to anyone.

simison:
Could you explain how to replicate that `user = ;` problem?

modules/core/server/views/layout.server.view.html
```diff
@@ -56,7 +56,7 @@
 
   <!--Embedding The User Object-->
   <script type="text/javascript">
-    var user = {{{ user }}};
+    var user = {{ json user }};
```
simison: I think helper is a bit overkill here, how about just doing:

```
var user = {{#if user}}{{{ user }}}{{else}}null{{/if}};
```

(I know, {{}{{}}} — handlebars...)

simison:
+1 for moving handlebars to somewhere where it's easier to require to e.g. email-sending-worker-script. We did the same already earlier with Swig: https://github.com/Trustroots/trustroots/blob/master/config/lib/render.js

modules/core/server/views/layout.server.view.html
simison: Also, adding a test for this would be cool. You could just copy+modify this one I made last week: https://github.com/Trustroots/trustroots/blob/master/modules/core/tests/server/core.server.config.tests.js

Wuntenn:
Sorry @simison for taking a while to get back I keep thinking about this one and getting sidetracked - not ignoring you. The code on master works fine when I'm logged in and populates the `window.user` object with the data from the `safeUserObject` but for some reason it renders `user = ;` when logged out.

I tried the code that you told me above: `var user = {{#if user}}{{{ user }}}{{else}}null{{/if}};` (looks like list :D:D) somehow it didn't work and gave the same error.

I'd not setup the views before and was curious how to get the helpers into meanjs. When I figured how, and saw how people go about using the for parsing JSON moving them to their own config seemed the way to go. They're a lil hefty!

You're right though, it is most likely overkill but I like that it acts a simple stub to for future helpers and that it's re-usable so may in future cut down the odd one-liners used to stringify stuff.

Did you just say "sunlight" or the "T"-word!? :D Thanks for the test example. I need to write a ton of tests.

Issue #1118, opened 2015-12-24
Allow configurable owasp passwords requirements


brokyo:
I'm not even sure I have an incredibly strong opinion on this but creating an issue and asking for opinions was [suggested here] (https://groups.google.com/forum/#!topic/meanjs/FxdWbkVb0B4) so I'm going for it :)

I've really enjoyed building stuff out in mean.js but while developing I've found dealing with the password requirements to be a pain. I'm curious about how the community would feel about either making the requirements configurable (enabling developers to change minLength or minOptionalTestsToPass for example) or allowing users to bypass the owasp check altogether. I think it's cool that this enforces security but there's something that sits a bit weirdly with me in taking control away from someone building on top of the framework.

Would love to hear your thoughts!

mleanos:
You can configure the owasp test library, as described here https://github.com/nowsecure/owasp-password-strength-test#configuring

You would need to configure these settings on any instance of the owasp test library; both client-side & server-side. As for making them configurable, it's doable. However, I'm not sure if the framework should be concerned with that over the security of the default application.

In order to make this configurable, using the meanjs config library, we would have to somehow share the server-side configuration with the client-side. We don't have a precedent for that type of implementation yet.

I'm not completely opposed to the idea that you're suggesting, but I doubt it's something that we'd see as a high priority.

brokyo:
Thanks for the feedback. I'm afraid I'm a little confused as jloveland [says here] (https://groups.google.com/d/msg/meanjs/FxdWbkVb0B4/Juf1-Vu4BgAJ) that it's not configurable. Am I just being stupid?

Either way totally understand the value of defaulting to security

mleanos:
@brokyo Jason's comment is correct. However, he's referring to the fact that we didn't provide any out of the box implementation for configuring the owasp-password-strength library within the meanjs framework. You may implement your own custom configuration though, using the library's `config` feature.

jloveland:
Thanks for clarifying @mleanos. I think making the password strength configurable in mean.js is outside the scope of the framework, but the developer can choose the config they want an modify the mean.js tests to reflect the new config.

brokyo:
I apologize for missing your responses entirely. That all sounds good to me and what I was looking for. Seeing as this didn't get any other notice I'll just close it. Appreciate the help

PR #1487, opened 2016-09-08
Added configuration for owasp. Synchronize client owap configs with t…

wansco:
Implemented server side configuration for owasp. Exposed this to the client via a route in auth.server.routes.js. I'm not sure that is the best place for it, but it seemed logical.

If the client does not retrieve the settings in time (or fails), it should fallback on the default owasp configuration. I haven't thoroughly tested that part, but it should work since there is already a default configuration defined in owasp.

See #1118. Perhaps another approach would be to create a shared config file that would be distributed to both server and client. I kind of like having everything in the one file so i don't have to hunt for settings. I also dont know what other configurations would need to be shared. Maybe selecting which social accounts you want to support?

I also added the current time to the 'Invalid username or password' to provide the user some visual feedback that their subsequent login attempts also failed. Feel free to reject this one as it is only loosely related to the owasp stuff.

### DIFF
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -79,11 +79,11 @@ module.exports = {
     sandbox: false
   },
   owasp: {
-      allowPassphrases       : true,
-      maxLength              : 128,
-      minLength              : 10,
-      minPhraseLength        : 20,
-      minOptionalTestsToPass : 4,
+    allowPassphrases: true,
+    maxLength: 128,
+    minLength: 10,
+    minPhraseLength: 20,
+    minOptionalTestsToPass: 4
   },
   mailer: {
     from: process.env.MAILER_FROM || 'MAILER_FROM',
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -78,13 +78,6 @@ module.exports = {
     callbackURL: '/api/auth/paypal/callback',
     sandbox: false
   },
-  owasp: {
-    allowPassphrases: true,
-    maxLength: 128,
-    minLength: 10,
-    minPhraseLength: 20,
-    minOptionalTestsToPass: 4
-  },
   mailer: {
     from: process.env.MAILER_FROM || 'MAILER_FROM',
     options: {
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -78,6 +78,13 @@ module.exports = {
     callbackURL: '/api/auth/paypal/callback',
     sandbox: false
   },
+  owasp: {
+      allowPassphrases       : true,
+      maxLength              : 128,
+      minLength              : 10,
+      minPhraseLength        : 20,
+      minOptionalTestsToPass : 4,
+  },
   mailer: {
     from: process.env.MAILER_FROM || 'MAILER_FROM',
     options: {
--- a/modules/core/server/controllers/core.server.controller.js
+++ b/modules/core/server/controllers/core.server.controller.js
@@ -26,7 +26,7 @@ exports.renderIndex = function (req, res) {
 
   res.render('modules/core/server/views/index', {
     user: JSON.stringify(safeUserObject),
-    owaspConfig: JSON.stringify(config.owasp)
+    sharedConfig: JSON.stringify(config.shared)
   });
 };
 
--- a/modules/core/server/controllers/core.server.controller.js
+++ b/modules/core/server/controllers/core.server.controller.js
@@ -1,12 +1,13 @@
 'use strict';
 
-var validator = require('validator');
+var validator = require('validator'),
+  path = require('path'),
+  config = require(path.resolve('./config/config'));
 
 /**
  * Render the main application page
  */
 exports.renderIndex = function (req, res) {
-
   var safeUserObject = null;
   if (req.user) {
     safeUserObject = {
@@ -24,7 +25,8 @@ exports.renderIndex = function (req, res) {
   }
 
   res.render('modules/core/server/views/index', {
-    user: JSON.stringify(safeUserObject)
+    user: JSON.stringify(safeUserObject),
+    owaspcfg: JSON.stringify(config.owasp)
   });
 };
 
--- a/modules/core/server/controllers/core.server.controller.js
+++ b/modules/core/server/controllers/core.server.controller.js
@@ -26,7 +26,7 @@ exports.renderIndex = function (req, res) {
 
   res.render('modules/core/server/views/index', {
     user: JSON.stringify(safeUserObject),
-    owaspcfg: JSON.stringify(config.owasp)
+    owaspConfig: JSON.stringify(config.owasp)
   });
 };
 
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -68,7 +68,7 @@
   
   <!--owasp config sync-->
   <script type="text/javascript">
-    owaspPasswordStrengthTest.configs = {{{ owaspConfig }}};
+    owaspPasswordStrengthTest.config({{{ owaspConfig }}});
   </script>
 
   {{#if livereload}}
--- a/modules/core/server/views/layout.server.view.html
+++ b/modules/core/server/views/layout.server.view.html
@@ -68,7 +68,8 @@
   
   <!--owasp config sync-->
   <script type="text/javascript">
-    owaspPasswordStrengthTest.config({{{ owaspConfig }}});
+    var sharedConfig = {{{ sharedConfig }}};
+    owaspPasswordStrengthTest.config(sharedConfig.owasp);
   </script>
 
   {{#if livereload}}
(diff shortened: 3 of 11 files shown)


lirantal:
@wansco would you like to add back some tests too?
I'm overall neutral on this but it seems ok.

wansco:
This broke a number of tests due to the extra http request to get the single source owasp config from the server. After digging around in this, I think a 2nd config file, or something a bit more generic, might have been the better approach. The problem is I don't know the optimal place to put that config file.

"Logically" it should go in /config/env/client.js (or /config/env/shared.js), but that is not accessible from the client (and due to the sensitive nature of the other files in that directory, the client should never have direct access to anything in there).

What do you think about adding /modules/core/shared/config.js? That may be a bit confusing since there is already /modules/core/client/app/config.js, but it serves a different purpose.

Or, rather than getting just the owasp config through a $http.get in password-validator.client.service.js where it needs it, maybe we could get it earlier in the application cycle along with any other shared configurations?



Recapping the fundamental issue I'm trying to resolve:
owasp gets its default config from 2 different places:
Server:
/node_modules/owasp-password-strength-test/owasp-password-strength-test.js
Client:
/public/lib/owasp-password-strength-test/owasp-password-strength-test.js

Also, password-validator.client.service.js has a hardcoded length requirement in the popover message.

lirantal:
@wansco what about exposing the relevant config options like owasp through a config variable in handlebars templating on the core module view's template (checkout the layout template). Similar to how we're doing the user object.

mleanos:
I'd much prefer Liran's suggestion of exposing this data through handlebars templating.

wansco:
Oh, that is much easier. I need to revert and clean up some of my changes, but that should be pretty simple.

simison:
Why different configs for `production` / `development` env?

config/env/production.js
```diff
@@ -78,6 +78,13 @@ module.exports = {
     callbackURL: '/api/auth/paypal/callback',
     sandbox: false
   },
+  owasp: {
+      allowPassphrases       : true,
+      maxLength              : 128,
+      minLength              : 10,
+      minPhraseLength        : 20,
+      minOptionalTestsToPass : 4,
+  },
```
simison: Why different configs for `production` / `development` env?

wansco: I'm fine moving this to the default config. The problem is it may break some tests until they can be updated to pull from the config. Right now some of the tests like password length are hard coded in the tests. I forgot to update the test config, but I'll be updating the PR shortly.

modules/users/server/models/user.server.model.js
```diff
@@ -12,7 +12,7 @@ var mongoose = require('mongoose'),
   generatePassword = require('generate-password'),
   owasp = require('owasp-password-strength-test');
 
-
+console.log(config.owasp, "========================================================", path.resolve('./config/config'));
```
simison: Forgotten debug code? :-)

wansco: yup. oops

modules/core/server/controllers/core.server.controller.js
```diff
@@ -24,7 +25,8 @@ exports.renderIndex = function (req, res) {
   }
 
   res.render('modules/core/server/views/index', {
-    user: JSON.stringify(safeUserObject)
+    user: JSON.stringify(safeUserObject),
+    owaspcfg: JSON.stringify(config.owasp)
```
simison: I think it would be fine to have more elaborate `owaspConfig`? Less cryptic.

wansco: fixed

hyperreality:
Nice PR, I had a similar way of doing it in my own project. I particularly like the more lenient password tests during development (and tests that use the config), perhaps that should be added to the documentation.

modules/users/server/config/strategies/local.js
```diff
@@ -22,7 +22,7 @@ module.exports = function () {
       }
       if (!user || !user.authenticate(password)) {
         return done(null, false, {
-          message: 'Invalid username or password'
+          message: 'Invalid username or password (' + (new Date()).toLocaleTimeString() + ')'
```
mleanos: Why is the data/time added here?

modules/users/server/controllers/users/users.password.server.controller.js
```diff
@@ -14,6 +14,7 @@ var path = require('path'),
 
 var smtpTransport = nodemailer.createTransport(config.mailer.options);
 
+
```
mleanos: Remove this new empty line.

modules/users/client/services/password-validator.client.service.js
```diff
@@ -11,6 +11,7 @@
   function PasswordValidator($window) {
     var owaspPasswordStrengthTest = $window.owaspPasswordStrengthTest;
 
+
```
mleanos: Remove empty line.

modules/core/server/views/layout.server.view.html
```diff
@@ -65,6 +65,11 @@
 
   <!--Application JavaScript Files-->
     {{#each jsFiles}}<script type="text/javascript" src="{{this}}"></script>{{/each}}
+  
+  <!--owasp config sync-->
+  <script type="text/javascript">
+    owaspPasswordStrengthTest.configs = {{{ owaspConfig }}};
```
mleanos: Shouldn't this be `owaspPasswordStrengthTest.config`, according to https://github.com/nowsecure/owasp-password-strength-test#configuring?

modules/users/server/models/user.server.model.js
```diff
@@ -4,12 +4,17 @@
...
 var mongoose = require('mongoose'),
+  path = require('path'),
+  config = require(path.resolve('./config/config')),
   Schema = mongoose.Schema,
   crypto = require('crypto'),
   validator = require('validator'),
   generatePassword = require('generate-password'),
   owasp = require('owasp-password-strength-test');
 
+owasp.configs = config.owasp;
```
mleanos: Same as my comment above.
https://github.com/nowsecure/owasp-password-strength-test#configuring

mleanos:
@wansco Thank you for updating this PR. It looks good now. I left some feedback, but nothing major. This should be good to merge once those comments are addressed.

mleanos:
One last nit pick, in regards to: https://github.com/meanjs/mean/pull/1487#discussion_r78059480

Can we just leave the config in the default config? If a developer needs to have these settings different between their environments, they can add it to whichever env config they want.

I'd probably only add the owasp config settings to my *local-development* config. The other env's would just continue to use the *default*.

modules/core/server/views/layout.server.view.html
wansco: Good catch. It seems like setting configs directly works, but its better to set it via the function as per the docs.

modules/users/server/config/strategies/local.js
wansco: This is not part of the owasp stuff, so feel free to reject it. I've always been bugged when I put in an invalid password and then the UI does not change on a subsequent invalid password. At least this give some feedback to the user that "yes, page reloaded (or did something) and the password was still invalid".

config/env/production.js
wansco: If the config is in default and someone changes it there (as opposed to duplicating it and changing it in the local), the tests might fail. Say you reduce the minlength requirement, that test would fail... although that may not be a bad thing.

modules/users/server/config/strategies/local.js
mleanos: I'd rather the client application be given the responsibility to provide that UI feedback. Most likely it should be in the form of an animation/ui-effect/sound, or something similar. Users of this framework will probably have different ways of conveying that the page did indeed reload or attempt to login again.

We should remove the date/time stamp from this piece.

config/env/production.js
simison: No one should touch `default.js` anyway, it's wrooong and should cause tests to fail. ;-) They will modify `local-{env}.json` files instead.

lirantal:
@wansco is there any reason to put the owasp config under a shared object?
It's cleaner if we just have the owasp object there in the config file. Sorry about the config churn.

wansco:
I restructured the config for future flexibility. Its a bit more verbose than it needs to be for just the owasp stuff, but I think it will be useful down the road. Basically if someone wants to pass a "global" config setting down to the client they can add it to config.shared and access it on the client through sharedConfig (or $window.sharedConfig). Whether they "should" be able to do that is up for debate.

They could certainly handle that on their own, but this at least provides a central and straightforward mechanism for that. This way you no longer have to muck with the core code to "add" additional configurations that are available to both server and client. Just add/change it in local-[env].js.

lirantal:
I'm ok with that, thanks.

PR #1494, opened 2016-09-09
Updated Dockerfile

frank-dspeed:
Use copy and not add. Add is a special command used for some unzip and permission magic that you shouldn't use if you don't know what it is.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,15 +64,15 @@ WORKDIR /opt/mean.js
 # and install node_modules/ everytime we build the docker, but only
 # when the local package.json file changes.
 # Install npm packages
-ADD package.json /opt/mean.js/package.json
+COPY package.json /opt/mean.js/package.json
 RUN npm install --quiet && npm cache clean
 
 # Install bower packages
-ADD bower.json /opt/mean.js/bower.json
-ADD .bowerrc /opt/mean.js/.bowerrc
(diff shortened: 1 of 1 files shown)


simison:
Is `COPY` just more common than `ADD` in Dockerworld or why do you suggest this change?

There isn't any real difference between the two in this case: https://stackoverflow.com/questions/24958140/docker-copy-vs-add

Please also see [commit guidelines](https://github.com/meanjs/mean/blob/master/CONTRIBUTING.md), PR's with just "Updated Dockerfile" without any explanation aren't really useful. :-(

frank-dspeed:
You should check the [ADD](https://docs.docker.com/engine/reference/builder/#add) and [COPY](https://docs.docker.com/engine/reference/builder/#copy) documentations for an exhaustive description of their behaviours, but in a nutshell the major difference is that ADD can do more than COPY and so leads to unexpected results in some situations:

ADD allows <src> to be an URL
If the <src> parameter of ADD is an archive in a recognised compression format, it will be unpacked

frank-dspeed:
at Docker Inc the company behond docker all image creators get informed to use COPY over ADD only use ADD if it is realy needed and the expected result that a file should get added extracted and moded

simison:
👍  Sounds good to me. :-)

PR #1495, opened 2016-09-09
fix(users) Stops error on signin/signup

### LINKED ISSUE #1284: Twitter login doesnt' redirect properly because passport.authenticate

### LINKED ISSUE #1290: OAuth Login Issues  Error: /api/auth/google/[object%20Object] is not a valid path.
Hi

Ive been working on a project for a while and this problem just popped up. Im not sure how far back the problem occured or on what commit.

When someone logs in with Facebook or Google it redirects to the server 404 page and shows

`Page not found  Error: /api/auth/google/[object%20Object] is not a valid path.`

However if i refresh the page the user is logged in.

Wuntenn:
Resolves issue relating to info object being used as the redirect path.

- Improves use of the passport `info` object allowing slight simplification of login; and 
- removes the need to temporarily cache the redirect within the session.

Fixes #1290, Fixes #1284


mleanos:
Restarted the build, since it looked like it was just a timeout issue with the last run.

mleanos:
LGTM. I've pulled this down, and tested. I can verify this indeed fixes the issues of the redirects.

However (there's always a "however"), I ran into an issue that's unrelated to the aim of this PR..

With [this PR](https://github.com/meanjs/mean/pull/1250), we introduced a fix to enabled the `sparse` index option. This handles the case where a provider doesn't send back an email when our User is registering as a new user. The issue I just ran into is that [here](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.authentication.server.controller.js#L161), the `email` field is set to an empty string if `providerUserProfile.email` is `undefined`, since we have a default value of the `email` field [here](https://github.com/meanjs/mean/blob/master/modules/users/server/models/user.server.model.js#L55). This will, in effect, negate our use of the `sparse` option.

In my PR that introduced the `sparse` option, I wrote a test to verify the `sparse` was working correctly, but failed to realize the impact of the default value of `email`. Doh!

@Wuntenn I think we can fix this issue in this PR. What say you?

My two suggested fixes are:
1) Remove [this line](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.authentication.server.controller.js#L161) & add `user.email = providerUserProfile.email` after we instantiate the new User object. This may look like a band-aid or hack-ish fix but we *should* only run into this issue when creating a new User during this Provider authentication process.
OR
2) Remove our [default value](https://github.com/meanjs/mean/blob/master/modules/users/server/models/user.server.model.js#L55) of the Email field. I can't think of any reason why we have this default value, and I don't see any negative impact of doing so.

Option #2 may have more "business rule" impact for our users (i.e. more controversial to our users) , so I'm sort of leaning more toward option #1.

PR #1496, opened 2016-09-09
fix(core): Redirect to login page rather than 403 if not signed in

hyperreality:
feat(users): Redirect to login page rather than 403 if not signed in

Non-signed in users are being redirected to the unfriendly 403 forbidden page instead of the sign in page as should be the case. Fixed the tiny bug in the code that caused this to happen.

### DIFF
--- a/modules/core/client/config/core.client.route-filter.js
+++ b/modules/core/client/config/core.client.route-filter.js
@@ -25,7 +25,7 @@
 
         if (!allowed) {
           event.preventDefault();
-          if (Authentication.user !== undefined && typeof Authentication.user === 'object') {
+          if (Authentication.user !== null && typeof Authentication.user === 'object') {
             $state.transitionTo('forbidden');
           } else {
             $state.go('authentication.signin').then(function () {


mleanos:
@hyperreality Aren't there instances where we would have an `undefined` User in the Authentication service?

Maybe we should check for both `null`, and `undefined`?

hyperreality:
I considered that, but after playing around with the app a bit I couldn't reach a state where Authentication.user was undefined. `var user = null` is embedded in the layout. I liked specifying `null` directly because it is more explicit.

I'm not sure we need to worry about the case where people will create a different layout view in which the user object isn't embedded, because then they would have to rewrite the entire `stateChangeStart` function anyway. Thoughts?

mleanos:
Perfect answer! Thanks! LGTM.

We'll wait to see if others want to chime in over the next day or so.

simison:
BTW one curious bit of JavaScript knowledge:

```
foo === null
```

Matches only `null` as it should.

but...

```
foo == null
```

equals:

```
foo == null || foo == undefined
```

hyperreality:
@simison, yes I like that shortcut too, unfortunately eslint complains about the lazy comparison operator

frkwhiteangel:
I dont get this to work? its still not redirecting for me.

PR #1498, opened 2016-09-10
fix(lint): CSS alphabetize warnings

mleanos:
Fixes css lint warnings of properties not alphabetized.


mleanos:
Merging since this is super simple, and I would like to get rid of the warnings :)

PR #1499, opened 2016-09-10
feat(core): Minor UI fixes for mobile; autofocus

hyperreality:
feat(core): Minor UI fixes for mobile; autofocus

- Social login buttons now 80% of old size, they are too large on small screens such as the iPhone and oversized on the desktop too IMO
- Moved the 'sign in failed' warning above the login form, so that it is not cut off on small devices
- Fix broken password popover balloon 
- Added an autofocus directive to automatically move focus to relevant fields on MEAN.js forms. The directive only fires if screen width >= 800px because otherwise a large keyboard may open on a mobile device.
- The autofocus directive overwrites the 'autofocus' HTML5 attribute (the attribute is no good for angular apps as it doesn't work across state changes) - I don't believe there's any problem with doing that but I'm open to being corrected


mleanos:
@hyperreality Can you think of a test to write for this new directive? The coverage decreased, so that may be why the coveralls build failed. I think we have a threshold of coverage decreasing.

@lirantal I couldn't find the settings on Coveralls. Do you know what this setting is, and where I can see it?

hyperreality:
@mleanos Just realised that you can send keys on Protractor, I'll play around with it.

The coveralls settings are a tiny button on the top left.

hyperreality:
The test is failing on Travis - I suspect that's because the directive only fires if window width is over 800px. However, trying to get Protractor tests working with different windows sizes was a nightmare I don't fancy revisiting for the time being!

mleanos:
@hyperreality Ok. Understood. Perhaps we can just leave that test out then?

I'm surprised the coverage decreased so much by merely adding this one directive.

hyperreality:
I'll have one more crack at the test I think.

In the modules directory:
`find . -name '*.js' | xargs wc -l` reports ~10000 lines of code
`wc -l core/client/directives/` is 28 lines
Which is about 0.2%

lirantal:
@hyperreality let's see if we can kick this in before a coming 0.5.0 release

modules/users/client/views/settings/manage-social-accounts.client.view.html
```diff
@@ -8,8 +8,11 @@ <h3 class="col-md-12 text-center" ng-show="vm.hasConnectedAdditionalSocialAccoun
     </div>
     <div ng-repeat="(providerName, providerData) in vm.user.additionalProvidersData" class="social-account-container">
       <img ng-src="/modules/users/client/img/buttons/{{providerName}}.png">
-      <a class="btn btn-danger btn-add-remove-account" ng-click="vm.removeUserSocialAccount(providerName)">
-        <i class="glyphicon glyphicon-trash"></i>
+      <a ng-click="vm.removeUserSocialAccount(providerName)">
+        <img class="social-button" ng-src="/modules/users/client/img/buttons/{{providerName}}.png">
```
mleanos: These social provider images are being shown twice. If you see above at line 10, the `` is present there as well. Remove whichever one we don't need. I'm guessing we want this new one that you added.

modules/users/client/views/settings/manage-social-accounts.client.view.html
```diff
@@ -8,8 +8,11 @@ <h3 class="col-md-12 text-center" ng-show="vm.hasConnectedAdditionalSocialAccoun
...
+        <span class="btn btn-success btn-add-remove-account">
```
mleanos: This should be using the `btn-danger` class, rather than `btn-success`.

mleanos:
@hyperreality I left a review, but for some reason it doesn't mark the PR as "updated".

modules/users/client/views/settings/manage-social-accounts.client.view.html
```diff
@@ -8,8 +8,11 @@ <h3 class="col-md-12 text-center" ng-show="vm.hasConnectedAdditionalSocialAccoun
     </div>
     <div ng-repeat="(providerName, providerData) in vm.user.additionalProvidersData" class="social-account-container">
       <img ng-src="/modules/users/client/img/buttons/{{providerName}}.png">
```
simison: It would probably make sense to use [one time binding](https://docs.angularjs.org/guide/expression#one-time-binding) here. (`{{::providerName}}`)

hyperreality: Good call, there's several other places in the project where it would make sense too that we'll have to get round to eventually.

modules/users/client/views/settings/manage-social-accounts.client.view.html
hyperreality: Cheers

mleanos:
@hyperreality I'd like to merge this in, but I'm having a hard time reviewing this with all the other commits from your rebase included. I don't know which commits are specific to this PR. Would it be possible for you to clean this up?

mleanos:
@meanjs/contributors @hyperreality 

I pulled this branch down locally, and was able to review. Everything seems to be working. I'm merging this now since it introduces a few important fixes, and we need to finalize the other open PR's to get ready for 0.5.0.

Issue #94, opened 2014-06-10
Enhancement: Verify user with email


amobrem:

trevis-baker:
What is the status on this?  Is it implemented?

ilanbiala:
This hasn't been implemented. If you'd like to contribute this feature, that would be great. The only thing is that we should merge #343 first because that will affect the user model. Take a look at #343 and see if you can suggest any changes first, and then we can add in user verification.

lirantal:
Let's make sure we're doing it on 0.4 if possible

ilanbiala:
@lirantal As part of 0.4.0 or in 0.4.x? I'm assuming you mean 0.4.0 since it would probably be a breaking change.

lirantal:
in 0.4.x I think, because we don't want to add anymore work on the 0.4 branch - we just want to get it out already.

freemind-:
Hey guys, who's in charge for this one? Any proof of concept has been committed or what? If not, let me know.

lirantal:
I'm closing this infavor of the work being done on the related PR https://github.com/meanjs/mean/pull/1382
cc @freemind-

PR #1500, opened 2016-09-11
feat(users): Email verification

hyperreality:
feat(users): Email verification

This PR continues work that was started and abandoned in #1382, and was requested as far back as #94

I've got email verification up and running. You can check it out by starting https://github.com/djfarrelly/MailDev and using the following in `config.mailer.options`:
`port: 1025,
  ignoreTLS: true,`

There are a number of concerns that I want to bring up for discussion here:
@mleanos pointed out in the previous PR that the verification code needed to be much better organized. I fully agree with him, but found that due to the way node-email-verification works it is rather difficult to move any of it around.

As it is, I have put the signup functionality in `modules/users/server/controllers/users/users.verification.server.controller.js`. Taking this file piece by piece:
 - It is hard to move the node-email-verification configuration into for instance a config file, because it itself relies on several configuration values. Furthermore, the [docs](https://www.npmjs.com/package/email-verification) state all node-email-verification steps should be done in one central file - it's also not possible to use templates for the email HTML bodies at the moment.
 - I don't think it's right to move the API functions for validating and resending emails; refactoring them would become confusing.
 - I didn't want to do it but I moved the signup function into this controller. If you can see an easy way for it to remain in the users.authentication controller yet still capable of invoking the node-email-verification createTempUser functionality in another file, then please let me know. 
 - Right now a config variable decides whether a new user needs email verification or is saved the old way. You would want to easily switch email verification off during development. This led to the rather ugly hack of checking whether the response is a string or object post-sign up to work out whether verification was required or not.

In other words, it feels like the configuration, the verification API, and the creation of a user, should all be in different places, but in actual fact it's very hard to separate them. I can try to justify this by saying that `users.verification.server.controller.js` will simply handle signup... perhaps it should be called something else.

This is far from a final version, and I'm posting to ask for your thoughts on the issue and how this might proceed.

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -46,5 +46,6 @@ module.exports = {
         fileSize: 1 * 1024 * 1024 // Max file size in bytes (1 MB)
       }
     }
-  }
+  },
+  verifyUserEmail: true
 };
--- a/modules/users/client/config/users.client.routes.js
+++ b/modules/users/client/config/users.client.routes.js
@@ -123,6 +123,24 @@
         data: {
           pageTitle: 'Password reset form'
         }
+      })
+      .state('resendVerificationEmail', {
+        url: '/verification',
+        templateUrl: 'modules/users/client/views/verification/resend-verification-email.client.view.html',
+        controller: 'VerificationController',
+        controllerAs: 'vm',
+        data: {
+          pageTitle: 'Resend Verification Email'
+        }
+      })
+      .state('verification', {
+        url: '/verification/:token',
+        templateUrl: 'modules/users/client/views/verification/verify-account.client.view.html',
+        controller: 'VerificationController',
+        controllerAs: 'vm',
+        data: {
+          pageTitle: 'Verify Account'
+        }
       });
   }
 }());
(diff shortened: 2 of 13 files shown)


mleanos:
@hyperreality Thank you for the detailed explanation of the different considerations in this feature. I'll try to review this thoroughly over the next few days.

My initial thought is that email verification might be out of the scope for this framework. There are many different flows that our users could choose from for their signup/email-verification process. Currently, our signup process is very simple, and organized in a way that makes it easy to integrate any other type of flow. Including email-verification. I've done this for one of my projects based off of 0.4.0 and it wasn't hard at all. However, the flow I chose was very specific to what I needed.

However, it this feature is indeed desired (by the community) to be included in the framework, it would be nice to see we can use a little finesse and organize this a little better. I haven't really got into the changes in this PR, so I don't have any insight into what can be done. Once I have a closer look, I'll report back.

lirantal:
Joining Michael, thanks for the elaborate explanation.

It seems that the issue might really lie with the use of the `node-email-verification` library. How about if we choose something else? More mature, more maintained. There surely are dozens of options out there that we can re-use a very simple config to pull this off.

hyperreality:
@lirantal A quick search of npm turns up nothing viable apart from the node-email-verification package unfortunately.

lirantal:
So let's abandon if there's nothing vitale.

mleanos:
Closing for now. We can readdress this in the future if we come up with a better solution.

PR #1508, opened 2016-09-15
Deprecated $http success/error promise methods

### LINKED ISSUE #1479: Fix deprecated `$http.error()` and `$http.success()`
https://docs.angularjs.org/api/ng/service/$http

>The `$http` legacy promise methods `success` and `error` have been deprecated. Use the standard `then` method instead. If [`$httpProvider.useLegacyPromiseExtensions`](https://docs.angularjs.org/api/ng/provider/$httpProvider#useLegacyPromiseExtensions) is set to false then these methods will throw [`$http/legacy`](https://docs.angularjs.org/error/$http/legacy) error.

Used at least here:
- https://github.com/meanjs/mean/blob/59e6daa63d65798886e446fc7696b6331c9e3aba/modules/users/client/controllers/settings/change-password.client.controller.js
- https://github.com/meanjs/mean/blob/c95bd7d1cb0c6e651f82b2b4cfa4e937d4dcf154/modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
- https://github.com/meanjs/mean/blob/59e6daa63d65798886e446fc7696b6331c9e3aba/modules/users/client/controllers/authentication.client.controller.js

mleanos:
Replaces the $http service calls with promise based methods of the client-side UsersService for the following:
  Users Change Password
  Users Manage Social Accounts
  Users Password Forgot
  Users Password Reset
  Users Signup
  Users Signin

Modifies tests to reflect changes.

Closes #1479

### DIFF
--- a/modules/users/client/controllers/authentication.client.controller.js
+++ b/modules/users/client/controllers/authentication.client.controller.js
@@ -5,9 +5,9 @@
     .module('users')
     .controller('AuthenticationController', AuthenticationController);
 
-  AuthenticationController.$inject = ['$scope', '$state', '$http', '$location', '$window', 'Authentication', 'PasswordValidator'];
+  AuthenticationController.$inject = ['$scope', '$state', 'UsersService', '$location', '$window', 'Authentication', 'PasswordValidator'];
 
-  function AuthenticationController($scope, $state, $http, $location, $window, Authentication, PasswordValidator) {
+  function AuthenticationController($scope, $state, UsersService, $location, $window, Authentication, PasswordValidator) {
     var vm = this;
 
     vm.authentication = Authentication;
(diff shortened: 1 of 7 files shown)


mleanos:
I would like to point out that the `UsersService` methods that I added make calls to the `api/auth..` should really go into the `Authentication` service, or a new service, since they don't really interact with the same base URL `api/users`. However, doing so would require a major refactoring of the `Authentication` service, which could probably wait for now. It's a little ambiguous of where these should go on the back-end as well since our `api/auth` routes are handled by User server-controllers.

codydaig:
@mleanos Is this ready for review?

mleanos:
Yes :)

lirantal:
@mleanos I'm fine with the change if you guys want to push in. Let's make sure we don't forget about this as I'd really want to kick 0.5.0 out the door already.
@hyperreality maybe you can help too with reviewing this.

codydaig:
Haven't tested, but LGTM from the code perspective. Glad to see this updated!

PR #1513, opened 2016-09-16
feat (articles): For New Article, delete button not required. if try to delete New Article that is not persisted to database than it create server error

kuldipem:
feat (articles): For New Article, delete button not required. if try to delete New Article that is not persisted to database than it create server error


lirantal:
It doesn't actually create a server error, just a tries to delete something that doesn't exist but the fix on the UI is good.
Thanks!

@hyperreality @mleanos I'm merging but just incase you want to further review on this.

PR #1516, opened 2016-09-16
attempting to update karma and gulp-load-plugins to check build status

lirantal:


mleanos:
@lirantal I say merge this in when you're done updating the dependencies.. just as long as the build passes, then we should be good.

lirantal:
Yep, looks good!

PR #1523, opened 2016-09-23
deploy(docker) fix(docker) - Updated Dockerfile production

FedeG:
Equals to #1494 but in production.


mleanos:
I haven't tested either, but LGTM.

@codydaig Feel free to merge as long as nobody else has any feedback.

Issue #369, opened 2015-01-19
Upgrading to use Material Design 


denzelwamburu:
The default notification and warning is kind of 'flat' i stumbled upon a fancy notification add on https://github.com/Salakar/angular-toasty. Trying to integrate this poses issue since i can't find the location of the <error> attribute in the whole find system specifically the one rendering the result to the client.

ilanbiala:
I suggest we use angular-material.

ilanbiala:
@denzelwamburu did you take a look at the angular-material $mdToast API? Also, angular-toasty seems to be neglected.

denzelwamburu:
I have been trying to work around with material and seems its not that hard,the front end tags will need to be modified heavily.The angular-material will need to be installed as a module using bower then declared in the modules of angular,maybe something like that.I haven't successfully done one yet but am in the process

amirmo:
angular-material ++ :+1:

aarosil:
+1

andresmanz:
Yeah, that would be awesome. +1

SOSANA:
1+ Add me to that list, maybe we can just add it to build process as an option?

denzelwamburu:
Yeah,as an option in the generator

ilanbiala:
@denzelwamburu I think we should start with porting over meanjs/mean to ngMaterial, and then moving the generator over. However, I think that's more appropriate for the next major version after 0.4, which hopefully will come along much quicker than 0.4 is.

lirantal:
agree with @ilanbiala, and @denzelwamburu if you're up for the challenge we'll be more than happy if you want to take a look into it.

amirmo:
Going with ngMaterial = big time community exposure for meanjs
+1

ilanbiala:
I've already worked with ngMaterial, so I can definitely help guide you, but it should definitely be off the 0.4 branch.

denzelwamburu:
Sure,i agree with all the suggestion.But it should not replace the bootstrap,it should be an option during installation in the generator.
I don't have that experience in ngMaterial but am ready to give it a try

neilhanekom:
I also using ngMaterial with the 0.4.0 branch. Obviously upgrading to
Angular 1.3 is required. This off course works great in Development mode. I
have picked up a few issues when minification is done through "grunt
build". Seems like $injector cannot pickup certain dependencies that being
injected. Dunno whether is due to ngAnnotate that is causing this.... If
[...]

PR #1532, opened 2016-09-30
feat(core): add notification feedback with angular-ui-notification

### LINKED ISSUE #1274: Remove in Article -> Add callback
> mean/modules/articles/client/controllers/articles.client.controller.js

```
function remove() {
  if ($window.confirm('Are you sure you want to delete?')) {
    vm.article.$remove($state.go('articles.list'));
}
```
I think it would be better to do as follow: 

```
function remove() {
  if ($window.confirm('Are you sure you want to delete?')) {
    vm.article.$remove(function (){
      $state.go('articles.list'));
    });
}
```
In Order to do so with a callback, and so it en the next digest.

### LINKED ISSUE #1531: Admin user able to delete self
Admin user is able to delete self

sujeethk:
feat(core): add notification feedback with angular-ui-notification

Added visual notification for user/article updates
angular-ui-notification config added to core client config
Notification idea from #369

### DIFF
--- a/modules/articles/client/controllers/admin/article.client.controller.js
+++ b/modules/articles/client/controllers/admin/article.client.controller.js
@@ -5,14 +5,13 @@
     .module('articles.admin')
     .controller('ArticlesAdminController', ArticlesAdminController);
 
-  ArticlesAdminController.$inject = ['$scope', '$state', '$window', 'articleResolve', 'Authentication'];
+  ArticlesAdminController.$inject = ['$scope', '$state', '$window', 'articleResolve', 'Authentication', 'Notification'];
 
-  function ArticlesAdminController($scope, $state, $window, article, Authentication) {
+  function ArticlesAdminController($scope, $state, $window, article, Authentication, Notification) {
     var vm = this;
 
     vm.article = article;
     vm.authentication = Authentication;
-    vm.error = null;
     vm.form = {};
     vm.remove = remove;
     vm.save = save;
@@ -22,6 +21,7 @@
       if ($window.confirm('Are you sure you want to delete?')) {
         vm.article.$remove(function() {
           $state.go('admin.articles.list');
+          Notification.success({ message: '<i class="glyphicon glyphicon-ok"></i> Article deleted successfully!' });
         });
       }
     }
@@ -40,10 +40,11 @@
 
       function successCallback(res) {
         $state.go('admin.articles.list'); // should we send the User to the list or the updated Article's view?
+        Notification.success({ message: '<i class="glyphicon glyphicon-ok"></i> Article saved successfully!' });
       }
 
       function errorCallback(res) {
-        vm.error = res.data.message;
+        Notification.error({ message: res.data.message, title: '<i class="glyphicon glyphicon-remove"></i> Article save error!' });
       }
     }
   }
--- a/modules/core/client/app/config.js
+++ b/modules/core/client/app/config.js
@@ -6,7 +6,7 @@
   var service = {
     applicationEnvironment: window.env,
     applicationModuleName: applicationModuleName,
-    applicationModuleVendorDependencies: ['ngResource', 'ngAnimate', 'ngMessages', 'ui.router', 'ui.bootstrap', 'ngFileUpload', 'ngImgCrop'],
+    applicationModuleVendorDependencies: ['ngResource', 'ngAnimate', 'ngMessages', 'ui.router', 'ui.bootstrap', 'ngFileUpload', 'ngImgCrop', 'ui-notification'],
     registerModule: registerModule
   };
 
@@ -20,4 +20,17 @@
     // Add the module to the AngularJS configuration file
     angular.module(applicationModuleName).requires.push(moduleName);
   }
+
+  // Angular-ui-notification configuration
+  angular.module('ui-notification').config(function(NotificationProvider) {
+    NotificationProvider.setOptions({
+      delay: 2000,
+      startTop: 20,
+      startRight: 10,
+      verticalSpacing: 20,
+      horizontalSpacing: 20,
+      positionX: 'right',
+      positionY: 'bottom'
+    });
+  });
 }(window));
--- a/modules/users/client/controllers/admin/user.client.controller.js
+++ b/modules/users/client/controllers/admin/user.client.controller.js
@@ -5,9 +5,9 @@
     .module('users.admin')
     .controller('UserController', UserController);
 
-  UserController.$inject = ['$scope', '$state', '$window', 'Authentication', 'userResolve'];
+  UserController.$inject = ['$scope', '$state', '$window', 'Authentication', 'userResolve', 'Notification'];
 
-  function UserController($scope, $state, $window, Authentication, user) {
+  function UserController($scope, $state, $window, Authentication, user, Notification) {
     var vm = this;
 
     vm.authentication = Authentication;
@@ -22,9 +22,11 @@
           user.$remove();
 
           vm.users.splice(vm.users.indexOf(user), 1);
+          Notification.success('User deleted successfully!');
         } else {
           vm.user.$remove(function () {
             $state.go('admin.users');
+            Notification.success({ message: '<i class="glyphicon glyphicon-ok"></i> User deleted successfully!' });
           });
         }
       }
@@ -43,8 +45,9 @@
         $state.go('admin.user', {
           userId: user._id
         });
+        Notification.success({ message: '<i class="glyphicon glyphicon-ok"></i> User saved successfully!' });
       }, function (errorResponse) {
-        vm.error = errorResponse.data.message;
+        Notification.error({ message: errorResponse.data.message, title: '<i class="glyphicon glyphicon-remove"></i> User update error!' });
       });
     }
 
--- a/modules/users/client/controllers/password.client.controller.js
+++ b/modules/users/client/controllers/password.client.controller.js
@@ -5,9 +5,9 @@
     .module('users')
     .controller('PasswordController', PasswordController);
 
-  PasswordController.$inject = ['$scope', '$stateParams', 'UsersService', '$location', 'Authentication', 'PasswordValidator'];
+  PasswordController.$inject = ['$scope', '$stateParams', 'UsersService', '$location', 'Authentication', 'PasswordValidator', 'Notification'];
 
-  function PasswordController($scope, $stateParams, UsersService, $location, Authentication, PasswordValidator) {
+  function PasswordController($scope, $stateParams, UsersService, $location, Authentication, PasswordValidator, Notification) {
     var vm = this;
 
     vm.resetUserPassword = resetUserPassword;
@@ -22,7 +22,6 @@
 
     // Submit forgotten password account id
     function askForPasswordReset(isValid) {
-      vm.success = vm.error = null;
 
       if (!isValid) {
         $scope.$broadcast('show-errors-check-validity', 'vm.forgotPasswordForm');
@@ -37,7 +36,6 @@
 
     // Change user password
     function resetUserPassword(isValid) {
-      vm.success = vm.error = null;
 
       if (!isValid) {
         $scope.$broadcast('show-errors-check-validity', 'vm.resetPasswordForm');
@@ -55,13 +53,13 @@
     function onRequestPasswordResetSuccess(response) {
       // Show user success message and clear form
       vm.credentials = null;
-      vm.success = response.message;
+      Notification.success({ message: response.message, title: '<i class="glyphicon glyphicon-ok"></i> Password reset email sent successfully!' });
     }
 
     function onRequestPasswordResetError(response) {
       // Show user error message and clear form
       vm.credentials = null;
-      vm.error = response.data.message;
+      Notification.error({ message: response.data.message, title: '<i class="glyphicon glyphicon-remove"></i> Failed to send password reset email!', delay: 4000 });
     }
 
     function onResetPasswordSuccess(response) {
@@ -70,12 +68,13 @@
 
       // Attach user profile
       Authentication.user = response;
+      Notification.success({ message: '<i class="glyphicon glyphicon-ok"></i> Password reset successful!' });
       // And redirect to the index page
       $location.path('/password/reset/success');
     }
 
     function onResetPasswordError(response) {
-      vm.error = response.data.message;
+      Notification.error({ message: response.data.message, title: '<i class="glyphicon glyphicon-remove"></i> Password reset failed!', delay: 4000 });
     }
   }
 }());
--- a/modules/users/client/controllers/settings/change-password.client.controller.js
+++ b/modules/users/client/controllers/settings/change-password.client.controller.js
@@ -5,9 +5,9 @@
     .module('users')
     .controller('ChangePasswordController', ChangePasswordController);
 
-  ChangePasswordController.$inject = ['$scope', '$http', 'Authentication', 'UsersService', 'PasswordValidator'];
+  ChangePasswordController.$inject = ['$scope', '$http', 'Authentication', 'UsersService', 'PasswordValidator', 'Notification'];
 
-  function ChangePasswordController($scope, $http, Authentication, UsersService, PasswordValidator) {
+  function ChangePasswordController($scope, $http, Authentication, UsersService, PasswordValidator, Notification) {
     var vm = this;
 
     vm.user = Authentication.user;
@@ -16,7 +16,6 @@
 
     // Change user password
     function changeUserPassword(isValid) {
-      vm.success = vm.error = null;
 
       if (!isValid) {
         $scope.$broadcast('show-errors-check-validity', 'vm.passwordForm');
@@ -31,13 +30,12 @@
 
     function onChangePasswordSuccess(response) {
       // If successful show success message and clear form
-      $scope.$broadcast('show-errors-reset', 'vm.passwordForm');
-      vm.success = true;
+      Notification.success({ message: '<i class="glyphicon glyphicon-ok"></i> Password Changed Successfully' });
       vm.passwordDetails = null;
     }
 
     function onChangePasswordError(response) {
-      vm.error = response.data.message;
+      Notification.error({ message: response.data.message, title: '<i class="glyphicon glyphicon-remove"></i> Password change failed!' });
     }
   }
 }());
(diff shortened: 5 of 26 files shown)


modules/articles/client/controllers/admin/article.client.controller.js
```diff
@@ -20,7 +20,10 @@
     // Remove existing Article
     function remove() {
       if ($window.confirm('Are you sure you want to delete?')) {
-        vm.article.$remove($state.go('admin.articles.list'));
+        vm.article.$remove(function() {
```
sujeethk: Fix for #1274

modules/core/client/app/config.js
```diff
@@ -20,4 +20,16 @@
     // Add the module to the AngularJS configuration file
     angular.module(applicationModuleName).requires.push(moduleName);
   }
+
+  // Toastr Config Global settings. For more customization check https://github.com/Foxandxss/angular-toastr
+  angular.module('toastr').config(function(toastrConfig) {
```
sujeethk: Toastr config

modules/core/server/controllers/core.server.controller.js
```diff
@@ -14,6 +14,7 @@ exports.renderIndex = function (req, res) {
       displayName: validator.escape(req.user.displayName),
       provider: validator.escape(req.user.provider),
       username: validator.escape(req.user.username),
+      _id: req.user._id,
```
sujeethk: Fix #1531

mleanos: This would indeed fix the issue. However, we purposely removed the `_id` field from the client-side User object for reasons of security. I don't see an easy way to handle this on the client-side.

One thing for sure is that we need to have a check on the server-side, to prevent a user from deleting their own account. This can easily be done [here](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/admin.server.controller.js#L46) by adding something like:
```javascript
if (user._id === req.user._id) {
  return res.status(400).send({
      message: 'You cannot delete yourself!'
    });
}
```

Additionally, we can do something similar to [this](https://github.com/meanjs/mean/blob/master/modules/articles/server/controllers/articles.server.controller.js#L36-L38). Which we would then have a field attached to the User that is being "edited" that specifies whether or not the current active User is the same as the one being edited. This field could be named something like `isCurrentUserSelf` (or something more clever), and this field would NOT be persisted to the database.

mleanos:
I'm not 100% sold on the need for adding a new library to the framework, just to have a toastr feature. However, this implementation is pretty light-weight & if others are find with adding this, then I'm OK with it.

@meanjs/contributors @Wuntenn @hyperreality @simison WDYT?

modules/core/server/controllers/core.server.controller.js
sujeethk: I wanted to keep the change simple. How about this?

`<a class="btn btn-primary" ng-click="vm.remove()" ng-if="vm.user.username !== vm.authentication.user.username">`

This should fix the issue + no extra code change needed except changing the _id to username.

simison:
Toastr seems like a great library otherwise but it requires jQuery. :-(

In general some kind of notifications seem like a basic enough feature for this boilerplate.

simison:
It seems like this PR is trying to fix many issues and should be divided to 3 different PRs, or did I understand something wrong?

mleanos:
@simison I agree about separating this out into multiple PR's.

So the `angular-toastr` library uses JQuery? That is a bummer.

I'd rather see a more robust notification system with SocketIO. Although with my app I use a Notification service that uses SocketIO, which pops toasts. It was a big implementation, and I'm not even sure I handled it all correctly. 

Having some sort of notification would be great though.

simison:
Similar (size/features/maintained) to Toastr but angular-only: https://github.com/alexcrack/angular-ui-notification

Wuntenn:
I finding toasts really useful in MD and @simison's last link (angular-ui-notification) doesn't need JQuery which is a plus. They make all kinds of errors more consistent across the site and save loads of unnecessary headache wondering how to present stuff - I'd say it's a good call.

I removed UIB from my project and replaced it with Angular Material Design a few months back. There's quite a lot of tools that are directly interchangeable with UIB's own and toasts are one of the additional features that I didn't realise how good they were until I started using them.

I'm not sure if or when a move to MD is on the cards for MeanJS (I'd be happy to put together a PR) It's worth looking into. You also get a load of ARIA support and notifications in the console if you don't meet a requirement (usually this means filling in some element attribute). This means you get to create more inclusive experiences as you go without tons of upfront research into ARIA.

sujeethk:
Split the PR, this one takes care of the notification. I can open seperate PR for the fixes. 

Took your advice and switched library to angular-ui-notification and kept it simple

modules/core/server/controllers/core.server.controller.js
mleanos: https://github.com/meanjs/mean/pull/1535

mleanos:
> @mleanos What's the benefit of having socket.io notifications over just Angular controller handling them?

@simison I have notifications that come from the server, not just the client-side. I opted for a client-side service that "pops" toasts (which uses a toastr-like angular service) that can be triggered from my Socket.IO event listeners, or by merely calling `Notifications.pop()` that accepts additional options (e.g. data, type, actions). So while neither my Socket.IO implementation nor my Toaster service, relies on the other, my notifications use both where needed.

I guess after hearing myself explain this, it's more a specific requirement of my application. For this framework, it makes sense to just keep it simple as @sujeethk has done.

mleanos:
👍 on using angular-ui-notification. I'll probably end up switching my application to use it.

modules/articles/client/controllers/admin/article.client.controller.js
```diff
@@ -21,6 +21,7 @@
     function remove() {
       if ($window.confirm('Are you sure you want to delete?')) {
         vm.article.$remove($state.go('admin.articles.list'));
+        Notification.success('Article deleted successfully!');
```
mleanos: @sujeethk After merging #1534 this is creating a merge conflict. It's probably best to put this in the `$remove` callback. Can you rebase, and this line?

modules/users/client/controllers/admin/user.client.controller.js
```diff
@@ -42,8 +44,10 @@
         $state.go('admin.user', {
           userId: user._id
         });
+        Notification.success('User updated successfully!');
       }, function (errorResponse) {
         vm.error = errorResponse.data.message;
+        Notification.error('User update error!');
```
mleanos: @sujeethk Do we still need to set `vm.error` in these cases? Or would it be better to add the response error as the body of the toast, and keep "User update error!" as the title?

sujeethk: @mleanos I had it set that way on the angular-toastr, looks like i kept it generic on this one. It is good idea to have the error on the toast and add title.

Removing vm.error would fail some tests that are checking it. Didn't find any harm in having it until we decided to use these notifications.

I will update the error message with title and post a screenshot. Will update tests, although frankly I have no idea how the test scripts are setup(never looked other than running them).

sujeethk:
Let me know if I over did it. I believe all form submission feedback is now using the new notifications.

- Added 3 User e2e tests that check the new popup content
- vm.error or vm.success is left as is. Since they are part of the unit testing for controllers. Removing them will prevent a few error case scenarios from being tested
- But removed the error/success blocks from the view files and removed a couple of success/invalid pages in the reset password flow and related routes

I do not have an email service account to do some manual testing around the reset flows. Can someone with access please let me know if those flows are good.

Some screenshots:
success
error

jloveland:
Can you post a picture of this working on mobile?

jloveland:
seems to be working nicely on mobile for the login screen:
image1

modules/users/client/config/users.client.routes.js
```diff
@@ -101,20 +101,6 @@
         url: '/reset',
         template: '<ui-view/>'
       })
-      .state('password.reset.invalid', {
```
mleanos: The removal of this route is causing a 400 response on the server, with [this redirect](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js#L115). 

This keeps the User on the Reset Password view if they click on an invalid reset link (e.g. the token is expired). We should keep this route since the User should be taken away from the Reset Password view, in the case of an invalid token; they wouldn't be able to do anything there anyways.

I see there are some good reasons to get rid of the *Success* route, but by keeping it we provide more flexibility for our users. And we could always redirect somewhere else in the case of an *Invalid* token, but I don't see a need to change this flow now.

IMO, we should keep the flow with these two routes. There can probably be some improvements to the Reset Password feature, but I think it's a bit out of scope for this PR.

modules/users/tests/client/users.client.routes.tests.js
```diff
@@ -257,44 +257,6 @@
         });
       });
 
-      describe('Invalid Route', function () {
```
mleanos: Keep these tests if we end up keeping the client-side routes; based on my comment above.

modules/users/tests/client/password.client.controller.tests.js
```diff
@@ -189,7 +189,7 @@
           });
 
           it('should redirect to password reset success view', function() {
-            expect($location.path).toHaveBeenCalledWith('/password/reset/success');
```
mleanos: Related to my comment above.

modules/users/client/controllers/settings/change-password.client.controller.js
```diff
@@ -33,11 +33,13 @@
       // If successful show success message and clear form
       $scope.$broadcast('show-errors-reset', 'vm.passwordForm');
       vm.success = true;
```
mleanos: Since these aren't used by the view any longer, we should remove them from the controller.

@sujeethk I know you mentioned having issues with the client-side tests that relied on these `vm` fields, but they you should be able to resolve them. Let me know if you have trouble, and I can take a look.

modules/users/client/config/users.client.routes.js
sujeethk: @mleanos really good feedback, appreciate it! 

I redirected the success to home with a toast. Looks like I missed the error route. Honestly I haven't see these flows in action as I don't have a MSP. Can someone recommend some simple free MSP so I can test these flows?

I simply removed those routes and deleted those two view files to reduce footprint. In those terms I could have the invalid scenario redirect to forgot password screen with an invalid/expired token toast so end-user can request another.

Or the opposite being what you recommend to keep the routes and views in case users need them.

I am neutral to both scenarios. Which one do you prefer?

mleanos: I'd prefer to keep the routes/views. We could always open up a discussion on how to change this flow, but I think having that discussion would distract us from the purpose of this PR.

I'm open to the discussion though, if others feel this should be addressed now.

modules/users/client/controllers/settings/change-password.client.controller.js
sujeethk: @mleanos the couple of lines you mentioned above have been removed with no issues with tests. But there are a few tests on other controllers that check for the status of vm.success and vm.error.

```javascript
it('should set success to response message', function() {
  expect(scope.vm.success).toBe(successMessage);
});
```

Do you want me to remove all of those tests? If not can you please let me know how I can check for a DOM element from the test script?

mleanos: Is it possible to modify this test (and others like it) to check that the toast was opened?

simison: Indeed, otherwise nothing is testing if the notification actually worked.

sujeethk: @mleanos @simison I added about 3 tests on user e2e that is checking for the notification and it's contents.

But I am stuck on these client controller tests. If we remove the vm.error from the controller and intend to continue unit testing the controllers for error we would need to check for the notification. The notification is DOM element which I don't have access to without using jasmine jquery. I know we can access DOM in protractor  e2e tests. But any guidance on how I can check for the notification on the controller tests?

mleanos: Im not at a computer right now so i cant give examples, or links..

In some of our Karma tests, you'll see we have a Spy on some of the angular controller services. You can add some in these tests that ensure the Notification service was called; I forget the syntax. If a simple spy doesnt work, you can look into Sinon's Stub feature.

sujeethk: Sure will take a look. Thank you.

mleanos: https://github.com/meanjs/mean/blob/master/modules/core/tests/client/interceptors/auth-interceptor.client.tests.js#L26
https://github.com/meanjs/mean/blob/master/modules/core/tests/client/interceptors/auth-interceptor.client.tests.js#L49

Something similar can be done. The first links shows how to setup the Spy to watch the specific service method. You can setup a Spy for each Notification method that you expect to be called; in this case, `Notification.error()` & `Notification.success()`.

sujeethk: Removed all vm.success and vm.error and replaced any existing tests for error or success to check for Notification called. All tests passed on my local. Please review carefully :), first time using Jasmine spies.

mleanos: Ok. Thank you. I'll pull this branch down again, and test locally.

modules/users/client/controllers/admin/user.client.controller.js
```diff
@@ -42,8 +44,9 @@
         $state.go('admin.user', {
           userId: user._id
         });
+        Notification.success({ message: '<i class="glyphicon glyphicon-ok"></i> User deleted successfully!' });
```
mleanos: Should be "User saved successfully!"

modules/users/client/config/users.client.routes.js
```diff
@@ -115,6 +101,20 @@
           pageTitle: 'Password reset success'
         }
       })
+      .state('password.forgot', {
```
mleanos: Not critical, but can these state's be added back to where they were before these changes. It's causing an unnecessary diff. This can be misleading when looking back at the history.

modules/users/client/controllers/password.client.controller.js
```diff
@@ -55,13 +53,13 @@
     function onRequestPasswordResetSuccess(response) {
       // Show user success message and clear form
       vm.credentials = null;
-      vm.success = response.message;
+      Notification.success({ message: response.message, title: 'Password request successful!' });
```
mleanos: Very nit-picky, but can be "Password Reset email successfully sent" or something more user friendly.

modules/users/client/controllers/password.client.controller.js
```diff
@@ -55,13 +53,13 @@
...
     }
 
     function onRequestPasswordResetError(response) {
       // Show user error message and clear form
       vm.credentials = null;
-      vm.error = response.data.message;
+      Notification.error({ message: response.data.message, title: '<i class="glyphicon glyphicon-remove"></i> Password request failed!', delay: 4000 });
```
mleanos: Similar to my nitty comment above.

mleanos:
@sujeethk Just a couple of minor changes/feedback, and this should be ready.

modules/users/client/controllers/password.client.controller.js
sujeethk: @mleanos That text is just the title of the error notification. The actual message from the server would be the message like(which is already in the server controllers):
`An email has been sent to the provided email with further instructions.`
`No account with that username has been found`
So the title would need to be generic to what the request is for right?

sujeethk:
@mleanos On a side note, I have been thinking to add an error notification into the `auth-interceptor.client.service` for error code `-1`.

This will show an error notification if the user lost his connection or the request timed-out or God forbid the server goes down. In these cases the client controllers will not be able to show their notifications(or vm.error in current master) as there is no response. Even if it is called res.data.message is not available, and there would be an angular error posted to the console(which probably the client isnt watching). So end user experience is that he clicks submit and waits, nothing happens.

Do you think it is good to have? If so, do you think I can add it in this PR or does it need a seperate PR?

modules/users/client/controllers/password.client.controller.js
mleanos: Sure. I understand. Just thought that the title wasn't very clear, or descriptive, for a user. Feel free to leave it as is.

mleanos:
@sujeethk Let's leave the auth-interceptor Notifications for a separate PR & discussion. It would be better to wrap this PR up, to get us closer to the 0.5.0 release.

modules/users/client/controllers/password.client.controller.js
sujeethk: image

@mleanos Did you mean a title like this? I thought the title and the message were decoupled as the email was not attempted to be sent, and so kept the title generic. But if this image makes more sense let me know and I can update the text.

sujeethk: On second thoughts, I actually get your point. The title is misleading. User is not requesting for a password. He is only requesting for a password reset email. I will add a couple of words and commit.

mleanos:
@meanjs/contributors @simison @jloveland LGTY?

codydaig:
My only question: In most controllers, I see `Notification.sucess()` after `$state.go()`. Will that ever display and is that just excess code?

mleanos:
@codydaig The Notification method still gets called & will display, after we transition to the next state.

Since the Notification is applied to the root-scope, the toast is independent of what current state we're on; the toast ends up showing up during our state transition, and remains after the new state is loaded.

codydaig:
@mleanos Thanks for the clarification.

mleanos:
@sujeethk After merging #1499 there are merge conflicts with this PR. They should be pretty minor. Can you resolve them, and then we can merge this. I apologize for the inconvenience.

sujeethk:
@mleanos No worries. Merged and committed.

mleanos:
@sujeethk Couple more conflicts. If you can rebase one more time, I'll merge right away.

sujeethk:
@mleanos I am currently our of station on vacation. Will rebase and merge Monday night. Apologies for the delay

sujeethk:
@mleanos Rebased

mleanos:
@sujeethk Thank you! Merging now.

@simison Thanks for your feedback as well.

PR #1535, opened 2016-09-30
fix(user): hide delete button for self

### LINKED ISSUE #1531: Admin user able to delete self

sujeethk:
fix(user): hide delete button for self

Remove delete button for self

Fixes #1531

### DIFF
--- a/modules/users/client/views/admin/view-user.client.view.html
+++ b/modules/users/client/views/admin/view-user.client.view.html
@@ -8,7 +8,7 @@
         <a class="btn btn-primary" ui-sref="admin.user-edit({userId: vm.user._id})">
           <i class="glyphicon glyphicon-edit"></i>
         </a>
-        <a class="btn btn-primary" ng-click="vm.remove()" ng-if="vm.user._id !== vm.authentication.user._id">
+        <a class="btn btn-primary" ng-click="vm.remove()" ng-if="!vm.isContextUserSelf()">
           <i class="glyphicon glyphicon-trash"></i>
         </a>
       </div>
--- a/modules/users/client/controllers/admin/user.client.controller.js
+++ b/modules/users/client/controllers/admin/user.client.controller.js
@@ -14,6 +14,7 @@
     vm.user = user;
     vm.remove = remove;
     vm.update = update;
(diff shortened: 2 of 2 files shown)


modules/users/client/views/admin/view-user.client.view.html
```diff
@@ -8,7 +8,7 @@ <h1 ng-bind="vm.user.username"></h1>
         <a class="btn btn-primary" ui-sref="admin.user-edit({userId: vm.user._id})">
           <i class="glyphicon glyphicon-edit"></i>
         </a>
-        <a class="btn btn-primary" ng-click="vm.remove()" ng-if="vm.user._id !== vm.authentication.user._id">
+        <a class="btn btn-primary" ng-click="vm.remove()" ng-if="vm.user.username !== vm.authentication.user.username">
```
mleanos: Using `username` here is OK. It always feels weird to me though, to compare using a field like this, vs the actual `_id` or some other more intentional field.

Instead of doing the comparison on this field in the template, how about we create a flag that we manage in the controller using `vm`. Something like: `vm.isCurrentUserSelf = isSelf();` in the controller. By doing it this way, we can avoid the need to update the template if we ever change how we're determined what "self" means in this context.

mleanos: @sujeethk Let me know how you feel about my suggestion. Again, thanks for the help.

mleanos: Or maybe a better name for the view-model field would be `vm.isContextUserSelf` or `vm.isEditedUserSelf`.. "isCurrentUserSelf" seems confusing.

sujeethk: @mleanos did you mean client controller or server controller? reason I ask is the existing 'isCurrentUserOwner' is done on the server controller. I believe you mean client controller as you referred vm and I am ok with that approach. Just wanted to make sure you are ok if it is not consistent across

mleanos: Yep. I meant client-side. Just to keep the logic out of the view.

sujeethk:
Updated per comments. Hoping coveralls doesn't drop.

PR #1536, opened 2016-09-30
feat(core): Add prod asset versioning

### LINKED ISSUE #1509: Include asset versioning
Since MEAN app depend heavily on Javascript files, it would be better if assets like CSS, JS and image files are versioned if they are changed. It helps it busting cache issues. In ASP.NET bundling provide minification and versioning. I was searching the same for MEAN and reached to [gulp-rev](https://github.com/sindresorhus/gulp-rev) npm module. But I am new to MEAN and I find it hard to integrate. If any one has used then please guide me for it. I also request to integrate it MEANJS itself.

sujeethk:
feat(core): Add prod asset versioning

Asset versioning for prod - idea from #1509

Fixes #1509

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,6 +24,7 @@ var _ = require('lodash'),
   protractor = require('gulp-protractor').protractor,
   webdriver_update = require('gulp-protractor').webdriver_update,
   webdriver_standalone = require('gulp-protractor').webdriver_standalone,
+  del = require('del'),
   KarmaServer = require('karma').Server;
 
 // Local settings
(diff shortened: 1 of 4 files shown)


sujeethk:
maxAge is currently set to 1 day for all the static routes assuming vendor assets will not change in a day. 

Do we need asset versioning for vendor assets as well? If so, they would have to be moved to dist folder and versioned. If this is accomplished then maxAge can be set to a year or more.

gulpfile.js
```diff
@@ -24,6 +24,8 @@ var _ = require('lodash'),
   protractor = require('gulp-protractor').protractor,
   webdriver_update = require('gulp-protractor').webdriver_update,
   webdriver_standalone = require('gulp-protractor').webdriver_standalone,
+  rev = require('gulp-rev'),
```
rhutchison: why not use plugins.rev() below?

sujeethk: Valid point :)

lirantal:
Looks good, thanks.

PR #1543, opened 2016-10-05
grunt(core): Remove Excess Grunt Dependencies

codydaig:
@sujeethk Just leftover. I must have missed them in the rebase. Thanks for catching that!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ RUN sudo apt-get install -yq nodejs \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install MEAN.JS Prerequisites
-RUN npm install --quiet -g grunt-cli gulp bower yo mocha karma-cli pm2 && npm cache clean
+RUN npm install --quiet -g gulp bower yo mocha karma-cli pm2 && npm cache clean
 
 RUN mkdir -p /opt/mean.js/public/lib
 WORKDIR /opt/mean.js
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ RUN sudo apt-get install -yq nodejs \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Install gem sass for grunt-contrib-sass
+# Install gem sass
 RUN gem install sass
 
 # Install MEAN.JS Prerequisites
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,9 +50,6 @@ RUN sudo apt-get install -yq nodejs \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Install gem sass
-RUN gem install sass
-
 # Install MEAN.JS Prerequisites
 RUN npm install --quiet -g grunt-cli gulp bower yo mocha karma-cli pm2 && npm cache clean
 
--- a/Dockerfile-production
+++ b/Dockerfile-production
@@ -47,7 +47,7 @@ RUN sudo apt-get install -yq nodejs \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Install gem sass for grunt-contrib-sass
+# Install gem sass
 RUN gem install sass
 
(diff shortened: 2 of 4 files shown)


mleanos:
LGTM

codydaig:
My only other question, is should this line be removed. I don't think it will break anything, but want to be sure. https://github.com/meanjs/mean/blob/8b5466961430627356dd6967b296c8528c044e23/config/assets/default.js#L56

sujeethk:
@codydaig I dont see any reference in the code to that key. But I will leave it up to the contributors to confirm.

Please see below references for Grunt still on the latest master after the other pull was merged.
[Dockerfile#L53](https://github.com/meanjs/mean/blob/master/Dockerfile#L53)
[Dockerfile#L57](https://github.com/meanjs/mean/blob/master/Dockerfile#L57)

[Dockerfile-production#L50](https://github.com/meanjs/mean/blob/master/Dockerfile-production#L50)
[Dockerfile-production#L54](https://github.com/meanjs/mean/blob/master/Dockerfile-production#L54)

[package.json#L113](https://github.com/meanjs/mean/blob/master/package.json#L113)

sujeethk:
@codydaig Looks like the 'gruntConfig' key you mentioned was used only in the gruntfile.js @ L37, L112, L125

Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN sudo apt-get install -yq nodejs \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Install gem sass for grunt-contrib-sass
+# Install gem sass
 RUN gem install sass
```
codydaig: Is this even needed??

Dockerfile-production
```diff
@@ -47,7 +47,7 @@ RUN sudo apt-get install -yq nodejs \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Install gem sass for grunt-contrib-sass
+# Install gem sass
 RUN gem install sass
```
codydaig: Same here, is this even needed anymore?

codydaig: @mleanos It's not Docker. Do users still have to have sass installed through ruby on their computer for it to work?

mleanos: I'm just shooting from the hip here, but I *think* that was only required for older version of Node. < 0.10.x?? I haven't looked into that in a while though.

sujeethk: I am on Win10 with no Sass/Ruby installed and I just did a simple scss file for core and gulp compiled it to css and i see the style change.

mleanos: When I first started using this project I was on Windows 7 & had Node v0.10.x installed; I was required to install ruby/sass with that setup. Now I'm using Windows 10 & Node v6.6.0.

What @sujeethk is saying makes sense, if it's indeed a Node version issue.

@sujeethk Just out of curiosity, what Node version are you running?

sujeethk: @mleanos I use NVM to switch versions. I checked on v6.3.1 and v4.5.0 and it works fine. Although when you switch versions you would have to rebuild node-sass using `npm rebuild node-sass`. But everything works fine(including sass compile to css and i see those styles applied) without need for installing ruby or sass.

I am not able to check on v0.12.0 as there is a build error. Honestly we should update node engine to remove support for those old versions.

lirantal:
Hey guys, looks good and indeed all the references from @sujeethk should be removed as well with regards to their grunt reference.

Dockerfile-production
lirantal: @mleanos @codydaig that gem install isn't required anymore for gulp.
I deprecated it when I introduced the node-sass library which is purely C based and doesn't need the ruby gem dependency

codydaig:
Changes have been made. WIth some LGTM, I will squash on merge.

lirantal:
@codydaig can you also remove these:
https://github.com/meanjs/mean/blob/master/package.json#L115
And in the following remove the grunt-cli global npm lib:
https://github.com/meanjs/mean/blob/master/Dockerfile#L57
https://github.com/meanjs/mean/blob/master/Dockerfile-production#L54

codydaig:
@lirantal Done

lirantal:
Great. Merge whenever you're ready :)

PR #1544, opened 2016-10-06
fix(core): Remove the <base> tag (continues #1230)

### LINKED ISSUE #1224: <base href="/"> breaks SVG IRIs and element CSS URLs

mleanos:
@0x24a537r9
>Fixes #1224. Everything looks good in the browser and server/e2e tests, but unfortunately the unit tests are broken, and I'll probably need some guidance from someone more familiar with Angular/Karma.

This is a continuation of #1230, in an attempt to finalize the changes by fixing the broken tests.

NOTE: The fixes to the client-side tests aren't ideal. At the moment, they will suffice. [This comment](https://github.com/angular-ui/ui-router/issues/212#issuecomment-60803437),
among others in that issue, led me to choose this method as the fix to
avoid having to change any core code.

### DIFF
--- a/modules/articles/client/config/articles-admin.client.routes.js
+++ b/modules/articles/client/config/articles-admin.client.routes.js
@@ -15,7 +15,7 @@
         template: '<ui-view/>'
       })
       .state('admin.articles.list', {
-        url: '/',
+        url: '',
         templateUrl: '/modules/articles/client/views/admin/list-articles.client.view.html',
         controller: 'ArticlesAdminListController',
         controllerAs: 'vm',
--- a/modules/articles/client/config/articles-admin.client.routes.js
+++ b/modules/articles/client/config/articles-admin.client.routes.js
@@ -15,8 +15,8 @@
         template: '<ui-view/>'
       })
       .state('admin.articles.list', {
-        url: '',
-        templateUrl: 'modules/articles/client/views/admin/list-articles.client.view.html',
+        url: '/',
+        templateUrl: '/modules/articles/client/views/admin/list-articles.client.view.html',
         controller: 'ArticlesAdminListController',
         controllerAs: 'vm',
         data: {
@@ -25,7 +25,7 @@
       })
       .state('admin.articles.create', {
         url: '/create',
-        templateUrl: 'modules/articles/client/views/admin/form-article.client.view.html',
+        templateUrl: '/modules/articles/client/views/admin/form-article.client.view.html',
         controller: 'ArticlesAdminController',
         controllerAs: 'vm',
         data: {
@@ -37,7 +37,7 @@
       })
       .state('admin.articles.edit', {
         url: '/:articleId/edit',
-        templateUrl: 'modules/articles/client/views/admin/form-article.client.view.html',
+        templateUrl: '/modules/articles/client/views/admin/form-article.client.view.html',
         controller: 'ArticlesAdminController',
         controllerAs: 'vm',
         data: {
--- a/modules/articles/client/config/articles.client.routes.js
+++ b/modules/articles/client/config/articles.client.routes.js
@@ -16,7 +16,7 @@
       })
       .state('articles.list', {
         url: '',
-        templateUrl: 'modules/articles/client/views/list-articles.client.view.html',
+        templateUrl: '/modules/articles/client/views/list-articles.client.view.html',
         controller: 'ArticlesListController',
         controllerAs: 'vm',
         data: {
@@ -25,7 +25,7 @@
       })
       .state('articles.view', {
         url: '/:articleId',
-        templateUrl: 'modules/articles/client/views/view-article.client.view.html',
+        templateUrl: '/modules/articles/client/views/view-article.client.view.html',
         controller: 'ArticlesController',
         controllerAs: 'vm',
         resolve: {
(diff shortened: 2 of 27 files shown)


lirantal:
@mleanos thanks.

mleanos:
@lirantal Yes. We can close to the other.

lirantal:
Ok I'll close the older one. Can you rebase your branch here with the latest changes?

mleanos:
I'm waiting for #1547 & #1532 to be merged, before I rebase. Those PR's should go in first, and will cause conflicts with this. I'll fix as soon as they go in.

mleanos:
@lirantal This has been rebased.

lirantal:
It's a big change through a lot of files so I want to test it a bit locally to confirm it's ok not breaking anything first and then I'll merge.

mleanos:
Also, when you merge, don't squash. I noticed with the other PR that I took over, the commits from the original contributor were squashed into mine; thus, the original author's commits aren't reflected in the history.

If you'd prefer, I can squash my commits down to 1 so there would be only two going in with this merge. Either way, it's not a big deal to me but we should try to at least keep @0x24a537r9's commits.

lirantal:
Of course, no problems.

lirantal:
Seems to work great, thanks @mleanos and @0x24a537r9 for all the help!

abumere:
Looks like removing the <base> tag also broke the ability to add a URL in the config/assets/default or production files. 

Normally you could add a URL as a css dependency (lets say to a google font) but now it appends localhost to the beginning of the URL

PR #1545, opened 2016-10-06
feat(users): change username to usernameOrEmail in signin

itelo:
It's quite boring to force your users to remember the username they've created in your app to signin. So I've made an option to put an email or username. It's far more interesting to user experience, and a lot of applications are doing that, eg. Snapchat.

### DIFF
--- a/modules/users/server/config/strategies/local.js
+++ b/modules/users/server/config/strategies/local.js
@@ -10,12 +10,16 @@ var passport = require('passport'),
 module.exports = function () {
   // Use local strategy
   passport.use(new LocalStrategy({
-    usernameField: 'username',
+    usernameField: 'usernameOrEmail',
     passwordField: 'password'
   },
-  function (username, password, done) {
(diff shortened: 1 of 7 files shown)


modules/users/server/config/strategies/local.js
```diff
@@ -10,12 +10,16 @@ var passport = require('passport'),
...
     passwordField: 'password'
   },
-  function (username, password, done) {
+  function (usernameOrEmail, password, done) {
     User.findOne({
-      username: username.toLowerCase()
+      $or: [{
+        username: usernameOrEmail.toLowerCase()
+      }, {
+        email: usernameOrEmail
```
simison: Emails are always saved lowercase in the db, so it's safe to set this `sernameOrEmail.toLowerCase()` as well.

simison:
We've been running pretty much identical signup controller for over a year now. Works great — no more "I forgot my username" support mails.

lirantal:
Looks good, thanks @itelo !

PR #1547, opened 2016-10-07
fix(core): Add custom 400 and 404 error messages

### LINKED ISSUE #1504: Error handling for articles

mleanos:
Supersedes #1510 

Other than the work put into this by @hyperreality, this PR adds:

Changed the error responses returned from the Sign Up & Sign In API
calls to use 422 rather than 400.

For insight into why this change was made:
https://github.com/meanjs/mean/pull/1510#issuecomment-247435378

For reference on why to use 422 over 400:
https://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm

Fixes #1504

### DIFF
--- a/modules/core/client/config/core.client.route-filter.js
+++ b/modules/core/client/config/core.client.route-filter.js
@@ -25,7 +25,7 @@
 
         if (!allowed) {
           event.preventDefault();
-          if (Authentication.user !== null && typeof Authentication.user === 'object') {
+          if (Authentication.user !== undefined && typeof Authentication.user === 'object') {
             $state.transitionTo('forbidden');
           } else {
             $state.go('authentication.signin').then(function () {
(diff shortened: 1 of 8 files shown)


lirantal:
LGTM, no issues on the 422 on my behalf.

PR #1551, opened 2016-10-08
feat(core): change console in angular to

### LINKED ISSUE #1541: console.log in controllers
Are these console.log in code necessary?

[users.password.server.controller.js#L65](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js#L65)
[users.password.server.controller.js#L71](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js#L71)
[articles.client.service.js](https://github.com/meanjs/mean/blob/master/modules/articles/client/services/articles.client.service.js#L50) - harmless yet unnecessary

Especially the one on the password server controller. It is logging the user's password reset link. I would assume it is so that one could test the reset flow even without MSP. But on a production env wouldn't it be a security issue?

itelo:
logs with $log in client idea proposed by @simison
also removed unnecessary consoles in [users.password.server.controller.js](https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.password.server.controller.js)

Fixes #1541

### DIFF
--- a/modules/articles/client/services/articles.client.service.js
+++ b/modules/articles/client/services/articles.client.service.js
@@ -5,9 +5,9 @@
     .module('articles.services')
     .factory('ArticlesService', ArticlesService);
 
-  ArticlesService.$inject = ['$resource'];
+  ArticlesService.$inject = ['$resource', '$log'];
 
-  function ArticlesService($resource) {
+  function ArticlesService($resource, $log) {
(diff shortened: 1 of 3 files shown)


simison:
LGTM

mleanos:
LGTM

PR #1553, opened 2016-10-09
fix(express): Incorrest uses of 400 error codes

mleanos:
Fixes incorrect usage of 400 HTTP responses being returned from the server, in favor of using 422.

Also, changed a couple return codes to 401 where it was more appropriate.

See this article for reasoning behind moving to 422, and why 400 isn't appropriate for these cases.
https://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm

Reference:
https://github.com/meanjs/mean/commit/6be12f8a061a2f7228e254c314a46ad7a78966d1

Related:
https://github.com/meanjs/mean/pull/1547
https://github.com/meanjs/mean/pull/1510

### DIFF
--- a/modules/articles/server/controllers/articles.server.controller.js
+++ b/modules/articles/server/controllers/articles.server.controller.js
@@ -17,7 +17,7 @@ exports.create = function (req, res) {
 
   article.save(function (err) {
     if (err) {
-      return res.status(400).send({
+      return res.status(422).send({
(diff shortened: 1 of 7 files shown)


mleanos:
In my haste to merge #1547, I forgot about the 400 error responses where we don't want to redirect the User. This resolves those issues.

PR #1560, opened 2016-10-12
promise(core): Add Promise Option to Mongoose

### LINKED ISSUE #1559: Mongoose 'default promise library deprecated' warning
Mongoose throws a warning on running tests.
```
Mongoose: mpromise (mongoose's default promise library) is deprecated, plug in your own promise library instead: http://mongoosejs.com/docs/promises.html
```

codydaig:
Fixes #1559

### DIFF
--- a/config/lib/mongoose.js
+++ b/config/lib/mongoose.js
@@ -29,6 +29,8 @@ module.exports.connect = function (cb) {
       console.log(err);
     } else {
 
+      mongoose.Promise = config.db.promise;
+
       // Enabling mongoose debug mode if required
       mongoose.set('debug', config.db.debug);
 
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -7,6 +7,9 @@ module.exports = {
     keywords: 'mongodb, express, angularjs, node.js, mongoose, passport',
     googleAnalyticsTrackingID: process.env.GOOGLE_ANALYTICS_TRACKING_ID || 'GOOGLE_ANALYTICS_TRACKING_ID'
   },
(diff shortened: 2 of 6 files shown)


config/lib/mongoose.js
```diff
@@ -29,6 +29,8 @@ module.exports.connect = function (cb) {
       console.log(err);
     } else {
 
+      mongoose.Promise = config.db.promise;
```
simison: Why not just?
```js
mongoose.Promise = global.Promise;
```

codydaig: I debated that and since things have been moving to be as configurable as possible, I figured I'd follow that trend. I'm not opposed to changing it though. 

You could in theory set it to be bluebird or q or any other promise library.

shanavas786: Isn't 

```js
db: { promise: global.Promise }
```
in `config/env/default.js` enough ?

mleanos: I'm sort of indifferent on whether or not to make this configurable. Personally, I like things to be as configurable as possible. However, in this case I can't think of any use case where having a different promise library to be used between, say, a dev env & test env.

If we do go with adding this into the environment configuration, I agree with @shanavas786 that having it in the default config is sufficient. This wouldn't stop someone from adding it to the others if they have a need to do so.

codydaig:
@shanavas786 @mleanos @simison I consolidated to the default.js.

mleanos:
@codydaig LGTM. Feel free to merge when you're ready.

PR #1564, opened 2016-10-13
feat: users : Abstract OAuth URL's

securityvoid:
feat: users : Abstract OAuth URL's & Move Scope to Strategy File

Changes the server routes for OAuth to utilize a request parameter so there is one pair of routes for all OAuth requests, verses one for every strategy (Twitter, Google, etc). To support this all scope definitions were moved from the routes to the actual strategy files. This keeps the strategy definitions isolated to the modules/users/server/config/strategies, which makes it much simpler to maintain and add new strategies.

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -83,18 +83,17 @@ exports.signout = function (req, res) {
 /**
  * OAuth provider call
  */
-exports.oauthCall = function (strategy, scope) {
-  return function (req, res, next) {
+exports.oauthCall = function (req, res, next) {
+    var strategy = req.params.strategy;
     // Authenticate
-    passport.authenticate(strategy, scope)(req, res, next);
+    passport.authenticate(strategy)(req, res, next);
   };
-};
 
 /**
  * OAuth callback
  */
-exports.oauthCallback = function (strategy) {
-  return function (req, res, next) {
+exports.oauthCallback = function (req, res, next) {
+    var strategy = req.params.strategy;
 
     // info.redirect_to contains inteded redirect path
     passport.authenticate(strategy, function (err, user, info) {
@@ -112,7 +111,6 @@ exports.oauthCallback = function (strategy) {
         return res.redirect(info.redirect_to || '/');
       });
     })(req, res, next);
-  };
 };
 
 /**
--- a/modules/users/server/config/strategies/facebook.js
+++ b/modules/users/server/config/strategies/facebook.js
@@ -14,7 +14,8 @@ module.exports = function (config) {
     clientSecret: config.facebook.clientSecret,
     callbackURL: config.facebook.callbackURL,
     profileFields: ['id', 'name', 'displayName', 'emails', 'photos'],
-    passReqToCallback: true
+    passReqToCallback: true,
+    scope: ['email']
   },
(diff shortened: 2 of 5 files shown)


modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -83,18 +83,17 @@ exports.signout = function (req, res) {
 /**
  * OAuth provider call
  */
-exports.oauthCall = function (strategy, scope) {
-  return function (req, res, next) {
+exports.oauthCall = function (req, res, next) {
+    var strategy = req.params.strategy;
     // Authenticate
-    passport.authenticate(strategy, scope)(req, res, next);
+    passport.authenticate(strategy)(req, res, next);
```
Wuntenn: Hey @securityvoid, I've just read the change again. I see what you did. I somehow assumed that I was seeing part of the config file but you actually set up the scope in the `passport.use()` call instead whilst creating the strategy. When I woke the next day I didn't re-read.

Nice! I've not seen scope set up anywhere else I dug though the source just to make sure it would be safe to rely on the use of [strategy](https://github.com/jaredhanson/passport-oauth2/blob/master/lib/strategy.js#L95) on the `method` for other strategies.

It should work in any strategy as long as they inherit from his OAuth2Strategy (as he does)[https://github.com/jaredhanson/passport-facebook/blob/master/lib/strategy.js#L63] when he created most of the common strategies or if they create a custom one with scope (which should be most strategies).

securityvoid: @Wuntenn No problem at all. I've also been working on getting the Azure-AD passport module integrated, which is when I read the Passport documentation and found out Scope was valid in the passport.use declaration. I wasn't positive about how the null would be handled on the passport.authenticate until I tested it though.

BTW, I should have another Pull request here soon with the passport-azure-ad module integrated. I've got that stuff done, but submitted a pull request to passport-azure-ad that I'm hoping they'll integrate first. I also wanted to see what happened with the two Pull requests I did here before sending another pull request, since it would change the number of files I'd have to modify.

mleanos: Integrating passport-azure-ad is a good idea. I'll be looking forward to seeing that come in.

We're just about ready to release 0.5.0. Which should be in the coming days.

We'll get this merged into *master* shortly after our release.

mleanos:
@securityvoid I apologize for losing track of this PR. I'm still very interested in seeing these changes implemented. Could you rebase from *master*, so we could do a final review. Also, do you still on submitting the changes introduced in #1566?

lirantal:
@mleanos just reviewed it, looks like a good change.
do you want to rebase for @securityvoid and push it?

mleanos:
@lirantal Thank you for reviewing. I can rebase? However, what's the best way to push it without creating a new PR? Should I just merge into my local *master* branch and push directly from that branch to upstream?

lirantal:
you should check it out to a local branch (preferably not master), then rebase/merge locally and push.

mleanos:
@securityvoid Thank you for these changes. I like them a lot :) It should make it easier to write tests for the various passport strategies.

@lirantal I merged into master, after fixing spacing issues & resolving a couple of merge conflicts. Everything looks good, and I've tested for a bit after the merge.

mleanos:
Closing since these changes were already merged manually.

PR #1566, opened 2016-10-13
feat: users - Dynamic OAuth / Social Providers

securityvoid:
## feat: users - Dynamic OAuth / Social Providers

This update adds a new REST call which pulls back a list of the available strategies which have settings defined within the env config (config\env). This REST end-point is then called from the user AuthenticationController in order to dynamically list strategies within authentication.client.view.html utilizing ng-repeat. The display attributes (Image, alt-text, oauth url) have default values based on the strategy name, but this can also be overridden in the environment configuration. 

With this pull request, and my pull request updating the router for Oauth calls; its now possible to add a new Oauth provider simply by adding a strategy.js, and adding the needed config items to the environment config. Everything else is dynamic. There is longer a need to update four places.

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -30,7 +30,13 @@ module.exports = {
   facebook: {
     clientID: process.env.FACEBOOK_ID || 'APP_ID',
     clientSecret: process.env.FACEBOOK_SECRET || 'APP_SECRET',
-    callbackURL: '/api/auth/facebook/callback'
+    callbackURL: '/api/auth/facebook/callback',
+    viewData : {
+      initURL : '/api/auth/facebook',
+      imageSRC : '/modules/users/client/img/buttons/facebook.png',
+      altText : 'Login with Facebook Today!',
+      name : 'Facebook'
+    }
   },
   twitter: {
     clientID: process.env.TWITTER_KEY || 'CONSUMER_KEY',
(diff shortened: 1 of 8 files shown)


securityvoid:
I looked at the checks that have failed, and I think they're all intentional. First set of tests failed because there is now a new HTTP Request (/api/auth/strategies). This is expected.

I'm not positive why the second one failed. The "coverage" test is harder to read. I think its something similar.

mleanos:
@securityvoid It looks like the commits from #1564 got into this PR. Perhaps it makes sense to combine the PR's.

Is this PR dependent on the other?

securityvoid:
@mleanos No, it is not directly dependent on the other. With the two, everything becomes completely dynamic. However, you could theoretically have the appearing/removal of Auth options even without the dynamic routes. With both of them in place it makes everything very dynamic, which I think makes a lot of sense, but the individual changes are independent.

Do you need me to do something different with the pull request so they don't come in as dependent? Sorry if I didn't bring these in correctly. These are actually the first two pull requests I've submitted.

mleanos:
@securityvoid This PR shouldn't have commits from your other open PR. Otherwise, it doesn't make sense to have them as separate PR's.

As this PR stands, it's difficult to review since it's not clear what the actual changes, and intent, is here.

securityvoid:
@mleanos Okay, any idea what the best way is to fix this? Do I need to do a new pull and kill this?

securityvoid:
@mleanos I think I fixed it. If there was a better way to do it than I did, please let me know. I've used the basics of git for awhile, but I'm very new to all this fancy "pulling out stuff".

securityvoid:
@mleanos I've now figured out how to make this stuff clean, but I couldn't seem to get it right on this branch. I'll resubmit the pull.

PR #1570, opened 2016-10-15
fix(config): Default config dangling comma

mleanos:
Removes the dangling comma from the default configuration db.promise
setting. This was causing a lint error.


mleanos:
Merging since it's a simple fix.

PR #1571, opened 2016-10-15
feat(users): get profile picture from provider in oauth signup

itelo:

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -160,7 +160,7 @@ exports.saveOAuthUserProfile = function (req, providerUserProfile, done) {
               lastName: providerUserProfile.lastName,
               username: availableUsername,
               displayName: providerUserProfile.displayName,
-              profileImageURL: providerUserProfile.profileImageURL,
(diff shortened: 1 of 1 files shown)


modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -160,7 +160,7 @@ exports.saveOAuthUserProfile = function (req, providerUserProfile, done) {
               lastName: providerUserProfile.lastName,
               username: availableUsername,
               displayName: providerUserProfile.displayName,
-              profileImageURL: providerUserProfile.profileImageURL,
+              profileImageURL: providerUserProfile.providerData.image.url,
```
mleanos: This field is only present on Google+. I tested with Facebook, Twitter, Google+, and GitHub. This breaks the authentication for each one that I tested, other than Google+. Also, the Google+ authenticates but the image reference is incorrect. The url has a "http://localhost" prepended to the url coming from the provider. **This actually seems to be an issue in the current master without this change**.

@itelo Can you look into whether or not there's a way to reconcile these urls, by inspecting the providerUserProfile fields that are returned for these provider?

PR #1573, opened 2016-10-15
fix(users): fix redirect when signup or add provider

itelo:

### DIFF
--- a/modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
+++ b/modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
@@ -5,15 +5,16 @@
     .module('users')
     .controller('SocialAccountsController', SocialAccountsController);
 
-  SocialAccountsController.$inject = ['$scope', 'UsersService', 'Authentication', 'Notification'];
+  SocialAccountsController.$inject = ['$state', '$window', 'UsersService', 'Authentication', 'Notification'];
 
-  function SocialAccountsController($scope, UsersService, Authentication, Notification) {
+  function SocialAccountsController($state, $window, UsersService, Authentication, Notification) {
     var vm = this;
 
     vm.user = Authentication.user;
     vm.hasConnectedAdditionalSocialAccounts = hasConnectedAdditionalSocialAccounts;
     vm.isConnectedSocialAccount = isConnectedSocialAccount;
     vm.removeUserSocialAccount = removeUserSocialAccount;
+    vm.callOauthProvider = callOauthProvider;
 
     // Check if there are additional accounts
     function hasConnectedAdditionalSocialAccounts() {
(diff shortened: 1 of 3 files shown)


modules/users/server/controllers/users/users.authentication.server.controller.js
```diff
@@ -85,6 +85,9 @@ exports.signout = function (req, res) {
  */
 exports.oauthCall = function (strategy, scope) {
   return function (req, res, next) {
+    if (req.query && req.query.redirect_to)
+      req.session.redirect_to = req.query.redirect_to
```
mleanos: missing semicolon

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
```diff
@@ -42,5 +43,15 @@
     function onRemoveSocialAccountError(response) {
       Notification.error({ message: response.message, title: '<i class="glyphicon glyphicon-remove"></i> Remove failed!' });
     }
+
+    // OAuth provider request
+    function callOauthProvider(url) {
+      if ($state.previous && $state.previous.href) {
+        url += '?redirect_to=' + encodeURIComponent($state.$current.url.prefix);
```
mleanos: Why is the prefix used here?

mleanos: I see that `#_=_` is added to the end of redirect_to querystring parameter.

mleanos:
@itelo This is mostly good. Just left a review requesting a change, and asking for clarification on one thing.

Also, please be in the habit of adding commit messages. It makes it easier to review in the commit history.

modules/users/client/controllers/settings/manage-social-accounts.client.controller.js
itelo: `$state.$current.url.prefix` contains /settings/accounts, is that your question?

mleanos: Using `prefix` does work, so I'm not questioning that. You're right though, that I was getting the two confused, and even so it wouldn't work to use `$state.current.url` for the exact reason you stated; it would return `/accounts`.

My confusion was reinforced by the fact that using `$state.$current.url` would work in this case due to the URI encoding converting the `url` property to a string, which returns `/settings/accounts`. 

All that aside, you are correct & using `prefix` is ok. It just wasn't clear as to why it was used. The only thing I would request is to remove the conditional statement since it's incorrect & it's not needed.

mleanos: Just for clarification, I checked, and the `.toString()` prototype method on `$state.$current.url` is
```javascript
UrlMatcher.prototype.toString = function () {
  return this.source;
};
```

mleanos:
LGTM.

Thank you @itelo! We'll be merging this soon, to get it in for the 0.5.0 release.

PR #1574, opened 2016-10-15
feat(user): add strict validations for username

### LINKED ISSUE #1204: strict validations for username?
A user can signup with username "mean mean & mean" , shouldnt there be strict rules for username?

sujeethk:
feat(user): add strict validations for username

Idea proposed by @sparshy #1204
Suggestions, rules and tests from Trustroots @simison
Added validations on user server model
Added relevant tests on user server tests

Fixes #1204

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -42,6 +42,9 @@ module.exports = {
   },
   logo: 'modules/core/client/img/brand/logo.png',
   favicon: 'modules/core/client/img/brand/favicon.ico',
+  illegalUsernames: ['meanjs', 'administrator', 'password', 'admin', 'user',
+                   'unknown', 'anonymous', 'null', 'undefined', 'api'
+                  ],
   uploads: {
     profileUpload: {
       dest: './modules/users/client/img/profile/uploads/', // Profile upload destination path
--- a/modules/users/client/controllers/authentication.client.controller.js
+++ b/modules/users/client/controllers/authentication.client.controller.js
@@ -15,6 +15,7 @@
     vm.signup = signup;
     vm.signin = signin;
     vm.callOauthProvider = callOauthProvider;
+    vm.usernameRegex = /^(?=[\w.-]+$)(?!.*[._-]{2})(?!\.)(?!.*\.$).{3,34}$/;
 
     // Get an eventual error defined in the URL query string:
     if ($location.search().err) {
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -74,7 +74,7 @@ module.exports = {
     options: {
       logResults: process.env.MONGO_SEED_LOG_RESULTS !== 'false',
       seedUser: {
-        username: process.env.MONGO_SEED_USER_USERNAME || 'user',
+        username: process.env.MONGO_SEED_USER_USERNAME || 'seeduser',
         provider: 'local',
         email: process.env.MONGO_SEED_USER_EMAIL || 'user@localhost.com',
         firstName: 'User',
@@ -83,7 +83,7 @@ module.exports = {
         roles: ['user']
       },
       seedAdmin: {
-        username: process.env.MONGO_SEED_ADMIN_USERNAME || 'admin',
+        username: process.env.MONGO_SEED_ADMIN_USERNAME || 'seedadmin',
         provider: 'local',
         email: process.env.MONGO_SEED_ADMIN_EMAIL || 'admin@localhost.com',
         firstName: 'Admin',
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -93,7 +93,7 @@ module.exports = {
     options: {
       logResults: process.env.MONGO_SEED_LOG_RESULTS !== 'false',
       seedUser: {
-        username: process.env.MONGO_SEED_USER_USERNAME || 'user',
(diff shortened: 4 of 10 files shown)


hyperreality:
Client side validation should be added too using ng-pattern and relevant ng-messages.

config/env/default.js
```diff
@@ -42,6 +42,11 @@ module.exports = {
   },
   logo: 'modules/core/client/img/brand/logo.png',
   favicon: 'modules/core/client/img/brand/favicon.ico',
+  illegalUsernames: ['meanjs', 'fwd', 'fwd:', 'reply', 'administrator', 'password', 'community',
+                   'unknown', 'anonymous', 'null', 'undefined', 'home', 'signup', 'signin', 'login',
+                   'edit', 'settings', 'demo', 'support', 'networks', 'profile', 'avatar', 'mini',
+                   'photo', 'account', 'api', 'modify', 'feedback', 'security', 'accounts', 'tribe', 'tag'
```
hyperreality: I recommend get rid of 'tribe' and add 'admin' to the list.

hyperreality: In fact I recommend getting rid of lots of these, such as avatar, demo, unknown etc... it should be up to the user to populate this list and MEAN should only reserve very important ones like admin out the box IMO.

sujeethk: @hyperreality `^[^.](?=.*[0-9a-z])[0-9a-z.\-_](?!.*(\.)\1).*[^.]{3,34}` can you please validate this? I am not really good with regex

sujeethk:
@hyperreality `^[^.](?=.*[0-9a-z])[0-9a-z.\-_](?!.*(\.)\1).*[^.]{3,34}` can you please validate this? I am not really good with regex

hyperreality:
Yeah, that regex is difficult to read and doesn't do the right thing at all. For instance, the `{3,34}` length rule is actually being applied only to `[^.]`, the positive lookahead `(?=.*[0-9a-z])` isn't covering the whole username, `[0-9a-z.\-_]` only covers a single character, and there are lots of `.*` which allow any character as many times as possible, which we don't want.

I think we should use this:

    ^(?=.{3,34}$)(?![_.])(?!.*[_.]{2})[a-zA-Z0-9._]+(?<![_.])$

Adapted from [here](http://stackoverflow.com/a/12019115)

sujeethk:
@hyperreality Lint is unhappy with that regex. It says the group is invalid on that regex.

Added a commit with existing regex, added client side validation with ng-pattern and relevant e2e tests. There is some issue with my protractor running e2e tests on my machine, so i couldnt run the tests. But client and server tests passed. Will wait for travis build to complete so i can check the e2e test results there.

sujeethk:
was able to fix my protractor issue. e2e tests passed.

mleanos:
@sujeethk Did you solve the issue with the linting errors on that regex? Also, what is the problem with the regex that @hyperreality said was incorrect? Was the actual issue with just combining it into one for the client-side check? Currently, it appears that the regex you're using here is the same as what @simison shared.

sujeethk:
@mleanos when I used the Regex that hyperreality provided above I got a lint error saying it was invalid regex. Then I resorted to using the same regex that Simison referred and used the same on the client side. All tests passed. So it's good for now.

modules/users/client/controllers/authentication.client.controller.js
```diff
@@ -15,6 +15,7 @@
     vm.signup = signup;
     vm.signin = signin;
     vm.callOauthProvider = callOauthProvider;
+    vm.usernameRegex = /^(?=.*[0-9a-z])[0-9a-z.\-_]{3,34}$/ && /^[^.](?!.*(\.)\1).*[^.]$/;
```
mleanos: I notice the client-side doesn't catch "user$lastname" as invalid, but the server-side does. This must be an issue with this client-side regex.

mleanos:
@simison Care to take a quick glance at this?

@meanjs/contributors

simison:
On holiday this week but i'll see if I can have a look at this tonight,
can't promise tho.

That forbidden words list is very TR specific, you should clean it out for
Meanjs.

sujeethk:
@mleanos @hyperreality When using hyperreality's suggestion above this is the error I get
`18:25  error  Parsing error: Error parsing regular expression: Invalid regular expression: /^(?=.{3,34}$)(?![_.])(?!.*[_.]{2})[a-zA-Z0-9._]+(?<![_.])$/: Invalid group`

Hence I used simison's suggestion. @mleanos you are right the same regex works on server side, but on the client side it allows special characters. I am trying to find why, is ngPattern using the regex differently or is it because i concatenated the two regex using `&&`. If any regex experts out here, your help is much appreciated!

sujeethk:
@mleanos Fixed the regex on client side using help from a regex IRC :). Used the same regex on the server side eliminating the need for a usernameRegex and dotRegex. Also added a couple of tests to check for minimum length and `log$in` special character

mleanos:
@sujeethk LGTM. I'll try to get this merged in the next couple of days. Thanks for taking this on.

mleanos:
@sujeethk Would you mind rebasing? I'll merge right after.

sujeethk:
@mleanos Rebased

mleanos:
Thank you @sujeethk!

PR #1576, opened 2016-10-16
feat(users): add option to import profile image from providers

itelo:


mleanos:
@itelo I don't this is the best approach to solve the problems that revolve around the profile images coming from the various providers.

We already have the profile image url info that comes from the provider itself when we authenticate using oauth. We should keep the logic that sets the User's `profileImageURL` inside the server-side passport strategies configurations.

Currently, we do have some inconsistencies with how we're gathering the profile image info from the implemented OAuth strategies. For instance, [Google's strategy](https://github.com/meanjs/mean/blob/master/modules/users/server/config/strategies/google.js#L31) is incorrect. The correct field coming from the Google provider data is `image.url` rather than `picture`; this line should be changed to the following (or something close to it):
```javascript
  profileImageURL: (providerData.image && providerData.image.url) ? providerData.image.url : undefined,
```

We just need to update these strategies to correctly set the user's profile image. My issue with the changes this PR introduces is that there's a lot of complicated logic inside the client-side controller that really should be handled server-side. Also, IMHO, this logic is unnecessary if we correctly set the profile image from the provider data when we first authenticate with each OAuth Provider.

Lastly, we're seeing an issue with reconciling the user's local provider profile images and their OAuth provider's images. This really stems from two things:
1) We're using a relative path for the local provider profile images
2) After removing the base-tag from the project we updated the profile image references with a prefix of "/" on the client-side.

To reconcile these issues we should correctly set the User's `profileImageURL` field to include the "/" when the user is updating their profile image through the upload feature server-side, and update the default value of the field in the model.

simison:
I second @mleanos on this one.

mleanos:
Closing since we haven't heard back from the contributor since this was originally opened.

PR #1577, opened 2016-10-16
Dynamic Loading of OAuth providers based On if they're in the config.

securityvoid:
## feat: users - Dynamic OAuth / Social Providers

This update adds a new REST call which pulls back a list of the available strategies which have settings defined within the env config (config\env). This REST end-point is then called from the user AuthenticationController in order to dynamically list strategies within authentication.client.view.html utilizing ng-repeat. The display attributes (Image, alt-text, oauth url) have default values based on the strategy name, but this can also be overridden in the environment configuration. 

With this pull request, and my pull request updating the router for Oauth calls; its now possible to add a new Oauth provider simply by adding a strategy.js, and adding the needed config items to the environment config. Everything else is dynamic. There is longer a need to update four places.

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -30,7 +30,13 @@ module.exports = {
   facebook: {
     clientID: process.env.FACEBOOK_ID || 'APP_ID',
     clientSecret: process.env.FACEBOOK_SECRET || 'APP_SECRET',
-    callbackURL: '/api/auth/facebook/callback'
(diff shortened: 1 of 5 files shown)


securityvoid:
I've been testing this a bunch more, and its buggy as is and needs more work. There are too many references to various clientID's etc. I'm going to close, and I'll resubmit when its ready.

PR #1581, opened 2016-10-18
fix(core): fix twitter meta tags

pgrodrigues:
fix(core): fix twitter meta tags

I believe the Twitter meta tags are outdated and this PR fixes them.
Taking into account the current MEAN.js tags and the Twitter docs, the most similar feature I found
is the Twitter Summary Card (https://dev.twitter.com/cards/types/summary), which requires 2 tags that currently do not exist (twitter:card and twitter:site) and there's another one already implemented in MEAN.js which is not referenced in Twitter docs (twitter:url).

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -33,6 +33,7 @@ module.exports = {
     callbackURL: '/api/auth/facebook/callback'
   },
   twitter: {
+    username: '@TWITTER_USERNAME',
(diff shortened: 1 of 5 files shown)


lirantal:
Thanks Pedro.

Could you by any chance also review the other social network logins? They might also need an update.

pgrodrigues:
I took a look at [Facebook docs](https://developers.facebook.com/docs/sharing/webmasters#markup) and those meta tags seem to be fine.

PR #1589, opened 2016-10-21
fix(users): Better MIME-type checking, remove image cropping library

hyperreality:
Follows on from work started here #1465 

Rationale behind this PR's changes [listed here](https://github.com/meanjs/mean/pull/1465#issuecomment-254016952). The new image upload page is more simple and automatically scales images >400px (height or width) down to 400px width to deal with large images as opposed to cropping them. Try it and let me know what you think.

And once again, I know that @lirantal [said that](https://github.com/meanjs/mean/pull/1465#issuecomment-243506527) magic-byte checking is out of scope for MEAN.js, but it still feels very wrong to ship a framework with arbitary file uploading capability out of the box. I'm more than willing to add to this PR the fix for this if I get the go-ahead.


mleanos:
@hyperreality Just a quick note.. I really like the image cropping, as it allows the end user to choose a file that might not properly display the way they want it; the user can "position" the image any way they like (for the most part). I don't think we should abandon the cropping feature altogether.

Have you seen https://github.com/CrackerakiUA/ui-cropper? It appears that the maintainers of ngImageCropFullExtended have moved it to a new repo. What, in particular, was the issue with this library when you first looked at it?

All the other changes in this PR look pretty good, but I'd like to give it a proper review.

hyperreality:
I like the image cropping too, but I just don't think that the ngImageCropFullExtended / ui-cropper library cuts it. I've had problems testing it on older smartphones and browsers, meaning that we can no longer say that MEAN works on all devices that Angular supports. As was noted before, the only really mature cropping libraries are Jquery based, for my own personal projects I'll be investigating those.

Furthermore, the file uploading feature is more modular without the cropping, users can more easily extend it to different filetypes, and come here if they want to see an example of how cropping is done. In summary, I don't feel like this particular cropping library can be justified as part of the core framework.

mleanos:
@hyperreality Ok. Thanks for explaining that. I'm ok with removing the image cropping for now. I'll try to test out the ui-cropper library, and determine if we can integrate it at a later time.

>Furthermore, the file uploading feature is more modular without the cropping, users can more easily extend it to different filetypes, and come here if they want to see an example of how cropping is done. In summary, I don't feel like this particular cropping library can be justified as part of the core framework.

I can agree with your last statement there, based on the fact that this library doesn't work well on mobile & older browsers. However, I don't think the image cropping feature affects the modularity of the file uploading feature. It's just a UI feature, that uses our back-end Multer implementation. Other features that deal with uploading other file types just wouldn't have that UI component.

config/env/default.js
```diff
@@ -43,8 +43,8 @@ module.exports = {
   logo: 'modules/core/client/img/brand/logo.png',
   favicon: 'modules/core/client/img/brand/favicon.ico',
   uploads: {
-    profileUpload: {
-      dest: './modules/users/client/img/profile/uploads/', // Profile upload destination path
+    profile: {
```
simison: Just a sidenote: should config changes be documented at changelog within the PR so that you ensure mentioning them when a new release goes out? (ping @lirantal)

hyperreality: @simison Good call, it's important that previous users notice the change.

modules/users/client/views/settings/change-profile-picture.client.view.html
```diff
@@ -1,21 +1,18 @@
 <section class="row">
   <div class="col-xs-offset-1 col-xs-10 col-md-offset-3 col-md-4">
-    <form class="signin form-horizontal">
+    <form name="vm.userForm" class="signin form-horizontal">
```
hyperreality: Good catch, removed it now.

mleanos:
@hyperreality LGTM. I restarted the Travis CI build, so we'll see if it passes.

@lirantal Assuming the build passes, I think this is ready. Do you want to give it a quick once over? There was some discussion on magic-byte that I wasn't a part of. Does that still need addressing?

PR #1592, opened 2016-10-23
feat(build): Ignore compiled sass/less files from git

shanavas786:
Compiling `sass/less` files to `<module>/client/css` directory makes it difficult to differentiate between compiled and raw css files.

By compiling them to appropriate directories, compiled files can be easily ignored.

### DIFF
--- a/config/assets/default.js
+++ b/config/assets/default.js
@@ -31,7 +31,7 @@ module.exports = {
       tests: ['public/lib/angular-mocks/angular-mocks.js']
     },
     css: [
-      'modules/*/client/css/*.css'
+      'modules/*/client/{css,less,scss}/*.css'
     ],
     less: [
       'modules/*/client/less/*.less'
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -169,9 +169,6 @@ gulp.task('sass', function () {
   return gulp.src(defaultAssets.client.sass)
     .pipe(plugins.sass())
     .pipe(plugins.autoprefixer())
-    .pipe(plugins.rename(function (file) {
-      file.dirname = file.dirname.replace(path.sep + 'scss', path.sep + 'css');
-    }))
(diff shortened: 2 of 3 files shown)


config/assets/default.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
       tests: ['public/lib/angular-mocks/angular-mocks.js']
     },
     css: [
-      'modules/*/client/css/*.css'
+      'modules/*/client/{css,less,scss}/*.css'
```
mleanos: How does this affect the `less` & `sass` settings below?

gulpfile.js
```diff
@@ -169,9 +169,6 @@ gulp.task('sass', function () {
   return gulp.src(defaultAssets.client.sass)
     .pipe(plugins.sass())
     .pipe(plugins.autoprefixer())
-    .pipe(plugins.rename(function (file) {
```
mleanos: Why was this removed?

shanavas786:
@mleanos Now, both `less and sass` files are compiled to `module/client/css/` directory [[1]](https://github.com/meanjs/mean/blob/master/gulpfile.js#L173), [[2]](https://github.com/meanjs/mean/blob/master/gulpfile.js#L184), 
 and only css files in `module/client/css/` directory is used as css assets [[3]](https://github.com/meanjs/mean/blob/master/config/assets/default.js#L34).

My proposal is to compile `sass/less` files to `module/client/(sass/less)` directory itself  instead of  `css` directory **(that is why directory rename is removed)** and  include css files from those directories also in `css` assets **(add scss/less directories to css asset path)**.

mleanos:
@shanavas786 Thank you. That perfectly answered my questions. Also, I agree with the goals of these changes. It's definitely gonna be easier to manage these compiled assets.

@meanjs/contributors comments?

Issue #1608, opened 2016-11-02
Could we manage our admin and user routes in another way ?


leodonny:
Hi guys,

I'm currently working on a project which requires to create user profile pages. I took a look on routing system on server side about users and I noted something annoying.
Today, /api/users/:userId route returns one specific user from an admin request.
Getting a user for another one is currently "impossible" : according to the fact the admin route /api/users/:userId is, as its name says, an admin route, it can't be used (it shouldn't be) by a random user.

This could be managed on server side (whether I'm admin or user) but my logic brain thinks it could be interesting to have two different routes which will also allow to split “by naming” the admin routes.

I propose to keep /api/users/* for a commun utilization (update/connection/delete and get a user with restricted informations about him) and add “admin” before actual admin routes : /api/admin/users/:userId. The user returned by an admin request could have more informations than a user returned after a user request.

In the future, I will need to add more admin routes so I have to organize my actual routes and I think the idea could interest you. But the main purpose why I'm posting this here is that you are certainly better than me and you already thought (maybe answered) about it and there is a good reason that I don't get.

So, guys, what do you think ? What should I do ?
Thanks.

mleanos:
This PR has a very long (and grueling at times) discussion on this matter:
https://github.com/meanjs/mean/pull/713

I agree with [this comment](https://github.com/meanjs/mean/pull/713#issuecomment-125081378). So I would prefer to have something pretty close to what is described here: https://github.com/meanjs/mean/pull/713#issuecomment-231674007

We should be using the same normalized RESTful routes for all types of entities in our database schema. The ACL policies should then handle permissions. 

For cases when you need to return additional fields for Admin users then you can handle that logic in your server-side controller route handlers. IIRC, the mongoose populate method accepts an array of field names. You can merely add your additional fields to a base set of field names, or just use different "whitelisted" arrays of field names for user's and non-Admin user's.

PR #1641, opened 2016-11-16
feat(users) Separate admin routes and cleanup

Wuntenn:
- removed unused users.authorization.server.controller
- moved admin routes to admin path


mleanos:
@Wuntenn The idea of adding *admin* to these types of routes was discussed at some length here:
https://github.com/meanjs/mean/pull/713#issuecomment-125023988

I'll spend some time reviewing that thread again, and come back with my thoughts. However, I'm leaning toward not wanting `/api/admin/***` type routes for the api. Can you offer up some reasoning, or ideas on why it would be better to use that pattern?

Wuntenn:
I want to add a singleton resource to users which works via a route like: `/api/users/myresource`, however I had issues with `myresource` being interpreted as in incorrect mongoose Id type.

_Now Looking at that thread, maybe it's an ordering issue, however I doubt I would have discovered this because I followed the convention in place and gave `myresource` it's own route file._

After digging through the routes I noticed (again) that the userById was redundant and that only admin used anything like it.

The `req.user` has all the details of the user already so I couldn't see the need for routes like:

`api/users/:userId/myresource`;and
`api/users/:userId/myresource/:myresourceId`

As we don't really need to look up the user. It a waste of a param. Only the admin will ever look up a user.

I liked the idea of being able to user the ACL with all user user routes and the idea of keeping the apis separate as a security precaution. Admin seemed harmless and like the direction being taken on the front-end.

I'm flexible though, maybe I'll have to go with the above route style...

simison:
>As we don't really need to look up the user. It a waste of a param. Only the admin will ever look up a user.

This boilerplate doesn't, but an app using this boilerplate might. ;-)

Sorry but adding`/admin` doesn't seem like a typical RESTful pattern to me... having both `admin` routes and ACL rules is kinda doing role management twice.

This also restricts to one role only ("admin"). Some app might need to have "moderators" or other admin-like roles.

While security precaution is a valid point, I recon it's a matter of having proper testing, really.

So I'd vote not merging this.

simison:
Oh and totally sorry if I'm repeating points already discussed elsewhere, no time to even skim through that discussion. ;-) Cheers!

mleanos:
@Wuntenn I can't recall the depth of the discussion, in that thread I referenced, so I may be repeating as well :)

I think the crux of the issue here is the non-standard User routes. They aren't using the `:userId` parameter, but IMO they should be. Furthermore, we should be enforcing ACL on **all** the User routes.

It may seem redundant to include `:userId` in these routes when more often than not, the requesting User is the same as the entity requested. However, we find ourselves in this situation where we're having to hack through these limitations. The more I use MEANJS for starter/prototype projects, the more I'm realizing the need to standardize these routes; I'm finding myself working around this issue far too often.

Daron, would you be interested in submitting a new PR to "fix" these non-standard User routes?

See https://github.com/meanjs/mean/issues/1608#issuecomment-258063762, for more or less my current view on the matter.

PR #1647, opened 2016-11-21
Incorrect production assets

mleanos:
Removed the ng-img-crop references from the production assets
configuration, since the library has been removed from the project.

Also, re-ordered the CSS section to be in alphabetical order.

### DIFF
--- a/config/assets/production.js
+++ b/config/assets/production.js
@@ -7,10 +7,9 @@ module.exports = {
     lib: {
       css: [
         // bower:css
+        'public/lib/angular-ui-notification/dist/angular-ui-notification.min.css',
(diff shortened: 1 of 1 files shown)


mleanos:
Every time I run `gulp build` this file would get updated since the `wire-dep:prod` task gets ran as well. This kept adding an unnecessary diff in my commits, since this file should have already been updated with these fixes.

I'll merge this in after the build is done, since it's just simple & obvious fix.

mleanos:
I apologize. In my haste, I forgot to adhere to the commit guidelines. Thus, the commit header is incorrect.

It should have been `fix(assets): Incorrect production assets`

Issue #1652, opened 2016-11-24
`npm install` error for Node v6.3+ due to "node-inspector" dev dependency in package.json


kimbaudi:
After cloning this repo, I executed `npm install` and got error. I am using node v7.2.0 and npm v4.0.2. I believe `npm install` error affects people using Node v6.3 or greater.

Here is the relevant snippet of `npm-debug.log`:
```
123423 error Darwin 16.1.0
123424 error argv "/Users/kimbaudi/.nvm/versions/node/v7.2.0/bin/node" "/Users/kimbaudi/.nvm/versions/node/v7.2.0/bin/npm" "install"
123425 error node v7.2.0
123426 error npm  v4.0.2
123427 error code ELIFECYCLE
123428 error v8-debug@0.7.7 install: `node-pre-gyp install --fallback-to-build`
123428 error Exit status 1
123429 error Failed at the v8-debug@0.7.7 install script 'node-pre-gyp install --fallback-to-build'.
123429 error Make sure you have the latest version of node.js and npm installed.
123429 error If you do, this is most likely a problem with the v8-debug package,
123429 error not with npm itself.
123429 error Tell the author that this fails on your system:
123429 error     node-pre-gyp install --fallback-to-build
123429 error You can get information on how to open an issue for this project with:
123429 error     npm bugs v8-debug
123429 error Or if that isn't available, you can get their info via:
123429 error     npm owner ls v8-debug
123429 error There is likely additional logging output above.
123430 verbose exit [ 1, true ]
```

The issue is "node-inspector" dev dependency in `package.json` (https://github.com/node-inspector/v8-debug/issues/7).

I resolved this issue by simply removing `"node-inspector"` dev dependency from `package.json` and re-running `npm install`.

codydaig:
Howdy! We don't currently support node v7. 

For future, you can always checkout our Travis.CI build page for what versions of node are currently working. Anything under allowed failures is versions we plan to support but are currently breaking the build. https://travis-ci.org/meanjs/mean

PR #1655, opened 2016-11-26
fix(core): update engines from package.json

simison:
Since tests are failing for Node v7, clarify information about version support.
- Update [`engines`](https://docs.npmjs.com/files/package.json#engines) information from `package.json`
- Mention Node.js and MongoDB version requirements at README.md
- Mention that generator installs outdated mean.js version

You can use https://jubianchi.github.io/semver-check/ to validate semver format. 

Related to #1652

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ Before you begin we recommend you read about the basic building blocks that asse
 ## Prerequisites
 Make sure you have installed all of the following prerequisites on your development machine:
 * Git - [Download & Install Git](https://git-scm.com/downloads). OSX and Linux machines typically have this already installed.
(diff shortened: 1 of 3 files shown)


codydaig:
What will this do when someone tried to run this with Node 7?

simison:
@codydaig The field is advisory only and currently `npm` doesn't yeld warnings with it (it used to); there might be some other tools that do, though. Most importantly it gives the info to users in a standard place. 

I could add the info to README.md as well, WDYT?

codydaig:
@simison What would be really cool would be to add a check in the /config/lib/app.js for the right node version and if it's not supported, output an error message. That's a common issue that seems to come up often. 

I'm also all for adding it into the README. I think that will get checked more than the package.json file by people new to the node world.

simison:
@codydaig installing requirements would probably fail first on wrong Node.js version anyway, so a check+warning at app.js feels a bit redundant.

I'll add this to readme tho!

simison:
@codydaig updated the PR, WDYT?

simison:
Closing as outdated.

PR #1657, opened 2016-11-29
feat(server): Add warning if using an unsupported version of Node

hyperreality:
As suggested by @codydaig at #1655. A lot of the issues here recently have been raised by people simply running the wrong versions of Node, and while the majority are errors during the `npm install` step, sometimes the user will inadvertently manage to run MEAN with the wrong Node version and encounter weird errors... hopefully this will prevent that.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -82,7 +82,7 @@ var validateNodeJsVersion = function (config) {
  * Validate config.domain is set
  */
 var validateDomainIsSet = function (config) {
-  if (!config.domain) {
+  if (!config.domain && process.env.NODE_ENV === 'production') {
     console.log(chalk.red('+ Important warning: config.domain is empty. It should be set to the fully qualified domain of the app.'));
   }
 };
--- a/config/config.js
+++ b/config/config.js
@@ -7,7 +7,8 @@ var _ = require('lodash'),
   chalk = require('chalk'),
   glob = require('glob'),
   fs = require('fs'),
-  path = require('path');
+  path = require('path'),
+  semver = require('semver');
 
 /**
  * Get files by glob patterns
@@ -68,7 +69,17 @@ var validateEnvironmentVariable = function () {
   console.log(chalk.white(''));
 };
 
-/** Validate config.domain is set
+/**
+ * Validate user is running supported Node.js version
+ */
+var validateNodeJsVersion = function (config) {
+  if (!semver.satisfies(process.version, config.meanjs.engines.node)) {
+    console.log(chalk.red('+ Important warning: your version of Node (' + process.version.substr(1) + ') is unsupported by MEAN.js. Please use a version that satisfies the following requirements: ' + config.meanjs.engines.node + ', as specified in the package.json file.'));
+  }
+};
+
+/**
+ * Validate config.domain is set
  */
 var validateDomainIsSet = function (config) {
   if (!config.domain) {
@@ -172,6 +183,7 @@ var initGlobalConfigFiles = function (config, assets) {
  * Initialize global configuration
  */
 var initGlobalConfig = function () {
+
   // Validate NODE_ENV existence
   validateEnvironmentVariable();
 
@@ -211,6 +223,9 @@ var initGlobalConfig = function () {
 
   // Validate session secret
   validateSessionSecret(config);
+
+  // Print a warning if using unsupported version of Node.js
+  validateNodeJsVersion(config);
 
   // Print a warning if config.domain is not set
   validateDomainIsSet(config);
--- a/config/lib/app.js
+++ b/config/lib/app.js
@@ -47,6 +47,7 @@ module.exports.start = function start(callback) {
       console.log(chalk.green('App version:     ' + config.meanjs.version));
       if (config.meanjs['meanjs-version'])
         console.log(chalk.green('MEAN.JS version: ' + config.meanjs['meanjs-version']));
+      console.log(chalk.green('Node version:    ' + process.version.substr(1)));
       console.log('--');
 
       if (callback) callback(app, db, config);


hyperreality:
I've also just taken the liberty to add an extra condition that disables the warning about an empty config.domain, unless in production environment mode. It is offputting for a new user to be presented with an "important warning" like that when they are running the app for the first few times on their local machines.

config/lib/app.js
```diff
@@ -47,6 +47,7 @@ module.exports.start = function start(callback) {
       console.log(chalk.green('App version:     ' + config.meanjs.version));
       if (config.meanjs['meanjs-version'])
         console.log(chalk.green('MEAN.JS version: ' + config.meanjs['meanjs-version']));
+      console.log(chalk.green('Node version:    ' + process.version.substr(1)));
```
simison: @hyperreality please ping me if this PR goes in and I'll remove [similar line](https://github.com/meanjs/mean/pull/1655/commits/875f2c104501f9c716fe76f3cad644d140e6f226#diff-e4a0ef5ca61cb235bb9ed42cac9853ffR47) from my PR.

simison:
Hmm, are there _really_ a lot of these cases where `npm install` on unsupported node version passes but the app doesn't run? I'm a bit sceptical. :-)

simison:
BTW a fine implementation of the check otherwise!

config/config.js
```diff
@@ -68,10 +69,20 @@ var validateEnvironmentVariable = function () {
...
+    console.log(chalk.red('+ Important warning: your version of Node (' + process.version.substr(1) + ') is unsupported by MEAN.js. Please use a version that satisfies the following requirements: ' + config.meanjs.engines.node + ', as specified in the package.json file.'));
+  }
+};
+
+/**
+ * Validate config.domain is set
  */
 var validateDomainIsSet = function (config) {
-  if (!config.domain) {
+  if (!config.domain && process.env.NODE_ENV === 'production') {
```
simison: A separate PR for this would be more suitable?

mleanos:
>Hmm, are there really a lot of these cases where npm install on unsupported node version passes but the app doesn't run? I'm a bit sceptical. :-)

I'm wondering this as well. I am feeling the pain a little bit with all the support issues coming in that are related to the Node version; *not to mention the issues revolving around the generator* :(

I do like the implementation of the check though. So I'm not opposed to this going in, even if it is redundant. I'm more concerned with what the real issues are with running Node 7. Are they easily fixed, or are they more-or-less issues with the packages we're using?

config/lib/app.js
mleanos: I do prefer the location of this line in Mikael's PR. Seems more apropos for it to be grouped with the app, server, and db settings display.

mleanos:
@codydaig I assigned this to you since it's very related to #1655 & has some overlap changes.

config/lib/app.js
hyperreality: @simison will do! Good point @mleanos

config/config.js
hyperreality: My rationale behind the change was the following:
- Seeing many people install new software, getting a red error message about configuration post-install is unexpected - the first thought is that something went wrong during the installation process, probably with Node.
- By the time users actually deploy, I would expect them to have run the app several times in production mode in a safe environment to check it: providing multiple opportunities to see the message.

Anyway, I can see where you're coming from @mleanos. I think what we really need is just a better and more descriptive message. What do you mean by _scary_ implications?. I believe you might be best placed to write the message ;). I'll remove the change for now.

hyperreality:
It's not just the `npm install` situation that wrong Node versions affect. A scenario you might find yourself in is accidentally running MEAN with a different version of Node than you installed it with, if you used nvm to switch Node versions for another project. This can be a source of really unusual bugs until you suddenly realize you're on an older version. Realistically, if your version is really old, a module will complain cryptically first... if only there were some way to check the Node version _before_ anything else...

hyperreality:
It turns out that we could probably eliminate this whole class of issues more efficiently, with a     "preinstall": "node scripts/checkPrerequisites.js" line in package.json, which would automatically run before `npm install`. Just testing the water, how does that sound?

hyperreality:
@mleanos I just tried MEAN with Node 7.2.0 and the only problem I had was with Node-sass, commenting that out of the gulpfile and everything ran smoothly. Obviously that's _after_ installing everything with a lower version of Node, https://github.com/meanjs/mean/issues/1652 had problems with `npm install`.

config/config.js
mleanos: I see what you're saying @hyperreality. However, maybe we could just "introduce" them to the idea of the domain setting needing to be set; can we show this message as a warning in non-Production environments, and display as a critical (red) error when in production?

I don't want to over-complicate that matter, but I do think we can address the underlying issue of user's getting confused. My statement of "scary" was more that after some time spent on development, with everything working as expected, and to all of a sudden see a message like that in a production environment (whether local or deployed) could give someone a moment of panic. Either way, we can pick this up in a separate PR.

mleanos:
@hyperreality I like the idea of the `preinstall` script. Let's go down that route.

hyperreality:
I just wrote the preinstall script which worked nicely on Node < 4.6.0 but strangely it didn't seem to run on Node 7. Turns out NPM 3 has a problem, actually running preinstall post-install: https://github.com/npm/npm/issues/10379, which has been around for a year now.

So maybe this truly is impossible to achieve.

simison:
How about this?
```
scripts: {
  "check-prerequisites": "node scripts/checkPrerequisites.js",
  "install": "npm run check-prerequisites && npm install"
}
```

hyperreality:
It turns out that in Npm 3 the preinstall script does run before node_modules is installed, but only after http requests have been made. So it could still be useful.

However, because of course modules have not been installed at that stage the checkPrequisites file would have to bundle a fair amount of the Semver library code, what do you think about this @simison @mleanos?

simison:
@lirantal @mleanos, do we want this check in? If so i could write some notes about code changes

lirantal:
do be honest I don't think I like this in.
it should be clear by the README and package.json's engines to which node.js versions are supported. adding this boilerplate code seems overhead to me.

simison:
+1 @lirantal

On 15 Jun 2017 19:19, "Liran Tal" <notifications@github.com> wrote:

do be honest I don't think I like this in.
it should be clear by the README and package.json's engines to which
node.js versions are supported. adding this boilerplate code seems overhead
to me.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/meanjs/mean/pull/1657#issuecomment-308792421>, or mute
the thread

mleanos:
> it should be clear by the README and package.json's engines to which node.js versions are supported.

Agreed.

PR #1659, opened 2016-11-30
feat(articles): Simple test enhancement

### LINKED ISSUE #1658: Simple Test Enhancement
I'd like to propose a simple enhancement to the test cases for the example Article module.  The enhancement would cover cases where test cases fail without sharing useable debugging data when Articles can't be saved (which can happen for a variety of reasons).  

The change would happen on line 309, and contain a simple conditional checking for an error and printing it out if one exists.  I'm submitting a pull request shortly with this enhancement.  Other reference for this enhancement exists here: https://github.com/meanjs/generator-meanjs/pull/257

jrodenbostel:
feat(articles): Simple test enhancement

Simple test enhancement proposed by @jrodenbostel.  Eases troubleshooting when Article save fails during test execution.

Further reference here: https://github.com/meanjs/generator-meanjs/pull/257

Fixes #1658

### DIFF
--- a/modules/articles/tests/server/article.server.routes.tests.js
+++ b/modules/articles/tests/server/article.server.routes.tests.js
@@ -308,7 +308,7 @@ describe('Article CRUD tests', function () {
     // Save the article
     articleObj.save(function (err) {
       if (err) {
-        done(err);
+        return done(err);
       }
       request(app).get('/api/articles/' + articleObj._id)
         .end(function (req, res) {
--- a/modules/articles/tests/server/article.server.routes.tests.js
+++ b/modules/articles/tests/server/article.server.routes.tests.js
@@ -306,7 +306,10 @@ describe('Article CRUD tests', function () {
     var articleObj = new Article(article);
 
     // Save the article
-    articleObj.save(function () {
+    articleObj.save(function (err) {
+      if (err) {
+        done(err);
+      }
       request(app).get('/api/articles/' + articleObj._id)
         .end(function (req, res) {
           // Set assertion


jloveland:
looks good

modules/articles/tests/server/article.server.routes.tests.js
```diff
@@ -306,7 +306,10 @@ describe('Article CRUD tests', function () {
     var articleObj = new Article(article);
 
     // Save the article
-    articleObj.save(function () {
+    articleObj.save(function (err) {
+      if (err) {
```
mleanos: @jrodenbostel If the intention is to catch/handle unexpected (possibly random) errors here, then why not just fail the test when `err` exists? This test should only pass when all conditions and expectations are met, right?

Maybe I'm thinking about it wrong. I guess we have two options, either we assert that `err` doesn't exist to fail the test when it does exist, or we pass the err to the callback to propagate the actual error back to the test runner; perhaps, the former solves both?

jrodenbostel: My intent was only to make diagnosing test failures easier when the failure is related to the article not being saved.  When that save fails, the definitely fails, just not in a clear way.  

Nothing crazy, but dealing with that was a enough of a hassle at the time to motivate me to submit this patch.

modules/articles/tests/server/article.server.routes.tests.js
```diff
@@ -306,7 +306,10 @@ describe('Article CRUD tests', function () {
...
+        done(err);
```
simison: This should really be `return done(err);` instead.

lirantal:
@meanjs/contributors  can you guys keep an eye on this PR. if it doesn't get addressed in a week or so let's just re-PR it and merge so we don't have many stale PRs.

jrodenbostel:
@codydaig I just pushed the requested changes.

simison:
LGTM

codydaig:
LGTM

@mleanos Look good to you?

lirantal:
👍

mleanos:
LGTM

PR #1686, opened 2016-12-22
feat(core): Enhancement page title directive

PierreBrisorgueil:
Proposal to automate the generation of PageTitle based on state

We split state, for example article.view -> Article - View
And the pageTitle in route configuration just add the last part -> Article - View - "title of article"

It's more easy in projects because with this, it is defined on all page. Without any configuration. 

And for me it's not just a pageTitle for the Head, but i display this on the web page (http://romeanet.weareopensource.me), in order to explain to the end user where he is in the website (Like Gitlab, just a path "Home / ... / ...."). 

This require you correctly name your state :p 

Edit : i added lodash in front .... no opinion on this, but why not ?


hyperreality:
It's certainly a good idea. I do wonder whether some users might want different names for states and titles though.

> i added lodash in front .... no opinion on this, but why not ?

I think the question is why? You can easily do those things without it, and save downloading an extra unnecessary library.

PierreBrisorgueil:
Actually it's only the title hide in the head ... so no reasons for me. 
If they want to use this in the page content, they can edit the directive :p, and add an attribute eventually. 

But I think in this way we cover the general case :)

73 KB for lot of performance gains, and lot of functions easy to use in one line. 

You don't need it, in the best of worlds ;)
The work is totally done on server side in case of simple CRUD models ... but when you project do other things, have some graph, there is no reasons to put all work on the server side. 

Sometimes you get a set of data, and need to clean it a little to display a graph, or you use three kind of graph who need different formats of data. 
In this kind of case, if your dataset is not to big, you want to keep a RestFull API simple, so you have one road to get the data, and change the format on the client side. 

It would be a shame to do several routes just for a different data format, if you have a big project ...

it's just an idea for example (not the great, i'm tired :p), but there is no rule !
Do everything on server side, do everything on client side, it's common sense and a balance :)

So for me the question is not why, but it is interesting for a better starter kit ? 

if not i can remove it in the PR quickly ;)

mleanos:
@PierreBrisorgueil @hyperreality Just a quick comment on the use of Lodash on the front-end. I actually do like the functionality, and utility of having Lodash in all area's of my application; I definitely see a huge benefit of having it in my client-side code.

However, I prefer to use [ng-lodash](https://github.com/rockabox/ng-lodash). It's better to keep the front-end implementation consistent with Angular's way of doing things, and adhering to our styleguide's Dependency Injection.

modules/core/client/directives/page-title.client.directive.js
```diff
@@ -19,13 +19,17 @@
 
       function listener(event, toState) {
         var applicationCoreTitle = 'MEAN.js',
-          separeteBy = ' - ';
+          separeteBy = ' - ',
```
mleanos: pre-existing typo... should be "separate". This should be fixed in these changes.

bower.json
```diff
@@ -14,7 +14,8 @@
     "angular-ui-router": "~0.2.18",
     "bootstrap": "~3.3.6",
     "ng-file-upload": "~12.1.0",
-    "owasp-password-strength-test": "~1.3.0"
+    "owasp-password-strength-test": "~1.3.0",
+    "lodash": "^4.6.1"
```
mleanos: Why not use [ng-lodash](https://github.com/rockabox/ng-lodash)? See my comment in the main thread of this PR.

modules/core/client/directives/page-title.client.directive.js
PierreBrisorgueil: sorry for my english, juste rename separateBy by separate ? 

i just kept the name of the original file

modules/core/client/directives/page-title.client.directive.js
```diff
@@ -19,13 +19,17 @@
...
+          stateTitle = applicationCoreTitle + separeteBy;
+
+        _.forEach(toState.name.split('.'), function(value, index) {
+          stateTitle = stateTitle + _.capitalize(value) + separeteBy;
```
simison: Including lodash into the project just for this is quite a lot of extra kbs...

PierreBrisorgueil: clearly, i suggest the addition. it's an example in order to discuss about this subject.

modules/core/client/directives/page-title.client.directive.js
simison: Sorry to bikeshed here: `separator` ?

PierreBrisorgueil:
> Both the current way, and this proposal, ends up defining the pageTitle in a static way. One is based on the configuration setting, and one is based on the route name. I actually think the former is more flexible, and doesn't tightly couple the page-title to the state name. The route config only gets set once per route, and it feels more natural to have the page-title explicitly set there.

As you want, for me it's just the <tilte> but we have an interest in design to use this in order to generate a path with link in the content, like in forums. I 

In this version you can add your specifications in the end. If your states are clear, there is no problem for this, they generate the begin of your title. I understand your point of view, but in the case of the stack starter, my way have the same result for every state. 

I have a very strict vision to try to standardize as much as possible.

bower.json
PierreBrisorgueil: no reason, just never try it :) i have just push this in order to launch the discussion about loadash :)

PierreBrisorgueil:
Quick clean of lodash, but delete capitalize, nothing native for this :) 

but i think we need to keep this debate, sorry for this, it's probably not the topic for, my bad idea !

modules/core/client/directives/page-title.client.directive.js
mleanos: `separator` is probably more concise.

@PierreBrisorgueil I actually meant `separateBy`, sorry I wasn't clear on that. However, I do prefer Mikael's suggestion.

hyperreality:
@PierreBrisorgueil from my point of view, this is an interesting PR, one that many users might feel is beneficial, but is probably unsuitable for the framework, as it is too 'custom'

To capitalise you can do `value.charAt(0).toUpperCase() + value.slice(1)`, bear in mind that anything lodash or other libraries do would merely be a wrapper around this anyway.

modules/core/client/directives/page-title.client.directive.js
PierreBrisorgueil: done

PierreBrisorgueil:
@hyperreality okok i see, in my opinion it's the opposite, create a var pageTitle for every page is more custom and work not required for a starter. 

> To capitalise you can do value.charAt(0).toUpperCase() + value.slice(1), bear in mind that anything lodash or other libraries do would merely be a wrapper around this anyway.

Of course, but there is so many way to do this. And this is something that is required in several places, the difference is that in a library, you can potentially have the better way, people work on this for you. Small code, more readable. 

for example, capitalize in lodash : https://github.com/lodash/lodash/blob/4.17.2/lodash.js#L14120

Whatever, i include your solution ;) 

Just another question, it makes sense to use this directive in different view no ? in order to replace h1 content : https://github.com/meanjs/mean/blob/master/modules/articles/client/views/list-articles.client.view.html No matter what we do, pageTitle based on states or not. The title of the head, should it not be common to what is displayed to the user and used in views ?

modules/core/client/directives/page-title.client.directive.js
```diff
@@ -1,4 +1,4 @@
-(function () {
+(function() {
```
simison: This should be left unchanged.

modules/core/client/directives/page-title.client.directive.js
```diff
@@ -19,13 +19,17 @@
...
+        });
         if (toState.data && toState.data.pageTitle) {
-          var stateTitle = $interpolate(toState.data.pageTitle)($state.$current.locals.globals);
-          element.html(applicationCoreTitle + separeteBy + stateTitle);
-        } else {
-          element.html(applicationCoreTitle);
+          stateTitle = $interpolate(stateTitle + toState.data.pageTitle + separator)(($state.$current.locals.globals));
         }
+        stateTitle = stateTitle.slice(0, 0 - separator.length);
+        element.html(stateTitle);
```
simison: Would just `element.text()` be enough here? I feel `.html()` being a bit risky for security.

modules/core/client/config/core.client.routes.js
```diff
@@ -44,8 +44,7 @@
           }
         },
         data: {
-          ignoreState: true,
-          pageTitle: 'Not Found'
```
simison: What ends up being the title of specific user/article pages if those aren't found?

PierreBrisorgueil: nop, just the state : "MEAN.js - Bad-request" "MEAN.js - Not-found"

modules/core/client/directives/page-title.client.directive.js
PierreBrisorgueil: Agree with you, it's the old code, don't want to change it, but agree. For me, it's important only if you display pageTitle in your content page, in order to add links to every parts. 
Done.

modules/core/client/directives/page-title.client.directive.js
PierreBrisorgueil: My JsBeautify, i will have a look on it config. 
Done.

modules/chat/client/config/chat.client.routes.js
```diff
@@ -15,8 +15,7 @@
         controller: 'ChatController',
         controllerAs: 'vm',
         data: {
-          roles: ['user', 'admin'],
-          pageTitle: 'Chat'
```
mleanos: I think `pageTitle` should be added back here, no?

mleanos: After looking back at the directive changes, I understand what you're doing here a bit better. I misunderstood the inclusion of the `pageTitle` option in the other routes for Articles, as an "override". But here you're still using the State name to build the resulting title.

modules/core/client/config/core.client.routes.js
mleanos: I'm not a fan of how those look.

modules/core/client/directives/page-title.client.directive.js
```diff
@@ -19,13 +19,17 @@
...
+
+        toState.name.split('.').forEach(function(value, index) {
+          stateTitle = stateTitle + value.charAt(0).toUpperCase() + value.slice(1) + separator;
+        });
         if (toState.data && toState.data.pageTitle) {
-          var stateTitle = $interpolate(toState.data.pageTitle)($state.$current.locals.globals);
-          element.html(applicationCoreTitle + separeteBy + stateTitle);
-        } else {
-          element.html(applicationCoreTitle);
+          stateTitle = $interpolate(stateTitle + toState.data.pageTitle + separator)(($state.$current.locals.globals));
```
mleanos: I misunderstood these latest commits, to be that you were using `pageTitle` as an override to using the State name as the resulting Page Title. I sort of liked that idea. Does this sound like it would be better than using a combination of State name, and `pageTitle`? For instance, a user could explicitly specify a Title using the `pageTitle` option in the route config to bypass the use of the State name. Otherwise, if the option isn't specified then the State name would be used.

I don't see much benefit in using both at the same time.

modules/core/client/config/core.client.routes.js
PierreBrisorgueil: Agree, i'm not fan of state with '-' ;)
but to manage this, we can replace - by space for title.

modules/core/client/directives/page-title.client.directive.js
PierreBrisorgueil: Actually i use a combination. For an article, in order to get the **articleResolve.title** and interpolate it to display it. This part just generate the end of the title from the resolve precise by a developer. 

foreach will generate : "MEAN.js - Article - View"
pageTitle will generate : "My first article"
The combination : "MEAN.js - Article - View - My first article"

_I don't think we have this in "toState", it's what i need the pageTitle var. I tried with the the resolve part of the var toState but without success._ 

I see your point, I had thought about it. But if we don't find another way to get the articleResolve.title, without the pageTitle option in the route config ( {{articleResolve.title}} ). This will imply that we will have two options in the route config. 

One to get the end of the automatic pageTitle -> the title of the resource
One to bypass the automatic Title if it's define -> to bypass the use of the State name.

I abandoned this idea to lighten the config file. 

my only personal edit on this PR, in my project is to add 

```
      scope: {
        kind: '=kind'
      },
```

In order to use the directive in my <head>
but also in my content page with kind="header", i will generate and html title with link, and display it in my page. 

Like this : 

screen shot 2016-12-24 at 10 32 09

bower.json
simison: Wait, why is this added in this PR again?

I don't see it being used anywhere.

mleanos:
@PierreBrisorgueil What's the status of this PR?

See Mikael's last comment on `lodash` being added to the Bower dependencies. It isn't used anywhere.

PierreBrisorgueil:
hello @mleanos, sorry for the delay, clean and rebase :)

mleanos:
LGTM. Thanks @PierreBrisorgueil 

I'll merge in the next couple of days, if there's no more feedback.

mleanos:
@simison @lirantal Any more comments here?

PR #1687, opened 2016-12-23
[v2] MEAN.JS API - Introducing the SequelizeJS library

lirantal:
* Introducing the SequelizeJS ORM for traditional SQL databases support
* Adding the MySQL library binding for SequelizeJS
* Removing of `node-inspector` library which is not compatible with 
* Refactoring of Mongoose,  `app.js`, models structure, and config files 

The new server instantiation is based on async/await and requires Node v7 (Babel isn't yet introduced in the codebase), so run with:

```
node --harmony-async-await server.js
```

Code Review: https://github.com/meanjs/mean/commit/7a5ffa3ae456abdf2c2f5eeba236453e4a867e2f


server/config/lib/app.js
```diff
@@ -6,32 +6,54 @@
...
+}
 
-module.exports.start = function start(callback) {
-  var _this = this;
 
-  _this.init(function (app, db, config) {
+// Boot up the server
+module.exports.start = function start() {
+  return new Promise(async function (resolve, reject) {
+    const orm = await startSequelize();
```
mleanos: @lirantal Why start both Sequelize & Mongoose by default? I understand there could be cases where it's useful to have both running side-by-side. However, it seems a little unclear as to why it's done like this by default, since the configuration option is also set by default.

server/config/lib/app.js
```diff
@@ -6,32 +6,54 @@
...
+
+/**
+ * Establish ExpressJS powered web server
+ * @param {object} db a Mongoose DB
+ * @param {object} orm an SQL DB
+ */
+function startExpress(db, orm) {
+  return new Promise(function (resolve, reject) {
+    // Initialize the ExpressJS web application server
+    const app = express.init(db, orm);
```
mleanos: What will `orm` be used for once it's implemented in `express.init`?

server/config/env/development.js
```diff
@@ -12,6 +12,17 @@ module.exports = {
     // Enable mongoose debug mode
     debug: process.env.MONGODB_DEBUG || false
   },
+  orm: {
```
mleanos: I expressed this in a comment below. Should this be a default config setting? Or can this be commented out?

mleanos:
I think we should keep an open dialog on the changes and/or plans going into this new branch. I'm interested to know your thoughts on how we should handle the business logic that will utilize these models inside our API methods. 

I've done some work recently with building a Node.js API from scratch using Typescript. I've managed to separate my models from the database logic, but the service layer that I built relies heavily on the DB I'm using (currently MongoDB), since the way we interact with the database is very specific to what data storage we use.

server/config/lib/app.js
lirantal: Right, it's missing an actual implementation as I didn't yet create any routes/controllers the models so it's still a "mystery" until I put more code there. It is however ready to be used if you need access to any of the ORM models, the sequelizejs instance or the Sequelizejs library itself.

server/config/env/development.js
lirantal: I'm performing changes in chunks so it is easier to follow and process, meaning now I only refactored mongoose to be able to support both that and sequelizejs. When we get to the configuration refactoring we will definitely improve that through-out the codebase.

lirantal:
@mleanos thanks for reviewing, with regards to your comments at https://github.com/meanjs/mean/pull/1687#pullrequestreview-14344830 about the databases - I want to keep the elegant and lean MEAN.JS philosophy which means not very opinionated but rather providing a lot of flexibility and options. In this sense, you could either use an SQL database using sequelizejs, or Mongoose, or both, it's completely your choice and you're not tied to one or the other.

codydaig:
Sorry I'm late to this conversation, however, I would like to mention that I am against sequelize. I used it for a project last year and ran into a lot of major issues along the way. I do not like the way the project is set-up and the way you have to handle advanced queries. I often found it making excessive calls to the database to populate items. I don't remember the exact method, but it ended up making ~5 database calls for information I didn't need. In the end, any queries other than a simple insert or retrieve was written manually using raw SQL. I think it would be better to write our own queries or use something less structured like knex.

mleanos:
>In this sense, you could either use an SQL database using sequelizejs, or Mongoose, or both, it's completely your choice and you're not tied to one or the other.

@lirantal How would one of our users choose one over the other? Meaning, if they chose to use **only** MongoDB (or vice versa) they would have to make very significant changes to the app logic, and structure to remove the discarded one from their project. This doesn't seem like a huge deal, but it does bring me to reiterating the fact that we should try to use the same models regardless of what database is used. The business logic that communicates with the database, should reside inside services that (more likely than not) has the responsibility of handling the db specific interactions; the code required to update MongoDB is much different than what would be used for SQL databases.

Perhaps it's reasonable to just leave both implementations (Sequelize & MongoDB) as they are being developed right now, and offload some of the responsibility to our users when/if they pick just one db. So in that sense, what you've done is created an example of using both. I'm still trying to wrap my head around how to best approach this concern of decoupling our models from the database.

I've been experimenting with a Typescript (Node.js) backend, and have managed to use the same completely decouple my models from the db logic. However, I still haven't decided how fond I am of what I did there.

Basically, I have a structure like this:
models/ - classes used throughout my project
services/ - handles db interaction (in my case MongoDB using Mongoose)
src/routes - application's express routing resides.
src/controllers - route handlers that rely on the services to perform actions against my models.

Here's the funky (questionable) part:
config/mongoose/models - I use these to configure my Mongoose models with pre/post hooks, statics, methods, and custom validations. My thinking was that I could use other databases in my project and just add any specific logic inside the config/ folder; for instance *config/sequelize/*.

codydaig:
What I do in projects is I have a db.js file that sets up the database. Then inside my models folder each model has their own file. Inside that model file, I export several functions for each operation that I need. 

ie.) exports.getAllUsers = function () {} // returns all the users.

That way, if I ever decide to change the underlying database, I just need to change the code inside those functions to use whatever database I want instead, but the parameters it takes in and returns will remain the same.

lirantal:
@mleanos even with the current code committed no one is forced to use sequelizejs, you can easily opt out by simply not configuring the db settings.

@codydaig I don't have any experience with sequelizejs yet and I'll check out knex, but obviously db abstraction layers have overheads. Did you move today to knex or anything else that you recommend from actual experience?
Also, the models are instantiated using their own model file in each module, it's very similar to the way things are with mongoose. I got your comment on the generic function exports, it's a good approach which is taken to the controller level but can definitely happen in the model itself by assigning methods to the sequelizejs model.

lirantal:
btw @codydaig, googling around I found this comment, which if true might prove disappointing for choosing knex:
image

codydaig:
I like the idea of abstracting the db operations to the models that way you don't have to touch controllers or every single place the same query is done if you want to swap databases, you only have to change the one spot and it changes everywhere. 

I've messed with Knex but have not taken it into production. I actually just do raw SQL queries in that project where I've abstracted the queries out. 

If we stick with sequelize, I'd still like to see it abstracted into the models and wrapped in our own functions. That way the DB is easily interchanged with whatever you want.

mg1075:
@lirantal - does this same commenter "taoeffect" later conclude knex is safe?
https://github.com/okTurtles/group-income-simple/issues/102

image

In general, ORM tools are nice for plug-n-play CRUD operations, but not so much for more complicated queries. Where mongoose is tailored for mongodb, and sequelize is tailored for sql-friendly dbs, there are other ORMs supporting both mongo and postgres. Not sure how good these are or what the pitfalls might be; e.g., reaction to waterline seems mixed. 

- CaminteJS
  http://www.camintejs.com/
- JugglingDB
  http://1602.github.io/jugglingdb/
- Waterline (helps power SailsJS)
  http://waterlinejs.org/
- Others?

Or forget ORM altogether...
https://blog.codinghorror.com/object-relational-mapping-is-the-vietnam-of-computer-science/

mleanos:
>I like the idea of abstracting the db operations to the models that way you don't have to touch controllers or every single place the same query is done if you want to swap databases, you only have to change the one spot and it changes everywhere.

This is similar to what I have. However, I prefer to have my db operations inside server-side services. @codydaig, your way works since you're exporting methods rather than ES6 classes. Once you start moving toward using classes, inheritance, interfaces, and exporting them as classes it will become more brittle. Even though I'm using Typescript, I think the same would be true of a project using ES6 classes with Javascript; considering the ts gets compiled to js anyways.

For me an ideal solution would be to use the same ES6 classes for my models, and plug them into whatever ORM (or some other service/library that handles db operations & relationships) I choose to use. Furthermore, if my db operations logic resides somewhere besides my models then I can easily mix in other database technology operations inside my services. For instance, I can have a service `UsersService` that has a method `getAll` which uses Mongoose, and another method `getStatisticalAnalysis` which pulls data from a SQL database where I store user usage stats.

vaucouleur:
Transactions are other important aspect to keep in mind.

Transactions often span across several models.

```
return db.sequelize.transaction(function (t) {
      // blah blah blah..
});
```

If your goal is to encapsulate the db logic, where will the transaction be created ? in the models as well ?

mleanos:
>IMHO, "setting the boundaries of the transactions" is more the concern of the controllers.

I take it a step further and treat my controllers as merely route handlers. These handlers are only responsible for interacting with the Express requests/responses, and they get the data from my server-side services.

PR #1694, opened 2016-12-27
[v2] MEAN.JS API - feat(tests): introducing ava

lirantal:
fix(tasks): re-factoring the tasks module:
* fixes incorrect api endpoints for CRUD actions
* refactor tasks module to be task->comments (one to many) relation which makes more sense, and has the user as an object instead of its own entity in the database (as it already exists on the mongodb)	
* introducing support for ava test runner and assertions
* adding integration tests to sanity test the API end-points
* refactoring app.js for better modularity, exposing a bootstrap method before starting the web server


server/modules/tasks/server/models/sequelize/comments.server.model.sequelize.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+module.exports = function(sequelize, DataTypes) {
```
mleanos: What is `sequelize` here? Why is it necessary to pass into here?

If it's an instance of the Sequelize library, I'm guessing that it's needed because the library doesn't expose a singleton the way that Mongoose does after the first require. Is that right?

lirantal: It's a sequelize model and we are instantiating them automatically from `config/lib/sequelize.js` where they get 'required' and called with those arguments.

Issue #1625, opened 2016-11-10
docker-compose up  doesnt pick the changes in new meanjs that i downloaded from git


jpca999:
it keeps on referring to another code in another directory and will reflect changes from that directory..  Can someone pls help me resolve this ?

Here's my docker-compose.yaml

version: '2'
services:
  web:
    restart: always
    build: .
    container_name: meanjs
    ports:
     - "3000:3000"
     - "5858:5858"
     - "8080:8080"
     - "35729:35729"
    environment:
     - NODE_ENV=development
     - DB_1_PORT_27017_TCP_ADDR=db
    depends_on:
     - db
    volumes_from:
     - web-data
  web-data:
    build: .
    entrypoint: /bin/true
    volumes:
     - ./:/opt/mean.js
     - /opt/mean.js/node_modules
     - /opt/mean.js/public
     - /opt/mean.js/uploads
  db:
    image: mongo:3.2
    restart: always
    ports:
     - "27017:27017"
    volumes_from:
      - db-data
  db-data:
    image: mongo:3.2
    volumes:
      - /data/db
      - /var/lib/mongodb
      - /var/log/mongodb
    entrypoint: /bin/true

Issue #1621, opened 2016-11-09
when running docker-compose up  it doesnt pick up th echnages in the files. 


jpca999:
I also get following error in the console. 

![image](https://cloud.githubusercontent.com/assets/1026325/20153204/9fabf564-a68f-11e6-8de4-884eb6959acb.png)

jpca999:
its wroking on the latest git clone..

Issue #1110, opened 2015-12-19
Live Reload Development and deployment With Docker


5un:
Hello, I'm not sure if the current docker run command with live reload inside README.md reflects the current structure and configurations of the project.

The current: 
```
docker run -p 3000:3000 -p 35729:35729 -v /Users/mdl/workspace/mean-stack/mean/public:/home/mean/public -v /Users/mdl/workspace/mean-stack/mean/app:/home/mean/app --link db:db_1 mean
```

- Does mean/app needs to be mean/modules ?
- I'm not sure why but I cannot get the container running properly by mounting both the /public and the /modules folder. On OSX (El Capitan), The container can start and go through all the grunt build processes and did start the web app at port 3000. But accessing the docker VM's address at port 80 in a browser shows a blank page. Turns out that javascript console says angular is undefined and my /public folder stays empty in my host file system. This is probably caused by docker being unable to write into my host's /public directory.
- I can get it working if I mount just the /modules directory, for example 
```
docker run -p 3000:3000 -p 35729:35729 -v /Users/mdl/workspace/mean-stack/mean/modules:/home/mean/modules --link db:db_1 mean
```

Is there any necessities for /public to be mounted as a volume? I think it's totally fine for the container to write into its own /public and I need to rebuild container if my non-build public files need update (for example public/robot.txt)

PR #1701, opened 2017-01-14
fix(core): Setting legacy watch option on nodemon for docker usage

mcicoria:
This change is needed to get nodemon to work properly when run inside of a docker container. 

To replicated the issue run docker-compose up then make changes to files. You'll notice nodemon does not restart. 

This may be related to: #1110 #1621 #1625


simison:
Isn't legacy watch pretty resource hungry? I.e. could this be `true` only when we're inside Docker?

lirantal:
@mcicoria are you using docker as an actual development environment though?
because I don't see the point in adding this for production reasons

PR #1751, opened 2017-04-08
Removing unused mock-fs library which caused tests to fail

lirantal:
mock-fs will override node's core fs calls which mess up the tests.


lirantal:
Node v6 and v7 pass all tests.
versions 4 and 5 fail due to protractor issues which I'll check in another PR.

PR #1760, opened 2017-04-11
build(snyk): enabling snyk again on ci

lirantal:


lirantal:
FYI - due to the fact that Travis CI builds are not exposing private environment variables to forked then snyk test will always fail, so for now this PR is on-hold until we can get a workaround from Snyk.

mleanos:
@lirantal Any update on this?

lirantal:
we're blocked on the snyk platform finding a solution to work around travis-ci implementation which doesn't allow the token to be used for forked repos which is what happening when PRs are being run.

We're already integrating with Snyk through GitHub so we're ok.

lirantal:
don't need this anymore, closing.

PR #1767, opened 2017-04-20
fix(users): Duplicate Provider Accounts

mleanos:
Fixes an issue where a provider account could be added to the logged in
user's additionalProviderData, when another user has already connected
that same account.

Refactors the saveOAuthUserProfile method to consider the results of the
provider search query, whether or not the current user is logged in.

Steps to reproduce:

1. Signup for a new MEANJS account using a social provider account.
2. Signup for a new MEANJS account using local provider.
3. While still logged in with the User created in step 2, add the same social provider account used in step 1.

Results will be that the local provider User will be connected to the social provider account that was created in step 1.

### DIFF
--- a/modules/users/server/controllers/users/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users/users.authentication.server.controller.js
@@ -122,72 +122,81 @@ exports.oauthCallback = function (strategy) {
  * Helper function to save or update a OAuth user profile
  */
 exports.saveOAuthUserProfile = function (req, providerUserProfile, done) {
-  // Setup info object
+  // Setup info and user objects
   var info = {};
+  var user;
 
   // Set redirection path on session.
   // Do not redirect to a signin or signup page
-  if (noReturnUrls.indexOf(req.session.redirect_to) === -1)
+  if (noReturnUrls.indexOf(req.session.redirect_to) === -1) {
     info.redirect_to = req.session.redirect_to;
+  }
 
-  if (!req.user) {
-    // Define a search query fields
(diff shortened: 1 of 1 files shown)


lirantal:
@mleanos  Seems that test pass but I'm not sure our tests actually go through the flow impacted by this code.

mleanos:
@lirantal I was thinking the same thing. I'll look at adding some unit tests. Some of the logic in this flow is handling edge cases, and we don't have unit tests written for the Social Provider feature :/

lirantal:
Right. It might be quite overwhelming to add some tests for that so if you could at least manually test some of the flows related to login with non-social as well as with social to make sure things are playing nice would be good as well at this point.

mleanos:
@lirantal Sorry for my lack of response. I was putting this off so I could seriously consider adding Social Provider tests to the coverage. At this point, it doesn't look feasible to add tests for the various social providers since it would involve setting up developer accounts for each 3rd party API we'd like to test. We'd need access to these credentials during the build process which could pose some additional considerations with security & infrastructure.

As for these changes, I've definitely manually tested this with the many scenarios that I could think of. Everything is behaving as expected & fixes the bug that this PR is intended for. I'm fine with merging this, and starting a discussion on test coverage for the Social Provider feature.

WDYT?

lirantal:
We could probably mock in a way the social logins but yeah, that's a big undertaking at this point with the project.

@mleanos tthanks for validating the manual testing for it, please merge.

mleanos:
@lirantal I thought about mocking the social providers as well. I'll try working on that over the next few days.

PR #1776, opened 2017-05-18
feat(dependencies): delete PhantomJs

PierreBrisorgueil:
Upgrading protractor in order to upgrade chromedriver to 2.29 + (today we are in 2.26)

This will fix some old issues with chrome driver


lirantal:
awesome, thanks!
seems like only v4 and v5 pass, but v6 and v7 still fail, can you take a look?

PierreBrisorgueil:
Awesome ? it's nothing ! :p

fixed on my side, on mean with material (roMEANet repo), I will add this in the PR quickly, we will see if it's ok.

I switch karma and protractor on chrome with a little fix for Travis, delete Firefox, and everything seems to be ok for us with Travis. 

This switch is important for us, for good test in e2e when you want to play with the size of the window for example ( bug of chrome driver before 2.28)

lirantal:
yeah, that makes sense.
I actually want to remove support for node 4 and 5 so if we can make the build pass for 6 and 7 that would be a great way forward.

PierreBrisorgueil:
hum, ok for linux, i will have a look for other env on Travis as soon as possible :)

lirantal:
@PierreBrisorgueil can you look at the changes I made for protractor setup/gulpfile and see if any of your changes are still required? if so, just update this PR with the relevant changes so I can review.

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: node_js
 sudo: false
+dist: trusty
```
simison: This went in already here: https://github.com/meanjs/mean/pull/1793

lirantal:
@PierreBrisorgueil would you like to rebase and re-try this PR?

PierreBrisorgueil:
@lirantal yep, have a look on this today

simison:
@PierreBrisorgueil did you see Travis failing for some OSs?

PierreBrisorgueil:
@simison Yes, probably due to chrome switch, it's like before the rebase,  i will have a look on this quickly 

Thx

PierreBrisorgueil:
@lirantal Hello, fix, with sudo: required ... :/ don't find other way for the moment to leave phantomjs and clean install of google on all env ...

lirantal:
interesting. @PierreBrisorgueil why would we want to explicitly install google-chrome on the server and not just use the current setup we have for it?
@mleanos since you dealt recently with the travis-ci setup, what do you think?

PierreBrisorgueil:
@lirantal I was updating protractor to update chromdriver for test problems in css, and I discovered it https://news.ycombinator.com/item?id=14105489

So I decided to start deleting PhantomJs and choose what we have on our dev environment, google-chrome

lirantal:
yeah that's a good point Pierre. So basically we're talking about the unit tests, not E2E.
So this isn't anymore about upgrading gulp, right?

package.json
```diff
@@ -105,7 +105,6 @@
     "karma": "~1.6.0",
     "karma-chrome-launcher": "~2.0.0",
     "karma-coverage": "~1.1.1",
-    "karma-firefox-launcher": "~1.0.0",
```
mleanos: We probably want to remove `karma-phantomjs-launcher` (a few lines below) as well, right?

mleanos:
Removing PhantomJS from the project seems very reasonable, after reading that YCombinator thread. I'm fine with it.

>So this isn't anymore about upgrading gulp, right?

#1805 Already upgraded gulp-protractor

To recap these changes, and to ensure I understand their impact..

1) We're removing PhantomJS which requires us to have Chrome installed in the environment. 

2) Now, since Chrome will be installed, we have no need for the Protractor webdriver-manager setup with in the Travis build config.

This sounds good to me, and we'll probably want to remove other references of PhantomJS from the project, correct?

mleanos:
I'm glad to see the Node 8 builds passing as well. Any idea what fixed that issue? Previously, we were getting an error thrown "Bower not found".

package.json
mleanos: and `phantomjs-prebuilt`

mleanos:
@simison WDYT?

There might be concern with becoming too dependent on Chrome existing in each environment that the client-side & E2E tests will be run from. I'm sure most devs & container/image environments would have the capabilities to install Chrome (chromedriver), so it's probably ok.

PierreBrisorgueil:
@lirantal yep, change during the road, i rename the pr :)

@mleanos yep, i remove this quickly

lirantal:
Alright cool.
What's the status on the PR then?

mleanos:
IMO, it's ready to be merged.

mleanos:
@simison WDYT?

I'm looking forward to moving away from PhantomJS. I'm still having issues with the builds on the `webdriver-manager update`.

lirantal:
Agree.

simison:
LGTM

I tested this by implementing it to Trustroots as well: https://github.com/Trustroots/trustroots/pull/525 — all works fine on Node.js 6 :thumbsup:

On Node.js 8 I had Chrome throwing a timeout, which was weird.

I highly recommend adding [karma-mocha-reporter](https://github.com/litixsoft/karma-mocha-reporter) — so much better than Karma's own "progress" reporter:

"progress"-style:
image

vs

"mocha"-style:
image


Thanks @PierreBrisorgueil, good job!

lirantal:
Great stuff, thanks for confirming as well!

PR #1768, opened 2017-04-25
feat(build): updating supported Node.js versions for CI

lirantal:
Deprecating older Node.js versions to streamline development for the LTS and newer versions.
See #1756

PR #1785, opened 2017-05-31
feat(travis): start testing against Node.js 8 and other CI improvements

simison:
- Configure Travis CI test against Node.js 8 and allow failures. See [Node.js 8 release notes](https://nodejs.org/en/blog/release/v8.0.0/).
- Set [fast finishing](https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing) on for TravisCI
- Style corrections to `before_install` TravisCI block
- Create `bower` task to `package.json` scripts and use it instead of direct commands


simison:
You might want to merge https://github.com/meanjs/mean/pull/1768 first add Node.js 8 to that PR instead.

mleanos:
LGTM. 

> You might want to merge #1768 first add Node.js 8 to that PR instead.

I agree

simison:
@mleanos @lirantal I can also update this PR to remove v4 and v5 if you don't want to deal with failing tests at that other repo.

lirantal:
@simison thanks for this PR.
The problem I'm still having with the other referenced PR is to get the E2E selenium tests running. Once we stable that we can definitely add v8 as allowed failures and remove 7 from it.

simison:
@lirantal Thanks, glad to rebase at any time, just ping me here. 

I have a bunch of open sourcing to be done this weekend already so probably won't be able to give hand but wishing you best luck!

lirantal:
@simison ready for you to rebase now after https://github.com/meanjs/mean/pull/1768 is in

simison:
@lirantal any idea what's going on with Travis and Node 8?

First it couldn't find `bower` (it is in `package.json`). I installed it globally at `before_install` section of Travis config. Now that bower works, it can't find simple `lodash` although it's also in `package.json`...

simison:
@lirantal now this passes, but only with Node 8 in allowed failures. I can't figure out why Node modules are not installed and even no error thrown. Maybe there's some cache setting on at Travis?

I tested `cache:false` flag as well but no dice.

I modified initial comment in this PR to list what I did.

lirantal:
you mean travis pass only with node 8 being in allowed_failures?
I haven't used 8 locally either so I'm not sure. does it work on your local machine as usual?

simison:
>you mean travis pass only with node 8 being in allowed_failures?

Yep.

>I haven't used 8 locally either so I'm not sure. does it work on your local machine as usual?

I'm also just observing it trough Travis since locally I'm running 6.10

>looking at travis log I don't see an attempt to try and install bower.

It's in `package.json` so it should happen with `npm install` (and it does work with v7 and others).

Just for the sake of testing, I did have this there earlier; `npm install bower -g` — and that worked, but then it hit in missing `lodash` package (also in `package.json`) so something weird is in play there with that `npm install` not working as it should.

lirantal:
node 8 will also use npm 5 and they are both rather new so maybe that's part of unstable results you're seeing.

simison:
Yup, I'll have to do some more studying and testing locally, just didn't expect something as basic as `npm install` having breaking changes. ;-)

I'm also puzzled because this worked just fine at [Trustroots's Travis config](https://github.com/Trustroots/trustroots/blob/master/.travis.yml#L4)

Meanwhile feel free to merge this already, as I don't really know when I'll get to this again. Hopefully within a week!

simison:
@lirantal if you have time, could you check if there are any "clear cache" buttons at Travis and then we could re-run [Travis build ](https://travis-ci.org/meanjs/mean/builds/241790383). WDYT?

mleanos:
@simison There is a [caching feature](https://docs.travis-ci.com/user/caching) within Travis. However, we don't have any cache's setup for meanjs. As a way of testing, couldn't you add `npm cache clean` to the travis config to be sure there is no NPM caching going on?

mleanos:
@simison It appears that `npm cache clean` doesn't work with NPM 5, unless we want to use `--force` with the command. So my suggestion to use it isn't going to help in this case.

image

simison:
@mleanos  yep, I've tested with `cache:false` in Travis config though.

lirantal:
@mleanos right about the --debug vs --inspect thing, that gulpfile also needs an update there.

simison:
@mleanos

>Unlike the Travis build, I didn't encounter any issues with npm install or Bower. 

So weird!

>One minor issue I ran into is that with Node v8 the --debug option isn't valid, that prevented the server from starting successfully, and caused the E2E tests to fail. Adding 8 here fixed the issue.

Good catch! Did that change. Could you test if it worked?

lirantal:
Looks like it builds fine.

mleanos:
LGTM. I haven't tested locally, but with the changes to the Gulp file (using semver) the build is passing for Node 7. So it's safe to say it works.

As for the Bower issue with Node 8 during Travis, I'm not sure what's going on. We can merge this, and keep an eye on the Node 8 Travis builds. I doubt the issue is with MEANJS at this point.

lirantal:
Agreed.

PR #1802, opened 2017-06-28
fix(security): updating vulnerable packages

lirantal:


mleanos:
These build errors with the Selenium driver 😕

lirantal:
oh my god, this is terrible :(

simison:
@lirantal did you go trough [helmet's changes](https://github.com/helmetjs/helmet/blob/master/HISTORY.md#300--2016-10-28) ?

You probably need to update at least [hsts options](https://github.com/meanjs/mean/blob/f53db0f2e24c5de67bc43566e1103cf7d53edf80/config/lib/express.js#L150) from milliseconds to seconds.

lirantal:
you're right, I haven't reviewed it carefully rushed quickly to resolve the vulnerable dependencies first to have a good baseline of packages. I also remembered they also changed the middleware names like contentSecurityPolicy to csp and such. 
With the build failing on the E2E it's a painful process to push changes and get quick feedback, we need to fix them.

simison:
I updated Helmet stuff for Trustroots, you might be able to just copypaste from there: https://github.com/Trustroots/trustroots/blob/master/config/lib/express.js#L197-L449

lirantal:
@simison you guys also use milliseconds representation for the SIX_MONTHS period so you need to adjust it, or did I miss something?

simison:
@lirantal correct, I noticed that too and that's how I looked up that `moment.duration` code bit. :-)

Thanks!

config/lib/express.js
```diff
@@ -141,7 +141,7 @@ module.exports.initModulesConfiguration = function (app, db) {
  */
 module.exports.initHelmetHeaders = function (app) {
   // Use helmet to secure Express headers
-  var SIX_MONTHS = 15778476000;
+  var SIX_MONTHS = 15778476;
```
simison: Would it make sense to just use `15778476` directly in hsts settings object since this variable isn't used anywhere else?

simison: Also it'll be useful for anyone modifying this to have a short comment mentioning it's seconds.

lirantal: nah because then it's just magic number, very unreadable.

lirantal: yeah I wasn't sure if to add that comment or not because someone who would change it will probably be aware of helmet

mleanos: FWIW, I prefer having a note that it's in `seconds` here. Even those that understand that Helmet expects seconds, it may be easy to forget or be overlooked by others coming in.

To me it looks less like a "magic" number when I see a comment about `seconds`.

lirantal: Sure, updated.

mleanos:
LGTM. Love seeing those green check marks!

simison:
LGTM

PR #1805, opened 2017-06-29
fix(build): E2E coverage broken on Travis CI

mleanos:
Fixes an issue with the Travis CI builds failing at the E2E tests due to
an apparent issue with the Mozilla Gecko web-driver used with Selenium.

1) Upgrades `gulp-protractor` to ensure Protractor v5> is used.
2) Removes `--firefox` option from the Travis CI Selenium
`webdriver-manager update` step.
3) Changes Protractor config to use Chrome instead of Firefox.

**NOTE:** This PR was originally created on a debugging branch to determine the issues with the E2E tests running on Travis CI builds. The below conversation will reflect the process of numerous experimental commits to track down the source of the issues. After a conclusion was made, the commits in this branch were squashed, and the commit details were changed here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_install:
   - npm install protractor
   - 'export DISPLAY=:99.0'
   - 'sh -e /etc/init.d/xvfb start'
-  - 'node_modules/protractor/bin/webdriver-manager update --standalone --firefox'
+  - 'node_modules/protractor/bin/webdriver-manager update --standalone'
   - 'node_modules/protractor/bin/webdriver-manager start 2>&1 &'
   - sleep 3
 #before_script:
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "gulp-mocha": "~3.0.1",
     "gulp-ng-annotate": "~2.0.0",
     "gulp-nodemon": "~2.2.1",
-    "gulp-protractor": "^3.0.0",
+    "gulp-protractor": "^4.0.0",
     "gulp-refresh": "~1.1.0",
     "gulp-rename": "~1.2.2",
     "gulp-rev": "^7.1.2",
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -7,7 +7,12 @@ var config = {
 
 if (process.env.TRAVIS) {
   config.capabilities = {
-    browserName: 'firefox'
+    // Without this setting, Travis CI would default
+    // to using Chrome anyway.
+    // NOTE: Firefox is currently not working with
+    // the Travis CI builds. For more info see:
+    // https://github.com/meanjs/mean/pull/1805
+    browserName: 'chrome'
   };
 }
 


mleanos:
@lirantal @simison Any thoughts here?

Why are we using Firefox, rather than Chrome?

Switching to Chrome and updating `gulp-protractor` seemed to have fixed the E2E tests.

However, I have no idea what's going on with this (it only happens with the first job on each Node version) 😕 

https://travis-ci.org/meanjs/mean/jobs/248564363#L516

Could this be an issue with Travis' "trusty" image?

lirantal:
I didn't setup the E2E initially so I don't know the historical versions of using firefox, maybe it was only available at that time on Travis CI.
Also good point on the trusty image since it's a change I did recently in attempt to speed up builds so maybe related and we can try to roll back that setting.

simison:
Perhaps something changed with on 21st? https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch

simison:
I remember a few months ago someone tried a PR for Trustroots where they changed browser to Firefox (because [no more PhantomJS](https://groups.google.com/d/msg/phantomjs/9aI5d-LDuNE/5Z3SMZrqAQAJ)). It threw some errors so I assume it had some similar issues.

Would be awesome to support Firefox but seems like not worth the effort right now.

simison:
@mleanos did you test this on Travis config?
```
addons:
  firefox: latest
```

mleanos:
No. I haven't. :) I've exhausted almost every other config/option that I could think of, or find on the webs. I'll try this now. Thanks :)

mleanos:
>However, I have no idea what's going on with this (it only happens with the first job on each Node version) 😕

>https://travis-ci.org/meanjs/mean/jobs/248564363#L516

Seems to have gone away.. again with the 😕 smh

simison:
@mleanos the problem wasn't neither one of these listed at trouble shooting, right?

https://docs.travis-ci.com/user/gui-and-headless-browsers/#Troubleshooting

mleanos:
This is the error that is getting thrown each time I try to use Firefox.. 
https://travis-ci.org/meanjs/mean/jobs/248961045#L2753

I tried both and neither worked (I saw that someone downgraded to FF 41 & it solved the issue for them)..
```bash
addons:
  firefox: latest
```
```bash
addons:
  firefox: 41
```

The troubleshooting link you sent seems to apply to Karma, and not Protractor, since our client-side tests run fine, but the E2E is failing.

It's gotta be an issue with that Gecko driver, and I've tried setting a `$PATH` for it, as you can see with a couple of my commits.

mleanos:
Please bear in mind that I'm learning the nuances of Travis CI as I go :) So there just might be something that I'm overlooking. As it stands, it doesn't look feasible to use Firefox at the moment. Unless we can figure out that Gecko driver path issue. It's odd that it popped up recently.

The last thing I haven't tried is to remove the use of the "trusty" image.

simison:
I think comment here says it all:
>The first overall goal of headless browsing is to make automated testing of Firefox less painful for developers.
https://bugzilla.mozilla.org/show_bug.cgi?id=1338004
:-D

Totally not worth the pain yet — hopefully this will be easier eventually.

lirantal:
Keep at it @mleanos!

mleanos:
Out of curiosity, what's the benefit of using Firefox over Chrome in this instance?

One thing that I've always felt, is that FF isn't tied to Angular, so it would probably be safe to say "if it works in FF, it will work in Chrome." - Is that the general sentiment?

mleanos:
> Keep at it @mleanos!

I may have run out of options though. So in order to get out builds to pass, we may just have to go with Chrome for now.

mleanos:
@lirantal @simison I propose we merge these changes in so our builds start passing again. We can just go with Chrome for now, and keep an eye on the Firefox issue.

I squashed the original commits in this PR, and cleaned up the commit message.

lirantal:
I don't have a specific preference for FF really. I think maybe it was the only option available back then for Travis CI.

Great work!

codydaig:
LGTM!

PR #1811, opened 2017-07-11
lock node/npm versions.

prashanthmadi:
lock node/npm versions else it would try to use npm5 if available which has some known issues and not stable yet.

I'm using node 6.9.1 as an example but may be we can lock it for other node version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,8 @@
     "url": "https://github.com/meanjs/mean.git"
   },
   "engines": {
-    "node": ">=6.10.0",
-    "npm": ">=3.10.8"
+    "node": "6.9.1",
+    "npm": "3.10.8"
   },
   "scripts": {
     "update": "npm update && npm prune && npm run bower",


simison:
What system is "trying to use npm5"? As far as I know these variables aren't really used anywhere and are merely just meant for information (and Node.js would probably just warn if there's a mismatch). Tell us more! :-)

>I'm using node 6.9.1 as an example but may be we can lock it for other node version.

Yep, locking exactly to minor version is a bit brutal... This should allow up to working tested version (Node.js 7 currently).

lirantal:
can you update the versions to include a range that allows the current versions and use a maximum limit of node v8 and npm 5? i.e: 8.0.0 <= nodeVersion <= 6.9.0

PierreBrisorgueil:
hello, some tests : 

- node > v7.x with npm 5.X & 3.X 👍 
- node > v6.11.x with npm > 5.X & 3.X OK 👍
- node v6.9.1 with nom 5.x Ok for me 👍

@prashanthmadi what are your issues ? 

On my side npm 5 or not change nothing. 

For me if npm 5 is installed by default, it because it's officially released with your node, and in this case the problem of instability is rated to npm, not us. But I see no instability. 

However in my case, a clean install of node 6 does not install npm 5. I have to upgrade it myself.

(Tested with nvm)

However, @lirantal it maybe the time to add the package-lock.json to the repo ? 

https://docs.npmjs.com/files/package-lock.json
https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

prashanthmadi:
Thanks for help with this.

Issue is running meanjs in Azure App Services. They have npm5 and below issue is fixed in npm 5.1 
https://github.com/npm/npm/issues/16766 

If we have npm>=x.x, it would try to use latest version and fail. This might be one-off scenario but might be a good practice to lockdown a version.

simison:
@prashanthmadi if there's a fix in 5.x series already for this issue, locking this down under <5.0.0 isn't really a solution, but upgrading your npm is. :-)

I'd say you should change that value for your project and we can keep it to what it was here.

Here's a semver which skips v5.1: `>=3.6 <5.1 || >5.1`

Here's a handly tool: https://semver.npmjs.com/

prashanthmadi:
ok. I have created a fork and it requires below commit diff changes to run meanjs in azure app services.
https://github.com/prashanthmadi/mean/commit/30a6a7ef4258cd69a22b5658e670cc7bf2a476b2

Issue #1814, opened 2017-07-13
Upgrade Mongoose to 4.11


simison:
Just wanted to give heads up:

Mongoose 4.11 has quite some changes which affect Mean.js and upgrading requires some changes to the codebase.

It could be a good change to update to [new mongoose connection logic](http://mongoosejs.com/docs/connections.html#use-mongo-client):
>Mongoose's default connection logic is deprecated as of 4.11.0. Please opt in to the new connection logic using the `useMongoClient` option, but make sure you test your connections first if you're upgrading an existing codebase!

For instance `connect-mongo` [now accepts promises](https://github.com/jdesboeufs/connect-mongo#re-use-a-native-mongodb-driver-connection-or-a-promise) as well.

I probably won't be able to make a PR any time soon though...

simison:
I'm also getting these errors when running tests when attempting upgrade:
```
MissingSchemaError: Schema hasn't been registered for model "User".
Use mongoose.model(name, schema)
    at Mongoose.model (/mean/node_modules/mongoose/lib/index.js:395:13)
    at Object.<anonymous> (/mean/modules/core/tests/server/core.server.config.tests.js:9:19)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at /mean/node_modules/mocha/lib/mocha.js:230:27
    at Array.forEach (native)
    at Mocha.loadFiles (/mean/node_modules/mocha/lib/mocha.js:227:14)
    at Mocha.run (/mean/node_modules/mocha/lib/mocha.js:495:10)
    at Object.<anonymous> (/mean/node_modules/mocha/bin/_mocha:469:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
```

I upgraded also Mocha so this might be coming from there as well.

lirantal:
Thanks for the heads up!

simison:
[...]

PR #1816, opened 2017-07-13
[WIP] Update Mongoose

simison:
Just work in progress branch of mine which currently has several things I was testing and learning around updating Mongoose from 4.10 to 4.11. I got the app working, but tests are still failing.

I might get back to this at some point, but meanwhile if anyone else wants to take a stab at updating Mongoose, feel free to look at this.

- Passing user/pass in connection's `options` throws deprecated warning so I removed it, I saw an issue where it was mentioned they might add it back but I lost it. Gotta double check.

- Using [new mongoose connection logic](http://mongoosejs.com/docs/connections.html#use-mongo-client) by setting: `useMongoClient:true`

- Above will make everything into promises so I'm using those in several places instead of callbacks, but I didn't spend too much time in refactoring. I think the whole thing could do with proper rewrite.

- I'm using `mongooseService.loadModels()`'s callback everywhere but this probably isn't needed, I was just trying to get rid of errors with tests.

- Running tests throws "Schema not defined" errors so here are some solutions I was testing

- Rename `mongoose` variables to `mongooseService` so that it's obvious its not referring to mongoose module but our own internal service.

Read more; #1814


mleanos:
@simison Any particular reason for the other package dependency upgrades in package.json?

I'm planning on spending some time on this, and agree that we could use a re-write of our mongoose service.

simison:
@mleanos nope, I should actually remove that commit from here and open a PR for it.

If you can take over this would be aaaawesome!

mleanos:
> If you can take over this would be aaaawesome!

That was my intention. Do keep an eye on it though. I'll probably need some guidance/review from you at some point 😸

simison:
Sure :thumbsup: just ping me when there's something to check, I'll be busy at a hackathon over the weekend but I can check over something if need.

simison:
Closing as duplicate of #1818

Thanks @mleanos !

PR #1817, opened 2017-07-13
feat(Karma): Add mocha style progress to Karma tests

simison:
Makes writing and debugging tests with Karma so much easier.

Looks like this:
image

Instead of this:
image

My only worry would be that I don't know how well it will be maintained, but on the other hand reversing is no-brainer.


simison:
>the report output has really bothered me for quite a while

Same, crazy I didn't realise to look into it earlier.

PR #1818, opened 2017-07-15
feat(config): Mongoose 4.11 upgrade

mleanos:
Upgrades Mongoose to ~~4.11.1~~ 4.11.3

Updates Mongoose connection implementation to accommodate deprecated
features & connection options.

Also, updates the Gulp Mocha tasks to reflect changes to the Mongoose
implementation.

Fixes tests to get the database from the existing Mongoose singleton.

Derives from changes in https://github.com/meanjs/mean/pull/1816

Closes https://github.com/meanjs/mean/issues/1814

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -5,10 +5,7 @@ var defaultEnvConfig = require('./default');
 module.exports = {
   db: {
     uri: process.env.MONGOHQ_URL || process.env.MONGODB_URI || 'mongodb://' + (process.env.DB_1_PORT_27017_TCP_ADDR || 'localhost') + '/mean-dev',
-    options: {
-      user: '',
-      pass: ''
-    },
+    options: {},
     // Enable mongoose debug mode
     debug: process.env.MONGODB_DEBUG || false
   },
--- a/config/lib/mongoose.js
+++ b/config/lib/mongoose.js
@@ -3,7 +3,8 @@
 /**
  * Module dependencies.
  */
-var config = require('../config'),
+var _ = require('lodash'),
+  config = require('../config'),
   chalk = require('chalk'),
   path = require('path'),
   mongoose = require('mongoose');
@@ -19,30 +20,31 @@ module.exports.loadModels = function (callback) {
 };
 
 // Initialize Mongoose
-module.exports.connect = function (cb) {
-  var _this = this;
-
+module.exports.connect = function (callback) {
   mongoose.Promise = config.db.promise;
 
-  var db = mongoose.connect(config.db.uri, config.db.options, function (err) {
-    // Log Error
-    if (err) {
-      console.error(chalk.red('Could not connect to MongoDB!'));
-      console.log(err);
-    } else {
+  var options = _.merge(config.db.options || {}, { useMongoClient: true });
 
+  mongoose
+    .connect(config.db.uri, options)
+    .then(function (connection) {
       // Enabling mongoose debug mode if required
       mongoose.set('debug', config.db.debug);
 
       // Call callback FN
-      if (cb) cb(db);
-    }
-  });
+      if (callback) callback(connection.db);
+    })
+    .catch(function (err) {
+      console.error(chalk.red('Could not connect to MongoDB!'));
+      console.log(err);
+    });
+
 };
 
 module.exports.disconnect = function (cb) {
-  mongoose.disconnect(function (err) {
-    console.info(chalk.yellow('Disconnected from MongoDB.'));
-    cb(err);
-  });
+  mongoose.connection.db
+    .close(function (err) {
+      console.info(chalk.yellow('Disconnected from MongoDB.'));
+      return cb(err);
+    });
 };
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -277,15 +277,15 @@ gulp.task('templatecache', function () {
 
 // Mocha tests task
 gulp.task('mocha', function (done) {
-  // Open mongoose connections
-  var mongoose = require('./config/lib/mongoose.js');
+  var mongooseService = require('./config/lib/mongoose');
   var testSuites = changedTestFiles.length ? changedTestFiles : testAssets.tests.server;
   var error;
 
   // Connect mongoose
-  mongoose.connect(function () {
-    mongoose.loadModels();
-    // Run the tests
+  mongooseService.connect(function (db) {
+    // Load mongoose models
+    mongooseService.loadModels();
+
     gulp.src(testSuites)
       .pipe(plugins.mocha({
         reporter: 'spec',
@@ -296,9 +296,13 @@ gulp.task('mocha', function (done) {
         error = err;
       })
       .on('end', function () {
-        // When the tests are done, disconnect mongoose and pass the error state back to gulp
-        mongoose.disconnect(function () {
-          done(error);
+        mongooseService.disconnect(function (err) {
+          if (err) {
+            console.log('Error disconnecting from database');
+            console.log(err);
+          }
+
+          return done(error);
         });
       });
   });
@@ -362,16 +366,17 @@ gulp.task('karma:coverage', function(done) {
 // Drops the MongoDB database, used in e2e testing
 gulp.task('dropdb', function (done) {
   // Use mongoose configuration
-  var mongoose = require('./config/lib/mongoose.js');
+  var mongooseService = require('./config/lib/mongoose');
 
-  mongoose.connect(function (db) {
-    db.connection.db.dropDatabase(function (err) {
+  mongooseService.connect(function (db) {
+    db.dropDatabase(function (err) {
       if (err) {
         console.error(err);
       } else {
-        console.log('Successfully dropped db: ', db.connection.db.databaseName);
+        console.log('Successfully dropped db: ', db.databaseName);
       }
-      db.connection.db.close(done);
+
+      mongooseService.disconnect(done);
     });
   });
 });
--- a/modules/articles/tests/server/admin.article.server.routes.tests.js
+++ b/modules/articles/tests/server/admin.article.server.routes.tests.js
@@ -23,7 +23,7 @@ var app,
 describe('Article Admin CRUD tests', function () {
   before(function (done) {
     // Get application
-    app = express.init(mongoose);
+    app = express.init(mongoose.connection.db);
     agent = request.agent(app);
 
     done();
--- a/modules/articles/tests/server/article.server.routes.tests.js
+++ b/modules/articles/tests/server/article.server.routes.tests.js
@@ -24,7 +24,7 @@ describe('Article CRUD tests', function () {
 
   before(function (done) {
     // Get application
-    app = express.init(mongoose);
+    app = express.init(mongoose.connection.db);
     agent = request.agent(app);
 
     done();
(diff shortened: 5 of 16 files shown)


mleanos:
I started on a bigger refactor of the Mongoose and Express app implementation. However, I scaled it back to just satisfy the Mongoose upgrade. We can discuss a bigger refactor in a separate issue and/or PR.

Regarding the `{ user: '', pass: '' }` connection options, there is [work being done](https://github.com/mongodb/node-mongodb-native/pull/1529) to add back support of this feature.

lirantal:
@mleanos are you synced with @simison? is this a superset of his PR?

config/lib/mongoose.js
```diff
@@ -24,25 +25,28 @@ module.exports.connect = function (cb) {
...
       // Enabling mongoose debug mode if required
       mongoose.set('debug', config.db.debug);
 
       // Call callback FN
-      if (cb) cb(db);
-    }
-  });
+      if (cb) cb(connection.db);
+    })
+    .catch((err) => {
```
simison: some ES6 here ;-)

gulpfile.js
```diff
@@ -296,10 +296,7 @@ gulp.task('mocha', function (done) {
         error = err;
       })
       .on('end', function () {
-        // When the tests are done, disconnect mongoose and pass the error state back to gulp
-        mongoose.disconnect(function () {
-          done(error);
-        });
+        mongooseService.disconnect(done);
```
simison: I almost did this myself as well, but that `error` being passed to gulp's `done` isn't error from mongoose disconnect. It's the error object from Mocha.

Perhaps that variable needs renaming so that it would be clear what's happening here?

gulpfile.js
```diff
@@ -362,16 +359,17 @@ gulp.task('karma:coverage', function(done) {
 // Drops the MongoDB database, used in e2e testing
 gulp.task('dropdb', function (done) {
   // Use mongoose configuration
-  var mongoose = require('./config/lib/mongoose.js');
+  var mongooseService = require('./config/lib/mongoose');
```
simison: Same as earlier:

Is there any reason why this `require` is inside the task and not on top of the gulp file with all the other dependencies?

gulpfile.js
```diff
@@ -277,15 +277,15 @@ gulp.task('templatecache', function () {
 
 // Mocha tests task
 gulp.task('mocha', function (done) {
-  // Open mongoose connections
-  var mongoose = require('./config/lib/mongoose.js');
+  var mongooseService = require('./config/lib/mongoose');
```
simison: Is there any reason why this `require` is inside the task and not on top of the gulp file with all the other dependencies?

modules/articles/tests/server/admin.article.server.routes.tests.js
```diff
@@ -23,7 +23,7 @@ var app,
 describe('Article Admin CRUD tests', function () {
   before(function (done) {
     // Get application
-    app = express.init(mongoose);
+    app = express.init(mongoose.connection.db);
```
simison: This <3, I just couldn't figure out what was wrong.

mleanos: This was a source of confusion to me too. I didn't see how the line above would work by just passing `mongoose`, but eventually it made sense based on this: https://github.com/meanjs/mean/blob/master/config/lib/mongoose.js#L27. `db` is a promise, no?

If these changes get merged then it's a moot point, I guess. Ideally, we need to refactor the Express & Mongoose implementations. As it stands, after we make the first connection, we're heavily relying on the fact that Mongoose is a singleton. This could be confusing, and probably not the best SOC design.

simison: Yup! 

I guess https://github.com/lirantal/Riess.js might have quite a lot of it thought already.

gulpfile.js
mleanos: If our Mongoose service is only being used in a couple places, why `require` it at the global scope?

My understanding of how `require` works, along with it's memory use, isn't great so maybe it doesn't matter.

simison: Then I don't know more than you do. :-) I was just curious, I guess for now we can trust someone knew what they were doing when they placed it inside the task.

gulpfile.js
mleanos: Yea, good points. It did occur to me, for a moment, about the `error` needing to be passed to the done. Then I think I just overlooked it. I'll change it to pass the Mocha error to the `done` callback. I guess we can just log the error to the console, when the `disconnect` fails; it would be a rare error in this case.

The name `error` seems fine to me. It's clear what's going on since it's right there.

modules/articles/tests/server/admin.article.server.routes.tests.js
simison: And that said I think it's a good idea to keep this to minimum to just get Mongoose upgraded and leave refactoring to a separate PR.

config/lib/mongoose.js
mleanos: Do you think it's ok?

Node.js `6.4.0>` can handle handle arrow functions. So why not?

simison: I'd love to see bigger discussion about this, actually. This topic just came up at Trustroots so I might open an issue here just to get some food for thought.

I don't think it's a good idea to introduce mixed coding style in PR like this.

It's more annoying to me to see mixed style rather than old school JS.

I'd love to see meanjs move on to ES6 but it might be better done with some strategy and thought and build tools (eslint) could probably be upgraded first.

WDYT?

mleanos:
@lirantal Mikael and I discussed in his other PR that I would be taking this over. Once I get a couple things clarified from the review comments, I'll rebase and make the necessary changes.

modules/articles/tests/server/admin.article.server.routes.tests.js
mleanos: Agreed 👍

config/lib/mongoose.js
mleanos: Everything you just said seems very reasonable to me. I'd like to discuss the introduction of ES6 as well. I think it started becoming a habit for me, since I'm using some features like array functions in my projects.

I'll remove my ES6 usage from this PR.

mleanos:
@lirantal @simison Final review?

Based on Mikael's review:
1) Removed arrow function use
2) Added back `done(error)` to the Mocha Gulp task to return error thrown by Mocha rather than the `mongooseService.disconnect` callback.

simison:
Looks good to me now!

config/env/development.js
```diff
@@ -5,10 +5,7 @@ var defaultEnvConfig = require('./default');
 module.exports = {
   db: {
     uri: process.env.MONGOHQ_URL || process.env.MONGODB_URI || 'mongodb://' + (process.env.DB_1_PORT_27017_TCP_ADDR || 'localhost') + '/mean-dev',
-    options: {
-      user: '',
-      pass: ''
-    },
+    options: {},
```
mleanos: Sure thing.

mleanos:
@simison @lirantal I've rebased and squashed my commits.

I also removed the `options` from https://github.com/meanjs/mean/blob/master/config/env/default.js#L10, that I had previously added in a commit on this branch. I just remembered that the default env config doesn't need the `options`, since they are being set in the individual env configs (i.e. development, production, and test).

config/env/development.js
mleanos: Done

gulpfile.js
lirantal: Amos added it way back in 0.4 release branch.
We can move it at the top if it makes sense now, but I think when Amos added it on 0.4 the mongoose.js worked differently from today, as in, just requiring it may have opened the database connection or instantiate models so that would explain why he would have wanted to limit it to the specific gulp tasks.

modules/articles/tests/server/admin.article.server.routes.tests.js
lirantal: Yep. And indeed the returned value of the mongoose.js line 27 was just the db object being resolved with.

modules/articles/tests/server/article.server.routes.tests.js
```diff
@@ -119,7 +119,7 @@ describe('Article CRUD tests', function () {
     // Save the article
     articleObj.save(function () {
       // Request articles
-      request(app).get('/api/articles')
+      agent.get('/api/articles')
```
lirantal: nice catch @mleanos

modules/users/server/config/users.server.config.js
```diff
@@ -11,7 +11,7 @@ var passport = require('passport'),
 /**
  * Module init function
  */
-module.exports = function (app, db) {
+module.exports = function (app) {
```
lirantal: this could potentially break vertical modules that rely on it.
even with the changes in this PR for `config/lib/mongoose.js` how would one be able to access the db instance?

the other alternative I can think of is if we set app.local.db or something like that which would expose the db connection where app is too.

WDYT?

simison: I like that, sounds simple and straightforward. I didn't like at all trying to follow where `db` is going around trough all those callbacks.

That said, this could also wait for more thorough restructuring of mongoose/db stuff as with promises etc.

lirantal: In leaning towards agreeing with 'this can wait' but I wonder what projects built based on master version will do if `db` is now not available for them and they wish to stay up to date...

mleanos: We should note the potential breaking change in the CHANGELOG. 

Anyone merging from master should be aware of the changes, and it seems like it would be easy for them to add back the `db` in their exports where they rely on it. Adding `app.local.db` wouldn't be a difficult task either.

I'm a bit skeptical that many use cases would arise that would require someone to rely on accessing the `db` instance from server-side config/controllers.

I'm not opposed to adding `db` back to all the callbacks though.

lirantal: yeah which is why I'm not totally against removing it.
so let's keep this PR as is and see how the mongodb refactoring continues on. I'm ok with that.

mleanos: Ok. I'm fine with that as well.

mleanos:
@lirantal Are you still waiting on changes before you can approve this?

lirantal:
Nope, I'm good.

config/lib/mongoose.js
```diff
@@ -24,25 +25,28 @@ module.exports.connect = function (cb) {
```
simison: Remove `var _this = this;` as it's never used (weird how eslint didn't warn about this?)

config/lib/mongoose.js
```diff
@@ -24,25 +25,28 @@ module.exports.connect = function (cb) {
...
+    .connect(config.db.uri, options)
+    .then(function (connection) {
       // Enabling mongoose debug mode if required
       mongoose.set('debug', config.db.debug);
 
       // Call callback FN
-      if (cb) cb(db);
-    }
-  });
+      if (cb) cb(connection.db);
```
simison: Personally I prefer using `callback` instead of `cb` just to keep it all the same and I dislike abbreviations in coding (just needs extra thinking when reading it), but this isn't a biggie. :-)

config/lib/mongoose.js
mleanos: Removed. The ESLint rule [no-unused-vars is turned off](https://github.com/meanjs/mean/blob/master/.eslintrc.js#L36), probably because there are a lot of ESLint errors generated from this rule in our tests and other places :/ That can be addressed in a separate PR.

mleanos:
@simison I addressed the changes requested in your last review. LGTY?

@lirantal If everything looks good now, I think this is ready to be merged.

lirantal:
Yep, I'm good.

simison:
LGTM! Feel free to merge.

PR #1822, opened 2017-07-18
fix(gulpfile): Update templateCache root path

### LINKED ISSUE #1820: Template cache in production build?
I am using 0.5.0.

I found that all of my `/modules/*/client/views/*.html` are put into a service named `$templateCache` in the production build `application-xxx.min.js`.

I checked the browser inspector, the browser will still makes new request when I trigger to display a HTML template, e.g. a modal.

I am wondering what `$templateCache` is used for, why do I need to include all the HTML files in the `application-xxx.min.js` while the application will only fetch the HTML when it is needed.

kitce:
fix(gulpfile): Update templateCache root path

In the production build `application-xxx.min.js`, the url of the mark up being put into `$templateCache` is incorrect, the `/` is missing.

Although the mark-ups are already put into `$templateCache`, the browser will still fetch the templates files when they are needed.

Fixes #1820

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -266,7 +266,7 @@ gulp.task('makeUploadsDir', function () {
 gulp.task('templatecache', function () {
   return gulp.src(defaultAssets.client.views)
     .pipe(plugins.templateCache('templates.js', {
-      root: 'modules/',
+      root: '/modules/',
       module: 'core',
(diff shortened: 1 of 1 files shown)


simison:
Tested locally and both `npm run start:prod` and `npm start` work well. Hitting `$templateCache` on prod now. :thumbsup:

simison:
@kitce thanks and congrats for your first PR. :-)

PR #1839, opened 2017-08-11
fix(users): don't fail on missing old image on image upload

### LINKED ISSUE #1778: Error: ENOENT: no such file or directory, unlink './modules/users/client/img/profile/uploads/df9f321cf61c57c89c3e4d0a416e1093.jpg'
Hi,

Please I always have this error
I don't know why

How could I solve it please ?

Thanks.

simison:
Silently fails when existing file didn't exist when unlinking it.

Fixes #1778

### DIFF
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -111,7 +111,15 @@ exports.changeProfilePicture = function (req, res) {
       if (existingImageUrl !== User.schema.path('profileImageURL').defaultValue) {
         fs.unlink(existingImageUrl, function (unlinkError) {
           if (unlinkError) {
-            console.log(unlinkError);
+
+            // If file didn't exist, no need to reject promise
+            if (unlinkError.code === 'ENOENT') {
+              console.log('Removing profile image failed because file did not exist.');
+              return resolve();
+            }
+
+            console.error(unlinkError);
+
             reject({
               message: 'Error occurred while deleting old profile picture'
             });
--- a/modules/users/tests/server/user.server.routes.tests.js
+++ b/modules/users/tests/server/user.server.routes.tests.js
@@ -6,6 +6,7 @@ var semver = require('semver'),
   path = require('path'),
   mongoose = require('mongoose'),
   User = mongoose.model('User'),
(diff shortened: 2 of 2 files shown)


modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -111,7 +111,15 @@ exports.changeProfilePicture = function (req, res) {
...
           if (unlinkError) {
-            console.log(unlinkError);
+
+            // If file didn't exist, no need to reject promise
+            if (unlinkError.code === 'ENOENT') {
+              console.log('Removing profile image failed because file did not exist.');
+              return resolve();
+            }
+
+            console.error(unlinkError);
```
simison: Not that I changed this to log to error log and this won't be logged if we silently fail.

lirantal: * I'm wondering if you maybe want to use the logger instead of console.log/error
* what do you mean it wont be logged?

modules/users/tests/server/user.server.routes.tests.js
```diff
@@ -1078,6 +1079,39 @@ describe('User CRUD tests', function () {
       });
   });
 
+  it('should be able to change profile picture and not fail if existing picture file does not exist', function (done) {
```
lirantal: good stuff on the tests

modules/users/server/controllers/users/users.profile.server.controller.js
simison: > I'm wondering if you maybe want to use the logger instead of console.log/error

Right on, I forgot meanjs has one.

>what do you mean it wont be logged?

Sorry, I meant won't be logged in _error log_ since function gets returned before that line.

simison: >> I'm wondering if you maybe want to use the logger instead of console.log/error
>
> Right on, I forgot meanjs has one.

Wait, doesn't the logger just redirect console.log? If not, then I don't know how to use it; any pointers?

lirantal: we have a generic one but I think we actually never re-vamped all those console.logs to use it properly so can forget it.

we're good to go :-)

PR #1843, opened 2017-08-14
Remove deprecated crypto package

simison:
`crypto` is now part of NodeJS core: https://www.npmjs.com/package/crypto

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "connect-flash": "~0.1.1",
     "connect-mongo": "~1.3.2",
     "cookie-parser": "~1.4.1",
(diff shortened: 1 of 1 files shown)


simison:
Yup and meanjs uses it [only once](https://github.com/meanjs/mean/blob/f65d4b90ca7343f919b4072dbb3f97d4daf17368/modules/users/server/controllers/users/users.password.server.controller.js#L24).  :-)

Since I didn't experience any issues with default nodejs installations, I assume we don't need to check if it's available: https://nodejs.org/api/crypto.html#crypto_determining_if_crypto_support_is_unavailable

PR #1844, opened 2017-08-14
fix(eslint): Inconsistent spacing before function parentheses

### LINKED ISSUE #1605: Inconsistent spacing before function parentheses
There are some places in the code-base that have spaces before function parentheses, and some that don't. I'm seeing changes to these in multiple PR's, that go either way (some are creating unnecessary diffs). We should enforce a linting rule in these cases for reasons of consistency.

One example: https://github.com/meanjs/mean/blob/master/modules/users/tests/e2e/users.e2e.tests.js#L411-L424

My personal preference is the same expressed in this comment: https://github.com/airbnb/javascript/issues/55#issuecomment-101360471

My proposal is to change [this eslint setting](https://github.com/meanjs/mean/blob/master/.eslintrc.js#L43) to

``` javascript
'space-before-function-paren': ['error', {
      'anonymous': 'always',
      'named': 'never',
      'asyncArrow': 'always'
    }]
```

http://eslint.org/docs/rules/space-before-function-paren#anonymous-always-named-never-asyncarrow-always

When testing this new setting, 242 instances where this rule is violated were reported by the linting task. This would be a big sweeping change, but worthwhile.

Anyone have thoughts on whether we want to implement my proposed change?

simison:
Defines `space-before-function-paren` eslint rule and changes files accordingly.

Resolves #1605

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,7 +40,11 @@ module.exports = {
     'one-var': [0, 'never'],
     'one-var-declaration-per-line': [2, 'always'],
     'padded-blocks': 0,
-    'space-before-function-paren': 0,
+    'space-before-function-paren': ['error', {
+      'anonymous': 'always',
+      'named': 'never',
+      'asyncArrow': 'always'
+    }],
     'space-in-parens': [2, 'never'],
     'spaced-comment': [2, 'always'],
     strict: 0,
--- a/config/lib/logger.js
+++ b/config/lib/logger.js
@@ -29,7 +29,7 @@ var logger = new winston.Logger({
 // Useful for integrating with stream-related mechanism like Morgan's stream
 // option to log all HTTP requests to a file
 logger.stream = {
-  write: function(msg) {
+  write: function (msg) {
     logger.info(msg);
   }
 };
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -336,7 +336,7 @@ gulp.task('karma', function (done) {
 });
 
 // Run karma with coverage options set and write report
-gulp.task('karma:coverage', function(done) {
+gulp.task('karma:coverage', function (done) {
   new KarmaServer({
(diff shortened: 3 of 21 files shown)


mleanos:
Coverage increased.. smh 😖

simison:
Yeah I was lazy to figure out how to configure `eslint -fix` to catch correct config files and read correct directories/files, but I wrote a quick gulp task instead. ;-) Essentially just `fix:true` option in our `eslint` task + some small fiddling to get it save changes over old files.

mleanos:
Would you like to share you Gulp task? I thought about doing the same yesterday :)

simison:
@mleanos sure :-)

```js
gulp.task('eslint-and-fix', function () {

  var gulpIf = require('gulp-if');
  var fs = require('fs');

  function saveFixedFile(file) {
    // Has ESLint fixed the file contents?
    if (file.eslint != null && file.eslint.fixed) {
      // Write file over with new fixed contents
      fs.writeFile(file.eslint.filePath, file.eslint.output, function(err) {
        if (err) {
          return console.log(err);
        }
      });
    }
  }

  var assets = _.union(
    defaultAssets.server.gulpConfig,
    defaultAssets.server.allJS,
    defaultAssets.client.js,
    testAssets.tests.server,
    testAssets.tests.client,
    testAssets.tests.e2e
  );

  return gulp.src(assets)
    .pipe(plugins.eslint({
      fix: true
    }))
    .pipe(plugins.eslint.format())
    // if fixed, write the file to dest
    .pipe(gulpIf(saveFixedFile, gulp.dest('./data/fixed')));
});
```

It's uggly, as I didn't figure out why `gulp.dest` didn't work (it's just spat files all over).

Here's the original https://github.com/adametry/gulp-eslint/blob/master/example/fix.js

Issue #1846, opened 2017-08-14
Update eslint


simison:
I feel that the best way to start moving towards v1.0 (#1666) is to first update eslint to latest.

- [ ] update to eslint 3.x,
- [ ] get rid of `eslint-config-airbnb`, 
- [ ] set CI builds fail on linting errors

lirantal:
isn't `eslint-config-airbnb` used as part of applying the rules through eslint?

simison:
@lirantal True, but it's geared towards React native and [it's deprecated](https://www.npmjs.com/package/eslint-config-airbnb#eslint-config-airbnblegacy), too.

simison:
Actually if we want a predefined set of eslint rules, we should consider [XO](https://www.npmjs.com/package/xo).

lirantal:
XO is very opinionated by a specific set of people. If we're at a cross-road to diverge how about [standardjs](https://standardjs.com) ?

Issue #1674, opened 2016-12-06
npm install failed


luebken:
I've just did a `npm install`on a fresh copy of master and got the following.

```
...
npm WARN eslint-config-airbnb@6.0.2 requires a peer of eslint-plugin-react@^4.0.0 but none was installed.
npm ERR! Darwin 15.3.0
npm ERR! argv "/usr/local/Cellar/node/7.2.0/bin/node" "/usr/local/bin/npm" "install"
npm ERR! node v7.2.0
npm ERR! npm  v3.10.9
npm ERR! code ELIFECYCLE

npm ERR! v8-debug@0.7.7 install: `node-pre-gyp install --fallback-to-build`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the v8-debug@0.7.7 install script 'node-pre-gyp install --fallback-to-build'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the v8-debug package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-pre-gyp install --fallback-to-build
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs v8-debug
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls v8-debug
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/mdl/workspace/redhat/examples/meanjs-app/npm-debug.log
```

```
$ node --version
v7.2.0
```

luebken:
I've fixed this by following the instructions from https://www.npmjs.com/package/eslint-config-airbnb

```
(
  export PKG=eslint-config-airbnb;
  npm info "$PKG@latest" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG@latest"
)
```

But to be honest I don't know what happened there. Is that something that can be 
a) automated?
or
b) documented?

lirantal:
@luebken can you explain more about your fix, I don't follow why/where we would need to do that.
You're welcome to propose a PR too.

Jageen:
So @luebken not decided to share solution ?, I am still having issue @lirantal do you have any information regarding this.

luebken:
[...]

PR #1847, opened 2017-08-14
feat(build): Update dependencies

### LINKED ISSUE #1569: Discussion - package.json cleanup
**Discussion:**
The below dependencies are listed but not used

```
file-stream-rotator (1)
jasmine-core (0)
mocha (33)
phantomjs-prebuilt (102)
```

The below dev-dependencies are listed but not used

```
eslint (127)
eslint-config-airbnb (1)
gulp-node-inspector (282)
gulp-util (94)
istanbul (59)
lcov-result-merger (93)
node-inspector (235)
```

Also `connect-flash (0)` is being used in the express.js file but what is using this flash? I do not see it being used anywhere other than the middleware initialized.

The numbers in brackets are a node count of the total recursive dependencies for those modules.

Am I missing something. Or Is there a reason why they cant be removed?

simison:
Remove dependencies which are already installed via other dependencies:
- eslint (gulp-eslint)
- istanbul (gulp-eslint)
- mocha (gulp-mocha)

Remove unused dependency:
- gulp-util

Remove blackbox `eslint-config-airbnb`. Related #1846, #1674 — thoughts?

Move installing `lcov-result-merger` to Travis config since it's the only place where it's ever used.

Update a bunch of outdated dependencies.

Closes #1569


lirantal:
if they're already installed through others then it's ok for now, we just need to keep an eye before they change it to peerDependencies and then we'll need to install them ourselves.

About airbnb I left a comment on #1846

simison:
>if they're already installed through others then it's ok for now, we just need to keep an eye before they change it to peerDependencies and then we'll need to install them ourselves.

Yep; our Travis builds should fail at that point (unless `node_modules` is cached there). I tested installation with NPM 5.2.0

Actually I'll drop eslint-airbnb change off from this PR, it's a separate issue to work on.

simison:
Done here.

BTW worth noting that I didn't update Mocha here yet, as it will require changes to test files as well.

lirantal:
alright.

PR #1851, opened 2017-08-16
feat(core): Add manifest.json

pgrodrigues:
This PR adds a JSON file which provides information about the app,
offering a quicker access and a richer experience for the users.

Not sure if this should be added right away since it is still in development,
however since it seems to be considered a best practice to have one, here it is.

Eitherway I kept the app data in the manifest as generic as possible.

More info on this: [Mozilla](https://developer.mozilla.org/en-US/docs/Web/Manifest) / [W3C](https://www.w3.org/TR/appmanifest/)


simison:
LGTM!

Just left one comment regarding icons.

pgrodrigues:
Ok @simison I can add that one to `/modules/core/client/img/brand`.

So basically it should use the .ico for 48x48 and the png for the other sizes?

I don't think it would make sense having 1 image file for each icon size if MEAN.js is just a starting point for an app. Maybe we could have that same 128px png to be used in every size (48x48, 72x72, 96x96, 128x128 and 256x256)?

simison:
@pgrodrigues sorry late reply. I recon just configuring one image size (128px) is enough, and people can then add the rest to manifest if they want.

pgrodrigues:
@simison LGTY?

Not sure if I messed up with git squash.

simison:
@pgrodrigues seems like your merge removed the actual `manifest.json` file, here it is again:

```json
{
  "name": "meanjs",
  "short_name": "meanjs",
  "start_url": ".",
  "display": "browser",
  "background_color": "#fff",
  "theme_color": "#000",
  "description": "Fullstack JavaScript with MongoDB, Express, AngularJS, and Node.js.",
  "icons": [{
    "src": "/modules/core/client/img/brand/icon.png",
    "sizes": "128x128",
    "type": "image/png"
  }]
}
```

After adding the file back, LGTM!

simison:
Oh and don't worry about squashing commits, I can do it while merging.

pgrodrigues:
@simison you're right. Sorry for the trouble.

simison:
No worries! Thanks!

Issue #1852, opened 2017-08-18
Deploy To XXX button


Ghalleb:
Hi,

Are you planning to put back the "one click deployment" button like there was in 0.4.X version? It was for Bluemix and there was a tutorial for Pivotal Web Service (https://github.com/meanjs/mean/blob/0.4.x/manifest.yml).
I know it was because Cloud Foundry was evolving too fast. But some platform are now more stable.

It could be a "deploy to heroku" button (https://devcenter.heroku.com/articles/heroku-button) which is simple or perhaps we could use Google Cloud Launcher (https://cloud.google.com/launcher/docs/).

If you didn't plan it but still want it, I can manage to do it myself.

Have a good day.

simison:
I don't think nobody has anything against such buttons but it really boils down to maintaining them actively.

https://github.com/meanjs/mean/pull/1476

If nobody from the core team is using those services, it's hard to follow their changes and no deploy button is better than broken deploy button.

@Ghalleb I'd say go for it if you _really_ can commit to long term support for such feature; follow their blogs/changelogs and update changes to here within reasonable timeframe.

Ghalleb:
@simison 
I have no problem commiting on that as I'm plannig to use Mean.js for my future projects.
First let me decide between Heroku and GCP. 
Heroku could have a "deploy to" button but GCP is a more packaged solution that need to be set on their page.
If you have a preference, feel free to share it...

simison:
@lirantal @mleanos thoughts?

lirantal:
Nothing against it if we can maintain it.

Ghalleb:
I did the deploy to Heroku button.

I had to trick a little bit the script because some dependencies needed to deploy in prod are in the devDependencies only when Heroku install only prod dependencies.
And I really wanted to deploy a production version of Mean.js.

I will take a look at the Google Cloud Platform soon.

Have a good day.

simison:
>I had to trick a little bit the script because some dependencies needed to deploy in prod are in the devDependencies 

@lirantal @mleanos any reason to keep those dependencies at `devDependencies` if they're needed to prod-build?

Ghalleb:
I had an config var: "NPM_CONFIG_PRODUCTION": "false" so the npm install will install all dependencies instead of production ones....

simison:
@Ghalleb I think it's better to move those dependencies to right place. Wanna PR since you're at it? :-)

Ghalleb:
[...]

PR #1854, opened 2017-08-21
Deploy to Heroku button

Ghalleb:
Add a "Deploy to Heroku" button on README.MD


app.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "meanjs",
+  "description": "Full-Stack JavaScript with MongoDB, Express, AngularJS, and Node.js",
+  "repository": "https://github.com/Ghalleb/mean.git",
```
simison: This was just temporarily pointed here for testing?

app.json
```diff
@@ -0,0 +1,28 @@
...
+  "logo": "http://meanjs.org/img/logo-small.png",
+  "keywords": [
+    "mongodb",
+    "express",
+    "angularjs",
+    "node.js",
+    "mongoose",
+    "passport"],
```
simison: Add "mean" ? :-)

simison:
(Issue https://github.com/meanjs/mean/issues/1852)

app.json
Ghalleb: I took the ones frome here:
https://github.com/meanjs/mean/blob/master/config/env/default.js

app.json
Ghalleb: Good catch....

simison:
via https://github.com/meanjs/mean/issues/1852#issuecomment-323652227:
>I had to trick a little bit the script because some dependencies needed to deploy in prod are in the devDependencies

Is this trick in these config files or am I missing something?

simison:
LGTM. With #1855 in, we can merge, @mleanos any thoughts on this?

Procfile
```diff
@@ -1 +1 @@
-web: npm start
+web: npm run start:prod
```
mleanos: This change would restrict other deployments, that use a `Procfile`, to running in "production" mode. The NODE_ENV configuration should be responsible for determining the mode the application is running in. There can be times that other environment configurations would want to be used; development, test, stage, qa, etc.

I understand that by clicking the "Deploy to Heroku" button for the first time, would run the app with the default development configuration. However, if an app by the name specified in `app.json` already exists, then the NODE_ENV can be set.

Ghalleb: In fact, if you change the npm script to use the NODE_ENV, even the first time you deploy it will be ok, because I set the NODE_ENV to production.
I will revert the proc file but let you change the script.

mleanos: Is this the script you're referring to? https://github.com/meanjs/mean/blob/master/gulpfile.js#L484

I plan to remove the `env:dev` task from the sequence.

Ghalleb: You said: "If I remove the 'env:dev' from the default Gulp task, the application runs in whatever NODE_ENV I set in my environment variables." 
So I think that was it.

mleanos:
@Ghalleb It looks like there's a conflict in the Readme file now. While you're squashing, can you rebase from the current *master* branch & resolve the conflict?

Ghalleb:
I merge the branches.
For the squashing, I don't really know how to do it but I will try to learn for next time if you don't mind.

mleanos:
@Ghalleb Don't worry about the squashing then. We can do it when we merge. However, make sure the latest commit has a message header & body, as described in the contributing guidelines.

PR #1855, opened 2017-08-21
fix(build): Require correct dependencies for prod build

simison:
Some of the dependencies required by `npm run start:prod` aren't currently installed with `npm install --production`.

See #1852 for reference.

This also moves some of the `require`'s in Gulpfile inside tasks so that they wouldn't be required if unrelated tasks are run (i.e. `start:prod` task).

Also changed several semver ranges from `^` to more strict `~`.


simison:
To test this, do:

```bash
git clone https://github.com/simison/mean.git mean-simison
cd mean-simison
git checkout restructure-prod-build
npm install --production
NODE_ENV=production npm run start:prod
```

simison:
I'd be especially interested to have @mleanos look at those protractor changes I made!

mleanos:
I ran into an issue with our [default Gulp task](https://github.com/meanjs/mean/blob/master/gulpfile.js#L481). It actually sets the NODE_ENV to "development". And the [Procfile](https://github.com/meanjs/mean/blob/master/Procfile) starts the app using `gulp default`. This will cause the application to run in Development mode regardless of what NODE_ENV is set to. In my case, I want my Heroku deployment to run in "production" mode.

If I remove the `'env:dev'` from the default Gulp task, the application runs in whatever NODE_ENV I set in my environment variables.

@simison @Ghalleb Do you have thoughts on this?

simison:
@mleanos I'm not really familiar with Heroku. What would you recommend?

simison:
@mleanos On second thought:

- `npm start` should probably just use existing `NODE_ENV` and default to `development`. In package.json scripts we could have:
```bash
'start': 'if [ $NODE_ENV == "production" ]; then npm run start:prod; else npm run start:dev; fi'
```
... but that test could be also at Gulpfile.js — might be even more clear to have it there? I'm not exactly sure how well bash conditionals work e.g. in Windows machines.
- `npm run start:prod` and `npm run start:dev` stick to using `'env:' Gulp tasks.

Ghalleb:
If you look at my PR #1854, I changed `npm start` to `npm run start:prod` in the procfile as this file is tipycally used for production only...

simison:
@Ghalleb thanks! So I'll merge this and that change can be done in your PR.

PR #1857, opened 2017-08-27
Save profile images to Amazon S3

Ghalleb:
This PR aimed to save profile images to a S3 Bucket.

I discovered that deploying to Heroku leaded to profile images not being saved.

This 3 environment parameters need to be filled:
S3_PROFILE_BUCKET: the name of the bucket where the images will be saved
S3_ACCESS_KEY_ID: Your S3 access key
S3_SECRET_ACCESS_KEY: Your S3 access key password

### DIFF
--- a/modules/users/client/views/settings/change-profile-picture.client.view.html
+++ b/modules/users/client/views/settings/change-profile-picture.client.view.html
@@ -3,7 +3,7 @@
     <form class="signin form-horizontal">
       <fieldset>
         <div class="form-group text-center">
-          <img ngf-src="vm.fileSelected ? picFile : '/' + vm.user.profileImageURL" alt="{{vm.user.displayName}}" class="img-thumbnail user-profile-picture" ngf-drop>
+          <img ngf-src="vm.fileSelected ? picFile : vm.user.profileImageURL" alt="{{vm.user.displayName}}" class="img-thumbnail user-profile-picture" ngf-drop>
         </div>
         <div ng-show="vm.loading" class="form-group text-center">
           <img ng-src="/modules/core/client/img/loaders/loader.gif" height="50" width="50" alt="Loading image...">
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -62,10 +62,10 @@ exports.changeProfilePicture = function (req, res) {
   var multerConfig;
 
 
-  if (config.uploads.profile.s3Bucket) {
+  if (config.uploads.storage === 's3' && config.aws.s3) {
     aws.config.update({
-      accessKeyId: config.s3Config.accessKeyId,
-      secretAccessKey: config.s3Config.secretAccessKey
+      accessKeyId: config.aws.s3.accessKeyId,
+      secretAccessKey: config.aws.s3.secretAccessKey
     });
 
     var s3 = new aws.S3();
@@ -73,7 +73,7 @@ exports.changeProfilePicture = function (req, res) {
     multerConfig = {
       storage: multerS3({
         s3: s3,
-        bucket: config.uploads.profile.s3Bucket,
+        bucket: config.aws.s3.bucket,
         acl: 'public-read'
       })
     };
@@ -118,7 +118,7 @@ exports.changeProfilePicture = function (req, res) {
 
   function updateUser() {
     return new Promise(function (resolve, reject) {
-      user.profileImageURL = config.uploads.profile.s3Bucket ?
+      user.profileImageURL = config.uploads.storage === 's3' && config.aws.s3 ?
         req.file.location :
         '/' + req.file.path;
       user.save(function (err, theuser) {
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -11,6 +11,7 @@ var _ = require('lodash'),
   multer = require('multer'),
   multerS3 = require('multer-s3'),
   aws = require('aws-sdk'),
+  amazonS3URI = require('amazon-s3-uri'),
   config = require(path.resolve('./config/config')),
   User = mongoose.model('User'),
   validator = require('validator');
@@ -134,24 +135,54 @@ exports.changeProfilePicture = function (req, res) {
   function deleteOldImage() {
     return new Promise(function (resolve, reject) {
       if (existingImageUrl !== User.schema.path('profileImageURL').defaultValue) {
-        fs.unlink(existingImageUrl, function (unlinkError) {
-          if (unlinkError) {
+        if (config.uploads.storage === 's3' && config.aws.s3) {
+          try {
+            var { region, bucket, key } = amazonS3URI(existingImageUrl);
+          } catch(err) {
+            console.warn(`${existingImageUrl} is not a valid S3 uri`);
 
-            // If file didn't exist, no need to reject promise
-            if (unlinkError.code === 'ENOENT') {
-              console.log('Removing profile image failed because file did not exist.');
-              return resolve();
+            return resolve();
+          }
+
+          aws.config.update({
+            accessKeyId: config.aws.s3.accessKeyId,
+            secretAccessKey: config.aws.s3.secretAccessKey
+          });
+
+          var s3 = new aws.S3();
+
+          var params = {
+            Bucket: config.aws.s3.bucket,
+            Key: key
+          };
+          s3.deleteObject(params, function (err) {
+            if (err) {
+              console.log('Error occurred while deleting old profile picture.');
+              console.log("Check if you have sufficient permissions : " + err);
             }
 
-            console.error(unlinkError);
-
-            reject({
-              message: 'Error occurred while deleting old profile picture'
-            });
-          } else {
             resolve();
-          }
-        });
+          });
+        } else {
+          fs.unlink(path.resolve('.' + existingImageUrl), function (unlinkError) {
+            if (unlinkError) {
+
+              // If file didn't exist, no need to reject promise
+              if (unlinkError.code === 'ENOENT') {
+                console.log('Removing profile image failed because file did not exist.');
+                return resolve();
+              }
+
+              console.error(unlinkError);
+
+              reject({
+                message: 'Error occurred while deleting old profile picture'
+              });
+            } else {
+              resolve();
+            }
+          });
+        }
       } else {
         resolve();
       }
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -9,6 +9,8 @@ var _ = require('lodash'),
   errorHandler = require(path.resolve('./modules/core/server/controllers/errors.server.controller')),
   mongoose = require('mongoose'),
   multer = require('multer'),
+  multerS3 = require('multer-s3'),
+  aws = require('aws-sdk'),
   config = require(path.resolve('./config/config')),
   User = mongoose.model('User'),
   validator = require('validator');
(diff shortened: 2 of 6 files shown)


mleanos:
Even though I used S3 with a few of my MEAN.js project, I think these changes might be too opinionated for this framework. There are many other storage services that can be used, and we can't include logic for each in the boilerplate implementation.

This could be a useful feature branch that can be an example of how to handle S3 storage with MEAN.js though. The implementation looks correct though, but I haven't tested.

Ghalleb:
I did this because otherwise deploying the boilerplate to heroku would not really work. The profile images are not saved. 
I also saw in the issues that other people had the same problem.
So it could be a good idea to add it to the master branch. If you don't specify the S3_PROFILE_BUCKET variable it will work as usual. In fact it's just a simple change in multer storage config.
I chose S3 because I know it but I cam add others if it's needed.
I demonstrate that mean.js can work in a multi node architecture and scale without problem.

Ghalleb:
If i can go further, when I first started with mean.js, with version 0.4.2, I found usefull that I could deploy to Cloud Foundry (Bluemix, Pivotal Web Services, ...) so easily. So I could focus on the code, not on devopts.
That's why I added the deploy to heroku button and the ability to use S3 (Heroku use AWS too...). With that you can deploy a production class app without losing time. And focus on important things.

simison:
I agree with @mleanos 's sentiment about adding S3 being too opinionated, but I think the need for easy deployment to get started outweighs that. Just for the sake of staying pragmatic.

So either we can:
- just go with this PR, 
- add note+tips on how to store files remotely (I think it's a common need also outside Heroku use)
- or merge this PR and add note/instructions/list for other possible cloud services

@lirantal thoughts?

simison:
Oh hey and thanks for coding & activism @Ghalleb ! It's awesome. :-)

lirantal:
I agree that we don't want to steer off with too many examples, but on the other hand MEAN.JS serves as a foundation for learning and flexible project boilerplate to start from, as such and with the AWS-franzy I don't think it's far fetched to include this, especially when it's anyway only enabled with proper config.

So I'm ok with this.

mleanos:
I'm fine with these changes too then. I just didn't want to jump down a rabbit hole by convoluting the code-base with too many non-boilerplate use cases. However, AWS is very common & since these changes only take effect when the S3 configuration is present, then it's ok. The implementation is clean enough as well, since it's only a few changes (mostly on the backend).

I'll do a thorough code-review and test this locally. Thanks @Ghalleb. I appreciate the willingness to jump right in here and take the initiative.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -95,7 +118,9 @@ exports.changeProfilePicture = function (req, res) {
 
   function updateUser() {
     return new Promise(function (resolve, reject) {
-      user.profileImageURL = config.uploads.profile.image.dest + req.file.filename;
+      user.profileImageURL = config.uploads.profile.s3Bucket ?
+        req.file.location :
+        '/' + req.file.path;
```
mleanos: This change is preventing the `deleteOldFile` from finding the existing file every time. I understand why this change was made. Otherwise, the front-end code wouldn't be able to find the location of the image. We might need to find a better way to handle this.

I'm on Windows, so it could be only happening in my environment. @Ghalleb Can you confirm if this is happening in your environment as well (if you're not on Windows, of course :) )

After adding a profile image, my user.profileImageUrl is `/modules\users\client\img\profile\uploads\12db59fbdd2193b3b8f99867e1640175`.

mleanos: To fix this issue, you can update this line https://github.com/meanjs/mean/blob/master/modules/users/server/controllers/users/users.profile.server.controller.js#L112 to resolve the path for the `unlink` method.

`fs.unlink(path.resolve('.' + existingImageUrl), function (unlinkError) {`

But it seems a little hacky to me.

Ghalleb: I'm not sure I understand what to do for this one.
On my environment (Ubuntu), everything is working fine.

For all the other changes, I will make them soon.

Thanks for the review.

mleanos: Can you verify that the existing file is getting deleted, and you're not getting an error logged to the console?

Ghalleb: I will...

Ghalleb: You are right, I have an error in the server logs:

`Removing profile image failed because file did not exist.`
`info: POST /api/users/picture 200 24.550 ms - 414`

Ghalleb:
@mleanos 
I made the changes you requested exept for the profile image path.
Should I go with your hacky trick or do you have a better solution in mind?

mleanos:
I haven't thought of any better solution to the profile image oath issue. See if you could come up with something.

As a first step in solving the issue, how about adding a test for this case? I'm orettybsure there isn't a test for it.

Test case:
1. Add profile image
2. Add new profile image
3. Make sure the first profile image was removed

Ghalleb:
I added your fix for file deletion and also added S3 file deletion.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -109,24 +135,54 @@ exports.changeProfilePicture = function (req, res) {
   function deleteOldImage() {
     return new Promise(function (resolve, reject) {
       if (existingImageUrl !== User.schema.path('profileImageURL').defaultValue) {
-        fs.unlink(existingImageUrl, function (unlinkError) {
-          if (unlinkError) {
+        if (config.uploads.storage === 's3' && config.aws.s3) {
+          try {
+            var { region, bucket, key } = amazonS3URI(existingImageUrl);
```
mleanos: This could be troublesome, if switching from "local" to "s3" storage. The first time a user updates their profile picture after switching the storage from one to another, this will create inconsistency with the data/files. This is probably more of an edge case, or migration consideration. Perhaps not a big deal?

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -109,24 +135,54 @@ exports.changeProfilePicture = function (req, res) {
   function deleteOldImage() {
     return new Promise(function (resolve, reject) {
       if (existingImageUrl !== User.schema.path('profileImageURL').defaultValue) {
-        fs.unlink(existingImageUrl, function (unlinkError) {
-          if (unlinkError) {
+        if (config.uploads.storage === 's3' && config.aws.s3) {
```
mleanos: Since this check is being used in multiple places, you could move it into a local variable at the top level of this request handler.

modules/users/client/views/settings/change-profile-picture.client.view.html
```diff
@@ -3,7 +3,7 @@
     <form class="signin form-horizontal">
       <fieldset>
         <div class="form-group text-center">
-          <img ngf-src="vm.fileSelected ? picFile : '/' + vm.user.profileImageURL" alt="{{vm.user.displayName}}" class="img-thumbnail user-profile-picture" ngf-drop>
+          <img ngf-src="vm.fileSelected ? picFile : vm.user.profileImageURL" alt="{{vm.user.displayName}}" class="img-thumbnail user-profile-picture" ngf-drop>
```
mleanos: You can add a config setting to the [Shared config](https://github.com/meanjs/mean/blob/master/config/env/default.js#L58) that would allow you to check whether the profile should be retrieved using local storage.

Then you wouldn't need remove the "/", thus solving path issues on the backend.

```javascript
shared: {
    profileImageIsExternal: process.env.UPLOADS_STORAGE ? true : false
    ...
}
```

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -109,24 +135,54 @@ exports.changeProfilePicture = function (req, res) {
...
-            reject({
-              message: 'Error occurred while deleting old profile picture'
-            });
-          } else {
             resolve();
-          }
-        });
+          });
+        } else {
+          fs.unlink(path.resolve('.' + existingImageUrl), function (unlinkError) {
```
mleanos: I'm not really a fan of this solution. See my comment about adding a setting to the Shared config. This might not be needed if we can keep the "/" on the front-end when we're using local storage.

package.json
```diff
@@ -38,7 +38,9 @@
   },
   "dependencies": {
     "acl": "~0.4.10",
+    "amazon-s3-uri": "0.0.3",
```
mleanos: This library is rather new, and doesn't have many stars. However, the implementation looks fine and isn't too complicated. We'll probably have to keep an eye on this library though.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -109,24 +135,54 @@ exports.changeProfilePicture = function (req, res) {
...
-            // If file didn't exist, no need to reject promise
-            if (unlinkError.code === 'ENOENT') {
-              console.log('Removing profile image failed because file did not exist.');
-              return resolve();
-            }
+            return resolve();
+          }
 
-            console.error(unlinkError);
+          aws.config.update({
```
mleanos: Is this needed? This is already happening up top.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -95,7 +119,9 @@ exports.changeProfilePicture = function (req, res) {
 
   function updateUser() {
     return new Promise(function (resolve, reject) {
-      user.profileImageURL = config.uploads.profile.image.dest + req.file.filename;
+      user.profileImageURL = config.uploads.storage === 's3' && config.aws.s3 ?
+        req.file.location :
+        '/' + req.file.path;
```
mleanos: This can be changed back to what it was before: `config.uploads.profile.image.dest + req.file.filename`, if we can still use the "/" on the front-end when using local storage.

modules/users/server/controllers/users/users.profile.server.controller.js
mleanos: Maybe something like: `var useLocalStorage = config.uploads.storage === 'local';`

Then here you can just ` if (!useLocalStorage) `

Ghalleb: I wanted to let place for another storage provider... 
But if it will only be local or S3, I'll do it.

modules/users/server/controllers/users/users.profile.server.controller.js
Ghalleb: I thought about that as well. Normally you have to configure your app once!
An edge case for me.

modules/users/server/controllers/users/users.profile.server.controller.js
Ghalleb: I'm not fan of the '/' in the front end. I prefer having the final URL in the database. Otherwise if we switch from local to s3, all old URL would be bad formed. And that add more logic to the front where it shouldn't be.

package.json
Isabeljason: Even we follow the library updates there are few minor changes that need to be update to it. As i am a new bee to this need support from others to get into and complicated issues like path, switching, controllers, pre & post updates need to be checked very carefully.
Update me on this if any wrong.

Ghalleb:
I made some changes. But not the '/' in the front-end as I prefer nor addind logic on the fron-end and storing the final url in the database is better in my opinion

mleanos:
@Ghalleb Thank you for your updates based on my feedback!

LGTM now.. I'll merge in the next couple of days, after I test locally once more.

mleanos:
@Ghalleb Thank you for contribution. I apologize it took so long to merge

I didn't catch that your commits didn't have messages, before now. Take a look at our [contribution guidelines](https://github.com/meanjs/mean/blob/master/CONTRIBUTING.md#commit-message-guidelines) for next time. Looking forward to seeing more contributions from you :)

Ghalleb:
Are you sure there is no messages?
I see them in my repo:
https://github.com/Ghalleb/mean/commits/s3-profile-image

nlaycock:
I just wanted to say a massive thank you for all the work on S3. I have just implemented it myself using your approach and it has saved me a HUGE amount of time. Thank you soooooooo much!!

PR #1858, opened 2017-08-27
fix(eslint): Make `space-before-function-paren` rule consistent with other rules

simison:
Fixes eslint errors in eslint 4.x:
```
Linter] Error running ESLint Error: /mean/.eslintrc.js:
	Configuration for rule "space-before-function-paren" is invalid:
	Severity should be one of the following: 0 = off, 1 = warning, 2 = error (you passed "error").
```

We're still in eslint 2.x. so apparently no problems as of yet, but will have to be fixed before updating eslint (#1846).

Related to previous PR #1844 and already closed issue #1605

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,7 +40,7 @@ module.exports = {
     'one-var': [0, 'never'],
     'one-var-declaration-per-line': [2, 'always'],
     'padded-blocks': 0,
-    'space-before-function-paren': ['error', {
+    'space-before-function-paren': [2, {
       'anonymous': 'always',
       'named': 'never',
       'asyncArrow': 'always'


simison:
Hmm, actually now that I double checked docs, it was correct: https://eslint.org/docs/rules/space-before-function-paren

Something else is going on here. So I'm experiencing this with Atom editor and I assumed it's using locally installed eslint config of the project, but that might not be the case.

I started seeing shortly after updagint to eslint 4.x. in my project, but I just realised I also updated Atom.

After changing `error` to `2`, I'm now seeing this in my Atom:
```
[Linter] Error running ESLint Error: /mean/.eslintrc.js:
	Configuration for rule "space-before-function-paren" is invalid:
	Value "[object Object]" no (or more than one) schemas match.
```

mleanos:
Any update here? I think I've noticed that when upgrading ESLint as well. So I doubt it's an issue with Atom.

simison:
@mleanos I had to update Atom, which somehow was stuck with updating itself, so I had to remove it and re-install.

Now it works fine, but I would still change that `error` to `2` just for consistency, or change everything else to string codes as well (they seem to use only strings now at eslint docs).

mleanos:
Ok. Let's just stick with using `2` for now. Then in a separate PR, we can update the settings to use the String parameter.

PR #1860, opened 2017-08-30
fix(gulpfile): show error on uglify

PierreBrisorgueil:
uglify will crash with ES6 foreach for example, we will display where will be the error with this modification


lirantal:
looks like it makes sense just to track errors?

PierreBrisorgueil:
yes, just for this :) actually the build can failed, but you don't know where is the error

lirantal:
Cool, thanks.

PR #1887, opened 2017-09-26
feat(deps): dependencies upgrade

lirantal:
Supercedes #1886


simison:
BTW really minor thing now that package-lock is in place, but a bunch of `~` changed to `^`.

lirantal:
👍 
yeah I had to re-do the package-lock thing so probably dude to that.

PR #1933, opened 2017-12-04
Manage articles - no articles alert missing.

### LINKED ISSUE #1930: Message about 'no articles yet' not showing up.
In the Admin - 'manage articles' view, if there are no articles yet there's supposed to be an alert

> No articles yet, why don't you create one?

However it's not showing up.

I believe in line 23 instead of: 

> articles.$resolved

It should be: 

> vm.articles.$resolved

AKA-AK:
fix(articles) Alert missing.

In the Admin - 'manage articles' view, if there are no articles yet there's supposed to be an alert.
However, it doesn't show up.

Fixes #1930


lirantal:
Thanks!

PR #1944, opened 2018-02-04
Manage Article - Updating the warning message condition

### LINKED ISSUE #1930: Message about 'no articles yet' not showing up.

tejas002:
fix(articles): Updated warning message condition

In the Admin - 'manage articles' view, the warning message will be displayed only when there is no article present.

Fixes #1930 ,Fixes #1933


lirantal:
thanks @tejas002!

Issue #1909, opened 2017-11-01
Moving away from Bower and switching to Yarn


imranzaidi:
Not sure if you guys have read this yet, but Bower is on it's way out:

https://community.plone.org/t/bower-is-now-officially-dead/4246

Bower has a page on migrating to yarn, I'm guessing it's for good reason. Is a migration already in the works?

https://bower.io/blog/2017/how-to-migrate-away-from-bower/

PierreBrisorgueil:
agree with this, feel free to make a pr or I  will have a look on this as soon as possible

Issue #1087, opened 2015-12-03
Use NPM+Browserify/Webpack and remove Bower?


simison:
(Topic was: "Move Bower modules out from /public/")

It's a slight security issue.

Now Bower libs end up to `/public/lib/**/*` and thus are open to public from the server.

I'd like to be in full control what my server is publishing. This way we might accidentally publish malicious JS _at the same domain_, so we aren't protected from XSS attacks via this.

Related to #1078

codydaig:
@simison What's your proposed fix while still using bower?

simison:
I don't have anything I can suggest as a best practise, but I guess it boils down into:
- symlinking (very error prone)
- bundling needed files into one js also at development (already done at production)
- keeping them under public only for dev and cleaning the folder out for prod
- using `npm main-bower-files`

simison:
...or another way is to push non-Bower way such as NPM+Browserify, but that's another discussion.

lirantal:
a lot of frameworks make `grunt build` create a really production-ready snapshot of the project, such as uglyfing, minifying, etc, but most importantly it creates a `dist/` directory such as:
```
dist/
dist/public
dist/server
package.json
```

which makes this `dist/` directory deployable: you can just copy that to a remote server, run npm install for the modules, and start the server.

what do you think?

lirantal:
@simison if that sounds like a good idea, how about taking this for a PR?

codydaig:
@lirantal @simison :+1:

trainerbill:
I use wiredep in one of my projects and love it.

https://github.com/taptapship/wiredep

Automatically reads the bower file and inserts into the layout page.  So you would just have to move the bower directory and put 2 pieces of code onto the layout.html page.

trainerbill:
@lirantal @codydaig @simison I think this would be pretty easy.  Move the Bower directory out of public, wiredep(), move the files to pubilc/lib, use gulp-inject to layout.html.  I wouldn't want to do this for grunt as well so when is the official move to Gulp?  Doing this would bring breaking changes to using grunt as we would be removing all of the bower packages out of the assets files as well.

codydaig:
@trainerbill I just submitted a PR for the removal of grunt to start discussion around what all needs to be done to officially remove grunt. I think the plan is to complete the migration before 0.5.0. 

I'm indifferent on how we go about this. @simison Are you interested in submitting the PR as discussed earlier?

lirantal:
[...]

PR #1950, opened 2018-03-03
feat(yarn) bower to yarn, related to #1909

PierreBrisorgueil:
Begin switch from bower to yarn, related to #1909
Switch done with the [bower-away](https://github.com/sheerun/bower-away)

(& clean some scripts in package.json, maybe some mistake to revert)


lirantal:
this is really great @PierreBrisorgueil!  🎉 
would you like to finalize the remaining work? (I wasn't sure about what you refer to in the 2nd bullet)

PierreBrisorgueil:
thx, but I deleted my bullets, I was missing a moment on the subject

I think i switch to fast to yarn to test it .. 
https://medium.com/netscape/npm-5-yarn-killer-ba69737b24d0

and I just read that : https://github.com/meanjs/mean/issues/1087

we need to choose between  ( in order to keep public/dist) :

- yarn & symylink between node_module/public & publis/dist 
- or npm Webpack
- or npm Browserify

it think npm Is better, because more used. 

for Webpack or Browserify http://blog.namangoel.com/browserify-vs-webpack-js-drama 

maybe the better choice is Browserify because we are on a simple stack without angular x

lirantal:
In terms of yarn vs npm I'm using yarn everywhere personally but npm is good too.
I'd go with the npm+webpack

PierreBrisorgueil:
ok, thx, i will quickly end this PR, and see after if i have time to test npm+webpack

ok for you ? 


But for the travis log https://travis-ci.org/meanjs/mean/jobs/348762357
it seems like we have another issue relative to https://stackoverflow.com/questions/39541739/chromedriver-error-chrome-version-must-be-52-using-nightwatch

PierreBrisorgueil:
closed to open a new clean PR.

PR #1952, opened 2018-03-13
fix(travis) : chrome install 

PierreBrisorgueil:
Fix this issue : Chromedriver error “Chrome version must be >= 63..”

PR #1953, opened 2018-03-13
feat(yarn) bower to yarn, related to #1909

PierreBrisorgueil:
Begin switch from bower to yarn, related to #1909
Switch done with the bower-away

(& clean some scripts in package.json, maybe some mistake to revert)

In futur we will be able to switch from yarn to npm+webpack, this is done for short term (like discussed [here](https://github.com/meanjs/mean/pull/1950)) 

Build pass with https://github.com/meanjs/mean/pull/1952

Now you can add client package with yarn add public@~3.3.10


lirantal:
@PierreBrisorgueil you'd have to rebase your changes on master to get the recent updates on the chrome fix so we can test this.

PierreBrisorgueil:
I have a fix to add on this, delete wiredep. it is only called in build prod, that's why the error is not raised before.

README.md
```diff
@@ -65,7 +61,7 @@ $ npm install
 This command does a few things:
 * First it will install the dependencies needed for the application to run.
 * If you're running in a development environment, it will then also install development dependencies needed for testing and running your application.
-* When the npm packages install process is over, npm will initiate a bower install command to install all the front-end modules needed for the application
+* When the npm packages install process is over, npm will initiate a yarn install command to install all the front-end modules needed for the application
```
mleanos: Can you elaborate here, on what the `postinstall` actually does? Some users won't understand how the client-side packages are linked to `public/lib`.

PierreBrisorgueil:
@mleanos @lirantal yep, I tested Browserify and webpack, stay on npm/yarn + gulp and wepback ( in order to introduce to webpack migration).

But with both, Browserify, and Webpack, everything is ok except for owaspPasswordStrengthTest.

I can't find how to get access to this var in front. it's the last problem. I have a look on this actually, the best solution might be to rework this lib quickly. I can push my version with webpack if you want to test quickly.

README.md
PierreBrisorgueil: done in new pr

PR #1954, opened 2018-03-13
feat(ES6-7): switch from Es5 to es6-7, add babel for production, delete uglify

PierreBrisorgueil:
begin switch from es5 to 6-7 with lebab https://lebab.io , on server side only. 

I can add this on client side with gulp-babel to transpil, i will have a look on this. 

like other PR we need to end fix travis to check build, but in local everything seems to be ok


lirantal:
Seems like we need to rebase this one as well!

PierreBrisorgueil:
done, failed due to bower, I think we have nothing to fix if we merge yarn PR :)

PierreBrisorgueil:
@lirantal i have push the client side in ES6 on this PR too 

work done : 
 
- add babel with preset
- add babel annotate
- add babel minify
- del uglify (only for es5)

Actually, just the production build use babel. Development use client side in es6, works fine. 

- are you agree with this or shloud we switch only the server side to es6 ?
- should we use babel to dev env, with the watch maybe, to test es5 in live ?  

All new dependencies are added in devDependencies part ( due to gulp-babel readme.md) 

I'm wainting this merge to update all dependencies and other quick PR. Can't really touch the code before that.

lirantal:
I think development build should also use babel. It should be consistent so if bug happens in dev they would be discovered before they show up on prod due to babeling or something. So let's make sure we use everywhere the same setup regardless of the env.

lirantal:
Great stuff @PierreBrisorgueil!

PierreBrisorgueil:
thx, ok, I realized this to follow the principle of uglify today, clean code realized only for the prod.
Browsers are often portable ES6, and free to dev who wants to go into detail to activate babel in dev :)

I'll take a look on this, but it comes to build and minify in public/dist in the dev environment on watch, maybe it should be a choice. We have to see how long it takes :)

lirantal:
@PierreBrisorgueil ok appreciate it.
ping me back what you end up going for and let's review

PierreBrisorgueil:
@lirantal 

I finished a first version with some modifications. I changed some gulp tasks for that. And added gulp-if-env.

The only thing changed that isn't in the optics of using babel is the call of '' env: prod ',' lint '" in the prod task. I did not understand the goal actually, the build task already contain lint and is also called. So I removed this double call of lint, and move the configuration of the env at the start of the prod task.

**For babel :** 

in dev 
- we now build js and css at every change
- this will create two files in public/dist application.min.js application.min.css without minify to allow debug from the inspector
- this will refresh the page

in prod 
- this will create two files in public/dist application-xxxxx.min.js application-xxxxx.min.css with minify 

Another change is that I deleted the watch on the production part. I do not think it makes sense. for me we should also delete nodemon ... 
But I can look to add him again, if we need this in the stack. As the file names change each time there is just a little problem to manage to refresh the template with the right dependencies.


I didn't have time to watch the TRAVIS problem for the moment. It's really strange, sometimes it's ok, sometimes no ..

lirantal:
gonna be a bit busy for a while but I want to push this forward and merge when we have a stable and working setup!

pinging @meanjs/contributors @meanjs/core if anyone can further review and follow-up please

gulpfile.js
```diff
@@ -58,28 +59,27 @@ gulp.task('nodemon', function () {
...
-  gulp.watch(defaultAssets.client.js, ['eslint']).on('change', plugins.refresh.changed);
-  gulp.watch(defaultAssets.client.css, ['csslint']).on('change', plugins.refresh.changed);
   gulp.watch(defaultAssets.client.sass, ['sass', 'csslint']).on('change', plugins.refresh.changed);
   gulp.watch(defaultAssets.client.less, ['less', 'csslint']).on('change', plugins.refresh.changed);
+  gulp.watch(defaultAssets.client.js, ['eslint', 'babel']);
+  gulp.watch(defaultAssets.client.css, ['csslint', 'cssmin']);
+  gulp.watch('public/dist/application*').on('change', plugins.refresh.changed);
 
-  if (process.env.NODE_ENV === 'production') {
+  if (process.env.NODE_ENV === 'e') {
```
mleanos: Is this a typo? "e"?

PierreBrisorgueil: fix in new pr

PR #2003, opened 2018-12-26
fix(gulp): Gulp tasks override NODE_ENV

mleanos:
Fixes an issue with the "defualt" & "build" Gulp tasks that override the current NODE_ENV. These two tasks should inherit the NODE_ENV that has been set.

Also, adds a sequence to the "prod" Gulp task to ensure the NODE_ENV is set to "prod" before the "build" task is run.


mleanos:
After upgrading to Gulp 4 these changes caused a lot of conflicts due to the format of the Gulp series/parallel tasks.

I'll open a new PR for these fixes.

PR #2045, opened 2020-04-11
Fixed "Resource Not found" error in chat profile picture

jawahirk:
Removed extra '/' which causes the error.


dalelotts:
Thanks for sharing. The project is abandoned so I don't think this PR will be merged. At least others can find it!

